PR #12, opened 2018-08-09
Submodules

roschaefer:
We should not reference a folder on the file system in the
docker-compose.yml. This adds git submodules, which are fine for
read-only access, and adds them as subfolders. Dependabot can update
these submodules automatically.

@appinteractive could you give me write permissions to the origin repository?


roschaefer:
@appinteractive and I cannot add repos of human connection to dependabot.

appinteractive:
@roschaefer added dependabot to this repo. Is it still correct to merge that request? When so its approved.

PR #22, opened 2018-09-17
Create .gitignore

Lulalaby:
Ignore .github folder on pull.
This directory does not need to get copied.


roschaefer:
Are you sure? :thinking: I thought we want to keep the issue templates in there. Maybe edit them later

roschaefer:
I think they should be copied locally. E.g. for editing.

Am I missing sth. important?

roschaefer:
aaand they're gone

mimicc83:
uhh yea editing would be helpful at some point ♐️ 
@Lulalaby maybe it should get back into repo? Does it hurt anyone? :)

appinteractive:
It can be edited inside GitHub. My Bad feeling in the stomach is just that it could just disappear 😬

Lulalaby:
It won't if you don't override it ^^
I don't think that anyone would create locally the same file and pushed it to override.
But if you bad feeling is too strong, let's undo this

roschaefer:
let's stay with it for the moment. It can still be edited on Github.

appinteractive:
Let’s keep it like it is for the moment.

PR #233, opened 2019-03-22
[User profile] Change password

### LINKED ISSUE #232: Change password
```gherkin
Feature: Change password
  As a user
  I want to change my password in my settings
  Because this is a basic security feature, e.g. if I exposed my password by accident
```

This settings page on `/settings/security` should have a three fields `old password`, `new password` and `passwordConfirmation`, if they both match I should be able to change my password:

https://nitro-master.human-connection.org/settings/security

screenshot 34

roschaefer:
close #232


cypress/integration/settings/ChangePassword.feature
```diff
@@ -0,0 +1,31 @@
+Feature: Change password
```
roschaefer: @datenbrei have a look into this user story + scenario

PR #237, opened 2019-03-22
List socialMedia links

mattwr18:
fixes#236


mattwr18:
hey @appinteractive @roschaefer, we have made some progress on this, where we are adding the socialMedia to the backend, and then we are iterating over the urls and displaying a click-able icon to the user's socialMedia account, but it is not updating in real time after a user adds their social media link, just after you refresh the page and it gets the currentUser again.

Could you have a quick look, obviously there are a number of things that still need to be done
- [ ] styling the links
- [ ] limiting the number of social media one can add
- [ ] finish testing
but some feedback on the general direction would be appreciated

Screenshot at 2019-03-26 17:42:39

webapp/pages/settings/my-social-media.vue
```diff
@@ -0,0 +1,73 @@
...
+      margin-top="base"
+      margin="x-small"
+    >
+      <div
+        v-for="socialMediaIconUrl in currentUser.socialMedia"
+        :key="socialMediaIconUrl"
+      >
+        <a>
+          <img
+            :src="socialMediaIconUrl.match(/^(?:https?:\/\/)?(?:[^@\n])?(?:www\.)?([^:\/\n?]+)/g)[0] + '/favicon.ico'"
```
roschaefer: I like the idea but do you know what happens if this doesn't work?

mattwr18: yes, if there is some url that is shared, which does not have an endpoint `/favicon.ico`, then it will display a "broken" image

mattwr18:
@roschaefer there are not backend validations for URL at the moment, but there will be so thanks for writing a test for that use case. At the moment, we are leveraging the styleguide to show an error message if the url is not valid, but of course it would be good to validate this.

This is still a `[WIP]`, what we were specifically looking for feedback at this point, is in my comment from 16 hours ago.  I'm assuming you either didn't notice or maybe don't have any feedback? That's why I asked for @appinteractive feedback as well. 

Thanks for adding those tests though :) it would be great to finish this before the end of the scrum, but I'm running out of time this week to work on it. :worried:

roschaefer:
@mattwr18 ah yeah I didn't noticed that we're blocked on this. I've found [this for backend validations](https://github.com/confuser/graphql-constraint-directive) but I haven't researched how to integrate this.

mattwr18:
> @mattwr18 ah yeah I didn't noticed that we're blocked on this. I've found [this for backend validations](https://github.com/confuser/graphql-constraint-directive) but I haven't researched how to integrate this.

that's great! thanks for the link... it looks like they have a `uri` validation
https://github.com/confuser/graphql-constraint-directive/blob/adb56454beda15e554606bcdf42394f64cdd4f1e/test/string.test.js#L795

backend/src/resolvers/user_management.js
```diff
@@ -100,6 +100,26 @@ export default {
 
         return encode(currentUser)
       }
+    },
+    addSocialMedia: async (_, { url }, { driver, user }) => {
```
roschaefer: @mattwr18 I would consider to create an extra node for `SocialMedia` which contains little but a `url`. This would have a couple of advantages:
* You can utilize `neo4j-graphql-js` for parts of the resolver
* You can add unique constrains on `SocialMedia.url`
* You can use [contraint-directives](https://github.com/confuser/graphql-constraint-directive)
* You can have counts on social media links

backend/src/resolvers/user_management.js
```diff
@@ -100,6 +100,26 @@ export default {
...
+      const session = driver.session()
+
+      const { email } = user
+      const result = await session.run(
+        `MATCH (user:User {email: $userEmail})
```
roschaefer: It does not change a lot but for consistency I would use `userId` here.

cypress/integration/SocialMedia.feature
```diff
@@ -0,0 +1,14 @@
...
+
+  Background:
+    Given I have a user account
+    And I am logged in
+    And I am on the "settings" page
+
+  Scenario: Adding Social Media
+    Given I click on the "My social media" link
+    Then I should be on the "/settings/my-social-media" page
+    And I should be able to add a social media link
```
roschaefer: I would define the action to be "adding the link" and the assertion to be "the link gets saved and displayed":
`When I add a social media link`
`Then it gets saved successfully`
`And the new social media link shows up on the page`

webapp/pages/settings/my-social-media.vue
```diff
@@ -0,0 +1,86 @@
...
+export default {
+  data() {
+    return {
+      value: ''
+    }
+  },
+  computed: {
+    ...mapGetters({
+      currentUser: 'auth/user'
+    }),
```
roschaefer: @mattwr18 you asked why the page does not get updated: You have to update `currentUser`. Import `mapMutations` put it in `methods` and use it when `apollo` successfully executes the mutation.

mattwr18: we tried doing that, but I think we didn't push up the code to have a look if we implemented it correctly... I can add it back and push it up

webapp/pages/settings/my-social-media.vue
```diff
@@ -0,0 +1,86 @@
...
+        <ds-list-item
+          v-for="link in socialMediaLinks"
+          :key="link.url"
+        >
+          <a :href="link.url">
+            <img
+              :src="link.favicon"
+              alt=""
+            >
+            {{ link.url }}
```
mattwr18: should this be the entire url @roschaefer, we were thinking of either just the username, or just having the icon as the link
Screenshot at 2019-03-28 14:43:48

webapp/pages/settings/my-social-media.vue
mattwr18: @aonomike and I were able to get it updating the store and now the icons and links are showing up when the user adds the social media url!! 

the favicons are quite small from twitter and facebook for example... not sure if there is a way to size them without blurring them... for the time being, I have added a 16x16 constraint on all icons, but it seems to small to me... also, we obviously will work on the other review items you brought up @roschaefer like limiting how many a user can add, validating the url in the back end, etc... 
Screenshot at 2019-03-28 15:46:49

webapp/pages/settings/my-social-media.vue
roschaefer: I think it has some advantage to see the url, people like to know where the icon points to.

If you know how to get the username out of an arbitrary url, go for it :thinking: :woman_shrugging:

webapp/pages/settings/my-social-media.vue
roschaefer: It's OK, I think we can resize them later. For styles and visuals you can also consult @appinteractive. 

It does not look so bad to me honestly.

webapp/pages/settings/my-social-media.vue
Lulalaby: I prefer that only the username is filtered from the URL and displayed with the icon.
Otherwise it might get too long.

roschaefer: @mattwr18 if it turns out that filtering the username from the URL is not that easy, don't spend too much time on it. I agree with @Lulalaby but a quick fix could also be that you simply shorten the URL like `/some/path...`.

mattwr18: I think potential it could be not too complicated @roschaefer... the regex that @Kachulio1 wrote :point_up: separates the the beginning of the URL from the username already, so potentially we just need to save what it cuts as the username, but we'll have a look at it

Lulalaby: yeah, just do it! 😸

backend/src/resolvers/user_management.js
mattwr18: > @mattwr18 I would consider to create an extra node for `SocialMedia` which contains little but a `url`. This would have a couple of advantages:
> 
> * You can utilize `neo4j-graphql-js` for parts of the resolver
> * You can add unique constrains on `SocialMedia.url`
> * You can use [contraint-directives](https://github.com/confuser/graphql-constraint-directive)
> * You can have counts on social media links

I need to ask for some clarification on this @roschaefer... if I understand what you mean correctly, you are suggesting
1. create a `type` called `SocialMedia`, "which contains little but a url"
2. create an relation between `User` and `SocialMedia`, when a user sends a link to add their social media?
3. Add the constraints listed above, i.e. unique urls, limited amount of social media links

Is this what you are suggested, or did I misunderstand?

roschaefer: Yes! Correct. Don't worry too much about the constraints though. Maybe two people share a link? (e.g. an organization or sth.) same same with the number of SocialMedia.

It's just that I believe you get most out of Neo4J if you model relationships through relationships not through properties.

roschaefer: Ah, by `Counts on SocialMedia` I meant a computed property like `socialMediaCount` and it was just an example for sth. you can do with `neo4j-graphql-js` easily which you can't do with properties so easily.

backend/src/schema.graphql
```diff
@@ -310,3 +310,12 @@ type SharedInboxEndpoint {
...
   uri: String
 }
+
+type SocialMedia {
+  id: ID!
+  uri: String
+}
+
+input urlInput {
+  url: String! @constraint(format: "uri")
```
roschaefer: @mattwr18 nice! Does it work??

backend/src/schema.graphql
```diff
@@ -310,3 +310,12 @@ type SharedInboxEndpoint {
   id: ID!
   uri: String
 }
+
+type SocialMedia {
+  id: ID!
+  uri: String
```
roschaefer: We have to find a solution for read-only fields. There are attributes that are read-only (disabled flag)  write-only (password) or hidden altogether (privateKey). I think this is a problem of `neo4j-graphql-js`.

If the above problems would not exist I would prefer to have:
```graphql
  addSocialMedia(url: SocialMedia): [SocialMedia]!
```
By the way: @mattwr18 is it not possible to return a SocialMedia object here?

backend/src/schema.graphql
mattwr18: no... you probably didn't see, but I wrote in my commit message the back end is not compiling with `Error: Unknown directive "constraint"`

I needed to go, so I pushed it up cause @Tirokk wanted to have a closer look

backend/src/schema.graphql
mattwr18: maybe, we were just starting to get this working, ran into some issues and I had to get on with other meetings... I'll just have time to look at it tomorrow I guess

Tirokk: We have still the error:
```
Error: Cannot find module 'graphql-constraint-directive'
```

I was not able to find out why … @mattwr18 @roschaefer

Tirokk: @roschaefer And I do not understand yet, why you like to have the social media URLs of a person as different nodes? Each URL is unique, because it contains the username of the user on this platform.
Or you like to have the domains of each social media platform in one node and the username stored at the user or as a property in the relation to this social media link?

mattwr18:
got passed the error :point_up: 
currently, creating a social media node is being blocked with this error message
```
{
  "data": {
    "createSocialMedia": null
  },
  "errors": [
    {
      "message": "Property values can only be of primitive types or arrays thereof",
      "locations": [
        {
          "line": 2,
          "column": 3
        }
      ],
      "path": [
        "createSocialMedia"
      ]
    }
  ]
}
```

backend/src/schema.graphql
roschaefer: Wow! [KISS-principle](https://en.wikipedia.org/wiki/KISS_principle) please! :pray:

The motivation for having `SocialMedia` as extra node is just a matter of design. Before we had social media urls modeled as properties on a user node. But actually, it is a one-to-many relationship and as I said, it's better to use Neo4J's relationships to model relationships, don't use properties. I mean, after all that's what those relationships are for. Plus, it is likely, that our friend `neo4j-graphql-js` understands better what we're doing here if we use our tools in the originally intended way.

The reason why I said "Don't worry about unique constraints" is that we did not think hard enough for use cases why two people might want to share the same URL. Think of a shared organization account as an example.

And don't model social media platforms as enumerable types. Just use URLs for now. We can make it more complex later on.

In general :kiss: kiss: keep-it-simple-stupid

roschaefer:
@mattwr18 interesting! Is it a progress? I cannot make sense of the error message right now.

mattwr18:
> @mattwr18 interesting! Is it a progress? I cannot make sense of the error message right now.

it is progress in that we are now able to register the directive, but now we need to completely change our resolver to create the social media node and associate it with the user

mattwr18:
Currently, blocked with a weird message when trying to log in, which apparently has been happening for a while. I was working on the back end part of the feature and so didn't notice it. 
When I try to log in, I get this message `Network error: Response not successful: Received status code 400`, it is adding the cookies, but then throwing an error... the backtrace points to this file `node_modules/apollo-client/bundle.umd.js`
```
var ApolloError = (function (_super) {
        tslib.__extends(ApolloError, _super);
        function ApolloError(_a) {
            var graphQLErrors = _a.graphQLErrors, networkError = _a.networkError, errorMessage = _a.errorMessage, extraInfo = _a.extraInfo;
            var _this = _super.call(this, errorMessage) || this;
            _this.graphQLErrors = graphQLErrors || [];
            _this.networkError = networkError || null;
            if (!errorMessage) {
                _this.message = generateErrorMessage(_this);
            }
```
highlighting `var _this = _super.call(this, errorMessage) || this;`

I've been able to trace back to the commit where the error started, which seems to be originating here https://github.com/Human-Connection/Human-Connection/pull/237/commits/a51190d4e9b1498805b74ca6159e2d0cb6757514

I have tried removing the changes between this and the previous commit, but to no avail, and the error persists through the subsequent commits. 

The back end seems to be operating as it should, in the sense that I am able to log in, and also able to `CreateSocialMedia`

any ideas @Kachulio1 @roschaefer @ulfgebhardt @appinteractive

mattwr18:
@roschaefer @Kachulio1 @Tirokk @ulfgebhardt @appinteractive 
it seems we are not exhaustively testing this, but wanted to get some feeback

backend/src/middleware/validUrlMiddleware.js
```diff
@@ -0,0 +1,18 @@
+const validURL = str => {
+  const isValid = str.match(/^(?:https?:\/\/)(?:[^@\n])?(?:www\.)?([^:/\n?]+)/g)
+  return !!isValid
+}
+
+export default {
+  Mutation: {
+    CreateSocialMedia: async (resolve, root, args, context, info) => {
```
roschaefer: @mattwr18 what about `UpdateSocialMedia`? :slightly_smiling_face: Or do we always want to Delete+Create?

backend/src/middleware/validUrlMiddleware.js
```diff
@@ -0,0 +1,18 @@
+const validURL = str => {
+  const isValid = str.match(/^(?:https?:\/\/)(?:[^@\n])?(?:www\.)?([^:/\n?]+)/g)
```
roschaefer: Would be great to see this unit tested. No need for the database to be involved.

roschaefer: E.g. I wonder if https is required or not.

roschaefer: and what about a missing domain? Like: `www.google`

backend/src/resolvers/socialMedia.spec.js
```diff
@@ -0,0 +1,49 @@
+import Factory from '../seed/factories'
+import { GraphQLClient } from 'graphql-request'
+import { host, login } from '../jest/helpers'
+
+const factory = Factory()
+
+describe('CreateSocialMedia', () => {
```
roschaefer: @mattwr18 who is allowed to `DeleteSocialMedia`? :thinking:

webapp/locales/en.json
```diff
@@ -15,7 +15,8 @@
     "followers": "Followers",
     "following": "Following",
     "shouted": "Shouted",
-    "commented": "Commented"
+    "commented": "Commented",
```
roschaefer: Oh yes, providing translations right from the beginning :sparkling_heart:

webapp/pages/profile/_id/_slug.vue
```diff
@@ -348,6 +379,19 @@ export default {
         return []
       }
       return this.uniq(this.user.contributions.filter(post => !post.deleted))
+    },
+    socialMediaLinks() {
+      const { socialMedia = [] } = this.user
+      return socialMedia.map(socialMedia => {
+        const { url } = socialMedia
+        const matches = url.match(
+          /^(?:https?:\/\/)?(?:[^@\n])?(?:www\.)?([^:\/\n?]+)/g
```
roschaefer: I wonder if we could avoid to roll our own URL regex. How about using NodeJS `url` module? https://stackoverflow.com/questions/30931079/validating-a-url-in-node-js

roschaefer:
@mattwr18 are you not tired yet by one code review after another? :laughing: Here's what I have found:

Screenshot (76)

backend/src/middleware/validUrlMiddleware.js
mattwr18: I was talking to @Tirokk about this this morning.
The ticket doesn't say anything about updating or deleting social media
So, I would probably create two separate issues for those use cases, which can be separately estimated and potentially would be beginner-friendly tickets for people to work on.

backend/src/middleware/validUrlMiddleware.js
mattwr18: I am not an expert on regex, but I believe I tested this and the `s` is optional, which is why it has a `?` on the right of it?

backend/src/middleware/validUrlMiddleware.js
roschaefer: @mattwr18 OK cool! That's a solution as well. Just make sure we don't forget these things. :+1:

backend/src/middleware/validUrlMiddleware.js
mattwr18: not sure what you mean by missing domain... as in no `.com`, etc?
I think this is the regex you came up with and I copied from `my-social-media.vue`... could be that we could improve on it...

webapp/locales/en.json
mattwr18: :smile_cat:

backend/src/middleware/validUrlMiddleware.js
roschaefer: I'm sorry, I meant **top-level**-domain.

webapp/pages/profile/_id/_slug.vue
mattwr18: that is an interesting idea.. I'm ok to do that, I just don't know how other people feel about adding an npm package for just that, but maybe I am reading a different answer on that SO post you posted :point_up:

backend/src/middleware/validUrlMiddleware.js
roschaefer: Up to you! If you want to get the PR merged, maybe create `good-first-issue` for refactoring?

webapp/pages/profile/_id/_slug.vue
roschaefer: [nodejs url](https://nodejs.org/api/url.html) is a package which is available by default on NodeJS. Maybe we have to polyfill it on the browser (if it's not already polyfilled by Vue&friends).

backend/src/middleware/validUrlMiddleware.js
mattwr18: I can take responsibility for adding the tickets :smile_cat:

backend/src/middleware/validUrlMiddleware.js
mattwr18: can be, that sounds like a good compromise that would potentially allow more conversation on the subject

backend/package.json
```diff
@@ -10,8 +10,8 @@
     "dev:debug": "nodemon --exec babel-node --inspect=0.0.0.0:9229 src/index.js -e js,graphql",
     "lint": "eslint src --config .eslintrc.js",
     "test": "run-s test:jest test:cucumber",
-    "test:before:server": "cross-env GRAPHQL_URI=http://localhost:4123 GRAPHQL_PORT=4123 yarn run dev",
-    "test:before:seeder": "cross-env GRAPHQL_URI=http://localhost:4001 GRAPHQL_PORT=4001 DISABLED_MIDDLEWARES=permissions,activityPub yarn run dev",
+    "test:before:server": "cross-env GRAPHQL_URI=http://localhost:4123 GRAPHQL_PORT=4123 yarn run dev 2> /dev/null",
```
ulfgebhardt: Not allowed under windows
```suggestion
    "test:before:server": "cross-env GRAPHQL_URI=http://localhost:4123 GRAPHQL_PORT=4123 yarn run dev",
```

backend/package.json
```diff
@@ -10,8 +10,8 @@
...
+    "test:before:seeder": "cross-env GRAPHQL_URI=http://localhost:4001 GRAPHQL_PORT=4001 DISABLED_MIDDLEWARES=permissions,activityPub yarn run dev 2> /dev/null",
```
ulfgebhardt: Not allowed under windows
```suggestion
    "test:before:seeder": "cross-env GRAPHQL_URI=http://localhost:4001 GRAPHQL_PORT=4001 DISABLED_MIDDLEWARES=permissions,activityPub yarn run dev",
```

PR #248, opened 2019-03-25
240 persistent links

### LINKED ISSUE #240: Persistent Links
```gherkin
Feature: URL with unmodifiable ID
  As a user
  I want to click on a link with an `href` having an unmodifiable id of a user
  In order to have persistent links even if the user changes a his/her slug
```

roschaefer:
close #240


cypress/integration/06.Search.feature
```diff
@@ -8,7 +8,7 @@ Feature: Search
     And we have the following posts in our database:
       | Author        | id | title                                         | content                                 |
       | Brianna Wiest | p1 | 101 Essays that will change the way you think | 101 Essays, of course!                  |
-      | Brianna Wiest | p1 | No searched for content                       | will be found in this post, I guarantee |
+      | Brianna Wiest | p2 | No searched for content                       | will be found in this post, I guarantee |
```
mattwr18: thanks for fixing this, I must have been crazy to create two posts with the same id!!

cypress/integration/06.WritePost.feature
```diff
@@ -17,7 +17,7 @@ Feature: Create a post
     for active citizenship.
     """
     And I click on "Save"
-    Then I get redirected to "/post/my-first-post/"
+    Then I get redirected to ".../my-first-post"
```
mattwr18: It's not so clear how this works until one checks out https://github.com/Human-Connection/Human-Connection/pull/248/files#diff-cc51d55abc8d6a1b33fd1f8885365953R227, but maybe there is nothing one can do about that

cypress/integration/identifier/PersistentLinks.feature
```diff
@@ -0,0 +1,41 @@
+Feature: Persistent Links
+  As a user
+  I want to click on a link with an 'href' having an unmodifiable id of a user
```
mattwr18: I doubt this is how a user of the site would think about it, but maybe I am just being pedantic.... 
I would expect them to think more along the lines of `I want to avoid clicking on links that lead nowhere or an error page`

webapp/mixins/persistentLinks.js
```diff
@@ -0,0 +1,32 @@
...
+      } = context
+      const idOrSlug = id || slug
+
+      const variables = { idOrSlug }
+      const client = apolloProvider.defaultClient
+
+      let response
+      let thing
+      response = await client.query({ query: queryId, variables })
+      thing = response.data[Object.keys(response.data)[0]][0]
```
mattwr18: I know naming is difficult, but `thing` is not at all clear as to what it is supposed to represent

roschaefer: resource?

roschaefer: at the moment it can be a `post` or a `user`

roschaefer:
Hm, I don't think this is a good choice for a component test. We have server-side-rendering involved and some data setup. That's why I believe a fullstack test is required here.

But in general: Yes I would prefer component tests, especially for a long list of test cases.

mattwr18:
> Hm, I don't think this is a good choice for a component test. We have server-side-rendering involved and some data setup. That's why I believe a fullstack test is required here.
> 
> But in general: Yes I would prefer component tests, especially for a long list of test cases.

ok, I'm happy to approve then, and if someone comes along that can think of some use case for component testing, we can always create another ticket

PR #362, opened 2019-04-02
Fixes #361

### LINKED ISSUE #361: ActivityPub: outbox#first results in HTTP 500 error
**Describe the bug**
I just seeded again with `activityPub` middleware enabled and I can see public keys on nitro:
Screenshot (67)
But I still cannot see my posts on Mastodon:
Screenshot (68)

I believe now it's because of a HTTP 500 error on our nitro:

**To Reproduce**
1. Visit https://nitro-staging.human-connection.org/activitypub/users/jenny-rostock/outbox
2. Click on ' "first": "https://nitro-staging.human-connection.org/activitypub/users/jenny-rostock/outbox?page=true",'
3. See "This page is not working" (500 error)

**Expected behavior**
I would expect the first post.

**Desktop (please complete the following information):**
 - OS: Archlinux
 - Browser `Chromium 73.0.3683.86 Arch Linux`

Mastercuber:


backend/src/activitypub/ActivityPub.js
```diff
@@ -22,16 +22,16 @@ let activityPub = null
 export { activityPub }
 
 export default class ActivityPub {
-  constructor (activityPubEndpointUri, internalGraphQlUri) {
-    this.endpoint = activityPubEndpointUri
+  constructor (internalGraphQlUri) {
+    this.endpoint = internalGraphQlUri
```
roschaefer: @Mastercuber are you sure? `GRAPHQL_URI` is an cluster-internal endpoint not exposed to the internet.

Screenshot (69)

Mastercuber: @roschaefer oh I remember, so there is a switch needed for using the internal endpoint for development and the other endpoint for production. Maybe when starting the Server in dev mode, an environment var could also be set?

roschaefer: Is it required? We proxy `/api` to `GRAPHQL_URI`. So if the backend knows `CLIENT_URI` the publicly available graphql endpoint is `$CLIENT_URI/api`. Does that help?

I still don't understand this PR by the way.

Mastercuber: Initially it was a fix for serving paged collections. Yes it helped, I deleted the second commit. So this is again a quick fix. 

By the way a recognized that on the staging not only the `/api/activitypub` endpoint, but also the `/activitypub` endpoint is exposed.

roschaefer: *and* `/.well-known`

PR #406, opened 2019-04-04
Don't expose private RSA key

roschaefer:
cc @Tirokk

### Issues

- [X] None

### Checklist

- [X] None

### How2Test

- [X] None

### Todo

- [X] None


Tirokk:
Fixes #411

Tirokk:
@roschaefer @mattwr18
I came to the point where I implemented a test, but was not able to generate a user with a „publicKey“ not „null“.

I changed the test to proof that the „publicKey“ is not generated by comment the things out which made the „publicKey“ invisible. So it should be visible in the error which I produced here intentionally.

To see run:
```
$ docker-compose exec backend yarn test:jest backend/src/resolvers/user_management.spec.js
```

backend/src/resolvers/user_management.spec.js
```diff
@@ -309,3 +310,109 @@ describe('change password', () => {
...
+      // logged out
+      client = new GraphQLClient(hostPrivateKey)
+      // Generate user with "privateKey".
+      const id = 'bcb2d923-f3af-479e-9f00-61b12e864667'
+      const name = 'Apfel Strudel'
+      const slug = 'apfel-strudel'
+      const password = 'xYz'
+      await client.request(gql`
+        mutation {
+          CreateUser(id: "${id}", password: "${password}", slug:"${slug}", name: "${name}", email: "${slug}@test.org") {
```
roschaefer: Use variables here instead of string interpolation.

backend/src/resolvers/user_management.spec.js
```diff
@@ -309,3 +310,109 @@ describe('change password', () => {
     })
   })
 })
+
+describe('don\'t expose private RSA key', () => {
+  const queryUser = params => {
+    const { queriedUserSlug } = params
+    return gql`
+      {
```
roschaefer: You can rewrite that whole query with something like:
```
query(slug: String!) {
  User(slug: $slug) ...
}

**Comment by Tirokk** [comment]
This PR can be reviewed @roschaefer .

File: backend/src/resolvers/user_management.spec.js [hunk 3]
```diff
@@ -309,3 +310,58 @@ describe('change password', () => {
...
+    }
+    await client.request(gql`
+      mutation($id: ID, $password: String!, $slug: String, $name: String, $email: String) {
+        CreateUser(id: $id, password: $password, slug: $slug, name: $name, email: $email) {
+          id
+        }
+      }`, variables
+    )
+  }
+
```
roschaefer:
```suggestion
describe('unauthenticated query of "publicKey"', () => {
  it('returns publicKey', () => {
    // check that an RSA key pair has been created at all
  })
})
```

Tirokk: I have done an authenticated query for „publicKey“, because I am not sure for the future, if unauthenticated queries of user data should be possible.

Would be nice to get a review by @mattwr18 or @ulfgebhardt , because @roschaefer can not review his own PRs.

backend/src/resolvers/user_management.spec.js
```diff
@@ -309,3 +310,81 @@ describe('change password', () => {
...
+    it('throws "Not Authorised!"', async () => {
+      await actionGenUserWithKeys()
+      await expect(
+        client.request(queryUserPrivateKey, { queriedUserSlug: 'apfel-strudel' })
+      ).rejects.toThrow('Not Authorised')
+    })
+  })
+
+  describe('authenticated query of "publicKey". To asure generation of "privateKey" is garanteed', () => {
+    it('throws "Not Authorised!"', async () => {
```
roschaefer:
```suggestion
    it('returns publicKey', async () => {
```

backend/src/resolvers/user_management.spec.js
```diff
@@ -309,3 +310,81 @@ describe('change password', () => {
...
+  describe('authenticated query of "privateKey"', () => {
+    it('throws "Not Authorised!"', async () => {
+      await actionGenUserWithKeys()
+      await expect(
+        client.request(queryUserPrivateKey, { queriedUserSlug: 'apfel-strudel' })
+      ).rejects.toThrow('Not Authorised')
+    })
+  })
+
+  describe('authenticated query of "publicKey". To asure generation of "privateKey" is garanteed', () => {
```
roschaefer:
```suggestion
  describe('unauthenticated query of "publicKey" (does the RSA key pair get generated at all?)', () => {
```

backend/src/resolvers/user_management.spec.js
```diff
@@ -309,3 +310,81 @@ describe('change password', () => {
...
+        query($queriedUserSlug: String) {
+          User(slug: $queriedUserSlug) {
+            id
+            publicKey
+          }
+        }`
+
+      await actionGenUserWithKeys()
+      await expect(
+        await client.request(queryUserPuplicKey, { queriedUserSlug: 'apfel-strudel' })
```
roschaefer: @Tirokk it is important that `client` is unauthenticated here. For ActivityPub, unauthenticated 3rd-party servers like Mastodon and Friendica must be able to access an actor's public RSA key to check a the signature of e.g. a Post.

So please move this test case out of the `describe` block that logs in the `client`.

Tirokk: Okay, I think I understand @roschaefer .
But do you think that any user data should be queried unauthenticated in the future? Like it is now …

roschaefer: :man_shrugging: dunno. Everything except email or password is OK to be public for now. I think we need more user testing and we should also consult @datenbrei in these matters

Tirokk: Okay. I noticed that as question to clarify.

backend/src/middleware/passwordMiddleware.js
```diff
@@ -11,10 +11,11 @@ export default {
     }
   },
   Query: async (resolve, root, args, context, info) => {
-    const result = await resolve(root, args, context, info)
-    return walkRecursive(result, ['password'], () => {
+    let result = await resolve(root, args, context, info)
+    result = walkRecursive(result, ['password', 'privatKey'], () => {
```
roschaefer:
```suggestion
    result = walkRecursive(result, ['password', 'privateKey'], () => {
```
Scared

Tirokk:
😖 Sorry! That shouldn’t happen.
I tested your suggestion again and it works fine !!!

backend/src/resolvers/user_management.spec.js
```diff
@@ -254,7 +255,7 @@ describe('change password', () => {
   }
 
   describe('should be authenticated before changing password', () => {
-    it('throws not "Not Authorised!', async () => {
+    it('throws not "Not Authorised!"', async () => {
```
mattwr18: please remove redundant `not`

mattwr18: nevermind... I just did it and pushed it up

PR #428, opened 2019-04-05
Create notifications for mentions

roschaefer:
## Pullrequest
Create a notification for a simple @mention by slug.

### Issues
 [X] None

### Checklist

- [X] None

### How2Test

- [X] `yarn run test`

### Todo

- [X] None

### DIFF
--- a/backend/src/middleware/index.js
+++ b/backend/src/middleware/index.js
@@ -10,6 +10,7 @@ import permissionsMiddleware from './permissionsMiddleware'
 import userMiddleware from './userMiddleware'
 import includedFieldsMiddleware from './includedFieldsMiddleware'
 import orderByMiddleware from './orderByMiddleware'
+import notificationsMiddleware from './notificationsMiddleware'
 
 export default schema => {
   let middleware = [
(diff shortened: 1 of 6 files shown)


ulfgebhardt:
Approved - but can you provide screenshots how it looks when you mention someone?
Also: is this methology compatible with comments and other forms of mentions required to push to the user as notification?

roschaefer:
@ulfgebhardt I cannot provide a screenshot. This is backend only. The method to create a notification is compatible if you say that both posts and comments can `(post)-[:NOTIFIED]->(notification)` and `(comment)-[:NOTIFIED]->(notification)`. The notification would then `(notification)-[:NOTIFIED]->(user)`.

The optimal solution would be to omit the additional node in between and simply create properties on the relationship itself: `(post)-[:NOTIFIED {read: false}]->(user)`. However I believe this creates headaches for our framework `neo4j-graphql-js` which takes GraphQL types as input to build queries and mutations and has no clue about properties on relationships.

roschaefer:
already covered by #445

Issue #392, opened 2019-04-04
💬 [Question] Design Differences of the PostPage


ulfgebhardt:
### Design Differences of the PostPage
There are Design-Differences between Alpha and Nitro regarding the display of the Post in the Feed.

- [X] Title and Author Position reversed
- [X] Design & Format of Post-Date + Position
- [X] No Category Names
- [X] Actions/Reactions & Tags reversed

### Alpha

![image](https://user-images.githubusercontent.com/1238238/55542224-de069f00-56c6-11e9-9bf5-29e56b5a926f.png)

### Nitro

![image](https://user-images.githubusercontent.com/1238238/55542111-a992e300-56c6-11e9-89fc-fb74bf2ae057.png)

PR #440, opened 2019-04-08
2019/kw15/design_differences_post

ulfgebhardt:
## Pullrequest

image

fixed design differences between alpha & nitro - Post Page
- [X] Title and Author Position reversed
- [X] Design & Format of Post-Date + Position
- [X] No Category Names
- [X] Actions/Reactions & Tags reversed

### Issues

- [X] fixes https://github.com/Human-Connection/Human-Connection/issues/392

### Checklist

- [X] None

### How2Test

- [X] None

### Todo

- [X] None


roschaefer:
@ulfgebhardt I think the build server is failing for good reason, could you check?

Tirokk:
Moin @roschaefer , the Jest component test of `RelativeDateTime` works fine and looks nice. Congrats !!!

webapp/components/RelativeDateTime/index.vue
```diff
@@ -0,0 +1,33 @@
...
+import formatRelative from 'date-fns/formatRelative'
+import { enUS, de, nl, fr, pt, es /*, pl*/ } from 'date-fns/locale'
+const locales = {
+  en: enUS,
+  de,
+  nl,
+  fr,
+  es,
+  pt
+  //pl
```
roschaefer: What's wrong with Polish? :thinking: 

@ulfgebhardt by the way, I think I missed one test case: What happens, if you have a `i18n.locale()` but this locale is not in the list above? Could you create a test case for it? I would like to have a default in that case.

roschaefer:
@ulfgebhardt the build server fails at `✖ 06.WritePost.feature` on both builds. I don't think this is a false positive (test fails, but implementation is fine) but a true positive (test fails because implementation is broken). Could you run the cypress tests locally and have a check?

webapp/components/RelativeDateTime/index.vue
ulfgebhardt: Polish is currently not implemented on library side.

ulfgebhardt:
image
image
image
image

Cypress tests are not working with chrome or electron
```
yarn cypress:setup
yarn cypress:run (changed browser)
```

Error message from travis is not reproducable
image
image

Posts can be created - works as expected

PR #444, opened 2019-04-09
Disabling activityPub middleware again

roschaefer:
## Pullrequest
I thought it might be a good thing to create RSA keys at least for
cucumber testing. At least that's where we test activitityPub. But it
caused too flaky tests and broken builds. @Mastercuber if you need to
create a user and you need the RSA keys, don't create it through seeds
but through the actual server.

### Issues

- [X] None

### Checklist

- [X] None

### How2Test

- [X] None

### Todo

- [X] None


ulfgebhardt:
@roschaefer

Issue #219, opened 2019-03-08
Add creation date to follow and shout relation


appinteractive:
We need to add a date to the shout and follow relationships to also maybe able to sort them.

Tirokk:
@roschaefer @appinteractive @ulfgebhardt @mattwr18 
Is it not interesting since when I do follow somebody?

KapilJ22:
@mattwr18, The issue has been resolved with the following PR:
https://github.com/Human-Connection/Human-Connection/pull/1853

mattwr18:
yes, if you write `fixes #<issue_number>` in the `PR` description, the issue will get closed automagically when the `PR` is merged... we added it to the `PR` template for this reason. 
thanks again for working on it :)

PR #449, opened 2019-04-11
refactor follow, unfollow, shout, unshout to custom mutations, but don't add date-time to relation yet

Tirokk:
Together with @mattwr18 . 😄

### Issues
- relates #219
- [X] Customises mutations follow, unfollow, shout, unshout.
- [ ] Add date-time to relation.

### How2Test

- [ ] `$ docker-compose exec backend yarn test:jest backend/src/resolvers/follow.spec.js backend/src/resolvers/shout.spec.js`


backend/src/resolvers/shout.js
```diff
@@ -0,0 +1,51 @@
+export default {
+  Mutation: {
+    shout: async (_object, params, context, _resolveInfo) => {
+      const { id, type } = params
+
+      const session = context.driver.session()
+      let sessionRes = await session.run(
```
mattwr18: I don't love the name `sessionRes`
I feel like we should be naming it what is the actual result, which would tempt me to name it `isShouted`, but then I see on line 19, we seem to get a list of records back that we then need to grab `isShouted` from.

I was having a read here https://neo4j.com/docs/driver-manual/current/sessions-transactions/, and it seems the result is from the transaction, not the session. Maybe it would be better `transactionRes`, but maybe I'm just being pedantic, and can be ignored.

Tirokk: I am really happy to rename `sessionRes`  to ´transactionRes´ @mattwr18 . That is a very good clarification we can use in the future for all customised mutations. 🤗

I am happy especially if there is nothing else to mention from your side. 😁

backend/src/resolvers/shout.js
```diff
@@ -0,0 +1,51 @@
+export default {
+  Mutation: {
+    shout: async (_object, params, context, _resolveInfo) => {
+      const { id, type } = params
+
+      const session = context.driver.session()
+      let transactionRes = await session.run(
+        `MATCH (node {id: $id})<-[:WROTE]-(userWritten:User), (user:User {id: $userId})
+          WHERE $type IN labels(node) AND NOT userWritten.id = $userId
```
roschaefer: A future improvement could be that you can omit the `ShoutTypeEnum`. I mean: The server knows what you can shout and what not, right? So, instead of having the type of the shouted object sent to the server, you just send the id and the server checks if the labels of the shouted object intersects with a well-known set of shoutable objects (by a set of labels `['Project', 'Post', ...]`).

For now it is fine though!

Tirokk: That is a concurrence between two things:
- KISS: Keep it stupid simple …
- But redundancy may be good for think and error checking: What type has the Webapp programmer thought to be shouted or followed and does this fit to what can be … on the server side?

I think you like to KISS … 😘

Lulalaby: 😆

roschaefer:
If this is work-in-progress, please prefix the title with `[WIP]`. If not, please request a review and I'll merge in case I approve.

However I don't see any tests in this PR :thinking:

mattwr18:
@roschaefer I believe that @Tirokk was asking for a review on this in a meeting, maybe not here officially, but I approved the changes this weekend, after looking at the code, and noticing that you had said it's find for now. Also, @Lulalaby had a look as well.

I think you don't see any tests as part of this `PR` because there were already tests for these mutations see https://github.com/Human-Connection/Human-Connection/blob/219-refactor-follow-shout-mutations/backend/src/resolvers/shout.spec.js and https://github.com/Human-Connection/Human-Connection/blob/219-refactor-follow-shout-mutations/backend/src/resolvers/follow.spec.js

Arguably, I should have just merged it in, but I thought since I approved the changes, maybe somebody else could merge.

Tirokk:
It’s like @mattwr18 said @roschaefer and test already existing.
But sorry, I have not made an official review request. My fault. I just did now!

PR #461, opened 2019-04-12
Yarn workspaces

roschaefer:
## Pullrequest
We could save us keystrokes by using [yarn workspaces](https://yarnpkg.com/lang/en/docs/workspaces/). The disadvantage is that we would expect all our users to use `yarn` and not `npm`. Opinions?

### Issues

- [X] None

### Checklist
- [ ] Docs need to be updated (no npm)
- [ ] Does it have an effect on development with docker?

### How2Test

- [X] None

### Todo

- [X] None


mattwr18:
funny that you bring this up.... it is one of the things I was referencing when I talked about that talk on Saturday about using a monorepo and leveraging all its benefits :smile_cat:

mattwr18:
Closing this as it is not seen as a priority while we are in the process of migrating Alpha to Nitro

Issue has been created to research this
https://github.com/Human-Connection/Human-Connection/issues/588

PR #464, opened 2019-04-14
Docs improve installation instructions

roschaefer:
## Pullrequest
* Don't fork but ask for write permissions
* Merge installation instructions in root folder with backend folder

### Issues

- [X] None

### Checklist

- [X] None

### How2Test

- [ ] Check out https://docs.human-connection.org/human-connection/v/docs-improve_installation_instructions/

### Todo

- [X] None


backend/README.md
```diff
@@ -1,137 +1,155 @@
...
 ```
+Configure the new files according to your needs and your local setup.
 
-Copy:
+Create unique indices with:
 
-```text
-cp .env.template .env
+```bash
+$ ./neo4j/migrate.sh
```
mattwr18: some commands are assuming a person is at the root level and some from the `backend` directory...
Also, most of this is a replica of `installation.md` at the root level...
what is the purpose of having the duplication @roschaefer

roschaefer: This PR tries to remove duplication. See [here](https://docs.human-connection.org/human-connection/v/docs-improve_installation_instructions/installation) it's not in the `installation.md` anymore.

The reason for having this in `./backend` is to have it as close to the code as possible.

roschaefer: But the folder must be fixed. It should be a comment in the code block
```bash
# in path/to/some/where
....
````

Issue #305, opened 2019-03-27
[Epic] Comments


datenbrei:
```gherkin
Feature: Comments
  As a user
  I want to comment on contributions of others
  In order to express my thoughts and emotions about these, discuss and add further information.
```

Tirokk:
# Upvote

Should it be possible in old Alpha Posts to up-vote comments?

Because no reply was possible, the order of the comments are in a kind of reply order.
Therefore I would suggest to disable the up-vote possibility in this case.

ulfgebhardt:
Use the Editor Component

Tirokk:
We keep it till the notification for disabling is done soon.

Tirokk:
The two open checkboxes have to be fulfilled! 👆🏼

Issue #382, opened 2019-04-03
🌟 [EPIC] Post


ulfgebhardt:
### Post
![image](https://user-images.githubusercontent.com/1238238/55485052-f4aae880-5629-11e9-83db-fb1e08816146.png)

Create Post:
- [ ] See Feature #379

Post:
- [X] Author
- [X] Post-Date
- [X] Title
- [X] Content
- [X] Categories
- [X] Tags
- [ ] Delete teaser image #1188

Content Menü (dots):
- [X] Report as User
- [X] Deactivate as Admin
- [X] Edit as Author
- [x] (optional) Delete Post as Author https://github.com/Human-Connection/Human-Connection/issues/330

Actions:
- [x] Emotions https://github.com/Human-Connection/Human-Connection/issues/384
- [X] Shouts & Shout Button

Comments https://github.com/Human-Connection/Human-Connection/issues/305: 
- [X] Number of Comments
- [X] Author
- [x] Comment-Date https://github.com/Human-Connection/Human-Connection/issues/305
- [ ] Upvote (Not necessary for migration! May should not be possible on old Alpha Posts!? Ask @Tirokk ) https://github.com/Human-Connection/Human-Connection/issues/305
- [X] Text
- [ ] Reply https://github.com/Human-Connection/Human-Connection/issues/305
- [x] Write Comment https://github.com/Human-Connection/Human-Connection/issues/305

Side Menu:
- [X] Post
- [X] Post - Hide unimplemented Links "Let's Talk" and "Versus" https://github.com/Human-Connection/Human-Connection/issues/386
- [ ] More Info https://github.com/Human-Connection/Human-Connection/issues/393
- [ ] Take Action https://github.com/Human-Connection/Human-Connection/issues/394
- [X] Sticky Menu - always visible https://github.com/Human-Connection/Human-Connection/issues/385
- [x] Scroll on Menu click https://github.com/Human-Connection/Human-Connection/issues/436

Others:
- [X] Design differences https://github.com/Human-Connection/Human-Connection/issues/392
- [ ] Repost a post that your followers and friends can see it #1374

### Userstory
```
Feature: Posts
  As a user
  I want to edit posts
  In order to share my personal interests and ideas with others.
```

### Design & Layout - Alpha

Post with Picture:
![image](https://user-images.githubusercontent.com/1238238/55485194-3d62a180-562a-11e9-98f3-de194526d99b.png)
[...]

PR #475, opened 2019-04-16
Add Comment Form

### LINKED ISSUE #260: Add form to write a comment
In order to write a comment, we need a `<comment-form>` component in the frontend. This comment form should create or update a comment, send it via `apollo` to the backend and update the UI.

mattwr18:
- to PagesPost_id_slugIndex.vue
- handle simple comment creation using auto-generated resolver

## Pullrequest

This adds a our Editor to an individual Post's page to allow adding Comments to the post
### Issues

- fixes #260 
- relates #305 

### Checklist

- [x] None

### How2Test

- [X] None

### Todo

- [X] None


mattwr18:
Currently looks like this:
Screenshot at 2019-04-16 15:55:11

ulfgebhardt:
I see you use the HcEditor component - awesome - once we improve it it will affect all occurrences like on create Post, edit Post & make/edit comment

mattwr18:
@roschaefer, @ulfgebhardt 
I tried unsuccessfully to add the Comment, and add it's relation in the same custom resolver :disappointed: 

What I was experiencing is that if I tried calling the mutation `CreateComment` with the `postId`, but without an id for comment, the comment was not created here `const comment = await neo4jgraphql(object, params, context, resolveInfo, true)`, even though that's it seems to say in the docs one would be generated. https://grandstack.io/docs/neo4j-graphql-js.html
```
If an ID typed field is specified in the type defintion, but not provided when the create mutation is executed then a random UUID will be generated and stored in the database.
```
and hard coding a comment id before a comment is created is not possible.
I am not happy with the current implementation because after a user comments and clicks on the submit button, the `CreateComment` is called, if that succeeds, then the `AddPostComments` is called to add a relationship between the `Post` and the `Comment`

backend/src/resolvers/comments.js
```diff
@@ -0,0 +1,12 @@
+import { neo4jgraphql } from 'neo4j-graphql-js'
+
+export default {
+  Mutation: {
+    CreateComment: (object, params, context, resolveInfo) => {
+      return neo4jgraphql(object, params, context, resolveInfo, false)
+    },
+    AddPostComments: (object, params, context, resolveInfo) => {
```
roschaefer: Please do not create a Comment without a Post. The frontend should assign the Post of the Comment already during creation. We don't want any left-over state, ie. comments without authors or posts.
```suggestion
```

backend/src/resolvers/comments.spec.js
```diff
@@ -0,0 +1,62 @@
...
+})
+
+afterEach(async () => {
+  await factory.cleanDatabase()
+})
+
+describe('CreateComment', () => {
+  const mutation = `
+    mutation($id: ID!, $content: String!) {
+      CreateComment(id: $id, content: $content) {
```
roschaefer: Adding the post here might get tricky. The author-to-post relationship gets created via request parameters, while comment-to-post needs to be created through additional input params.

backend/src/seed/factories/comments.js
```diff
@@ -7,19 +7,15 @@ export default function (params) {
     content = [
       faker.lorem.sentence(),
       faker.lorem.sentence()
-    ].join('. '),
-    disabled = false,
-    deleted = false
```
roschaefer: Nice! :+1:

backend/src/resolvers/comments.js
mattwr18: I do not want to leave it like this @roschaefer, which is why I said :point_up: I am not happy with it, but I spent a couple hours last night and could not seem to get it working, and there are no examples of creating any nodes passing in additional information.

We have a few where we assign an author, but that does not require passing in any additional info as we can access the user id with the context. 

When I try to modify the resolver, I run into the issues stated :point_up:

mattwr18:
Currently, the apollo store is failing to update, or `readQuery` with this
Screenshot at 2019-04-18 22:18:21

mattwr18:
Currently, looks like this
Screenshot at 2019-04-19 12:47:06

ulfgebhardt:
image
image

Awesome @mattwr18 this goes into the right direction. I posted the target Design above. Thats the current Alpha - things are not set in stone, but can be used to orientate ;-)
For more look here: https://github.com/Human-Connection/Human-Connection/issues/382

The clickdummy shows this:
image
image

Clickdummy can be found [here](https://preview.uxpin.com/24a2ab8adcd84f9a763d87ed27251351225e0ecd#/pages/99768901/simulate/no-panels?mode=i)

mattwr18:
I said a I would post here about the way we are updating the comments list after a successfully created `Comment`. At the moment we are updating it by doing a `refetch()` of the `Post`, but this takes a bit of time and is maybe not an ideal way of achieving the desired results(?)

Note, this is how we are currently updating the `hc-follow-button` here https://github.com/Human-Connection/Human-Connection/blob/master/webapp/pages/profile/_id/_slug.vue#L408, but @appinteractive noted that a better way of dealing with this is `apollo subscriptions`. 

Is this something we want to set up in this ticket? It seems like it could be a separate ticket(?)
Here is some information I found in the Apollo docs https://www.apollographql.com/docs/react/advanced/subscriptions... 

It states `In most cases, intermittent polling or manual refetching are actually the best way to keep your client up to date.`, but then goes on to list a few reasons why one would want to use them, which might be relevant in our use case(?)

@ulfgebhardt do you have experience working with `subscriptions`? 
what do you think @roschaefer, @appinteractive?

mattwr18:
@ulfgebhardt, I was going off the design from the ticket https://github.com/Human-Connection/Human-Connection/issues/260

Are you suggesting that is not the design, I should be going off of?

appinteractive:
I would separate the loading of comments from the post itself. So do two requests, we need that anyway as we will have to pagination the comments. 

Also I saw that you removed the async data part, that can lead to flashing of first time rendering the site and downgrades the search engine rankings (see Server Side Rendering)

roschaefer:
@mattwr18 contrary to @appinteractive I would suggest to go step by step. I believe that we can write tests and make them pass without having subscriptions and pagination, is that correct @mattwr18 ?

Sometimes it can be a bit overwhelming to get everything right from the beginning.

mattwr18:
@appinteractive, thanks for the suggestions, I didn't know that removing the async data part would have negative impacts. Maybe there is a way to accomplish both things.
I would like to be able to extract it in `/graphql`, so that it can be reused, and clean up the `webapp/pages/post/_id/_slug/index.vue` file since that query is soo long. 

I like the idea of having a different request for the comments, but I had difficulty requesting the Post by its slug, with just the comments, since it returns the same key `Post`. 

Yes, @roschaefer we can implement this without subscriptions/pagination, I didn't interpret what @appinteractive said in his review to mean adding those at this point, just to make a separate request for the comments.

I have implemented a custom query, which is working, and also added back asyncData. I would still like to extract this to its own file, but that is a refactoring ticket.

mattwr18:
>  oh, this looks very good! @mattwr18 well done!
> 
> Just one question out of curiosity: What let you implement the custom GraphQL query `CommentByPost` instead of using
> 
> ```
> query($id: ID!) {  
>   Post(id: $id) {
>     comments { ... }
>   }
> }
> ```

I commented about that :point_up: 
"... I had difficulty requesting the Post by its slug, with just the comments, since it returns the same key Post."

Tirokk:
Hey, looks like a very good start @mattwr18 , @aonomike !

Functional problems still exist, but could be solved by a later issue:

- After sending the first comment the text in the comment field is not deleted or field is not updated.
- The comment counter does not count, if I send a comment.
- Submit comment button is not disabled, if text is smaller a 3 characters.
- Comments appear in the wrong order under the post: Last as the topmost?

Later issues:

- Comments should be deleted or edited.
- Mentions with `@`should be working.

mattwr18:
great thanks @Tirokk, almost all if not all stem from the use of the Editor, which leverages `tiptap`

This is also failing the cypress tests because of the validations we have added this morning. for some reason it is sending an empty string to the backend, instead of sending the text.

"After sending the first comment the text in the comment field is not deleted or field is not updated."
this is what I pointed out this morning and also occurs when one clicks the cancel button, then types something, and then submits a comment or clicks the cancel button again. This is to due with the editor, as it doesn't occur on a `ds-input` field.

"The comment counter does not count, if I send a comment."
This is due to the fact that we are just refetching the comments, instead of refetching the entire post, which contains the commentCount attribute. We could go back to refetching the same Post, or maybe have a function that queries for the Post with only the comments and commentCount, but I had difficulty getting this working and commented on the issue above.

"Submit comment button is not disabled, if text is smaller a 3 characters."
This is because we are not able to validate the Editor in the frontend, our validation schema only works for `ds-input` fields... we could write our own custom validations though

"Comments appear in the wrong order under the post: Last as the topmost?"
I think this is working for me(as was intended), it has the latest comments at the top

"Comments should be deleted or edited."
yes, this is a separate ticket

"Mentions with @should be working."
great catch, have pushed up the code for them to work

mattwr18:
- @Tirokk, was able to properly clear the Editor, finally, haha so gotten rid of the first bug

- Also, updating the commentCount using `comments.length` since it's an array

- Added validations to disable the buttons if there aren't more than 3 characters.

- As I said, I think the order of the comments is what is intended, and got the mentions working as well.

What else you got??

@roschaefer I had another go at querying the Post with its comments in a separate function and couldn't get it working, but maybe I am doing something silly incorrect.

cypress tests are still failing for the reasons stated :point_up:, there are other cypress tests like `WritePost.feature`, which I am using the same steps even, but for some reason the `this.form.content` is not updating, though it is if one manually does it.

Tirokk:
@mattwr18 super that you could fix some of the issues already !!! 👏🏼

The comment order we have now, last comment comes first, is the opposite of the Alpha, last comment comes last. So the question is, what we gonna do? @roschaefer @appinteractive @ulfgebhardt 

We have a pairing in the afternoon and have a look on query comments of a post. @mattwr18

Tirokk:
Functional problems still exists:

- Before you type a comment, input field empty, the `Cancel`and `Submit Comment` buttons are enabled.

Everything else seems to work perfect.
Wonderful job @mattwr18 !!! 🚀

mattwr18:
Screenshot at 2019-04-25 16:43:08

This is what it's looking like these days

Tirokk:
A cypress test is failing @mattwr18 .

Bildschirmfoto 2019-04-26 um 16 00 02

mattwr18:
I saw that, I think it might not have to do with our latest changes, but rather the blocker wasn't actually fixed?

mattwr18:
Right, no I found the error... If you click on the step 24 `CLICK`,  you will see that it clicks on the `Comment` header next to the comment counter... this is because I am leveraging an existing step in `steps.js`, which is reusable for links and buttons, so the person who wrote it did not specify `cy.get('button')`, but rather `cy.contains(linkOrButton)`

Tirokk:
Yai, you got it @mattwr18 ! 🚀😻
Test and everything works fine with me.

Lulalaby:
Travis / Cypress seems to have problems with other modules.
Do we ignore it?

mattwr18:
> Travis / Cypress seems to have problems with other modules.
> Do we ignore it?

We have some intermittent cypress tests, meaning that they will sometimes fails, sometimes pass.
This is a common issue unfortunately. With cucumber there is a way to tell it to run the failed tests a second time in isolation, and if it fails again, to fail the build. It would be interesting to research if there exists something like this with cypress/cypress-cucumber-preprocessor.

In the meantime, I just manually rerun the build.

mattwr18:
it looks like they don't have support for retries at this moment, but there are some people making it work in different ways https://github.com/cypress-io/cypress/issues/1313

there is a plugin on cypress website however, which allows this https://docs.cypress.io/plugins/index.html
https://github.com/Bkucera/cypress-plugin-retries

gonna add it and put in a `PR` so we can get a sense of how it works

roschaefer:
It seems I am the only person who re-runs Travis CI builds. Why?

@Lulalaby @mattwr18 @matthib @ulfgebhardt @appinteractive @ogerly @Tirokk could you all check if you see the button below?

Screenshot (87)

mattwr18:
that is not true @roschaefer, I rerun the builds associated with my `PR`s if they are failing in a way that I know to be intermittent, I reran at least three times on Saturday before fixing the intermittent failing test `ReportContent.feature`.

I also regularly look at others `PR`s and rerun them.

roschaefer:
@mattwr18 #superhero 
hero

mattwr18:
:smile_cat: 
I think you definitely do the lion share of it @roschaefer... and I completely understand why you'd want to highlight it, as well as provide a screenshot for others who might not be familiar with how to restart the build.
It seems that that `PR` to fix the `ReportContent` might be working, or at least we don't have evidence yet that it doesn't work. :smile_cat:

Issue #479, opened 2019-04-17
🐛 [Doc] Modify Documentation for use with Monorepo


Tirokk:
## :bug: Bugreport

Documentation is not updated for Monorepo.

Tirokk:
Working on it at PR #480 .

PR #480, opened 2019-04-17
[WIP][Doc] Modify Documentation for use with Monorepo

Tirokk:
## Pullrequest

Updates documentation to Monorepo.

### Issues
- relates #479

### Checklist

- [X] None

### Todo

- [X] None

### DIFF
--- a/backend/README.md
+++ b/backend/README.md
@@ -22,7 +22,7 @@ $ docker-compose exec neo4j migrate
 
 {% endtab %}
 
-{% tab title="Without Docker" %}
+{% tab title="Local" %}
 
 For the local installation you need a recent version of [node](https://nodejs.org/en/)
 (&gt;= `v10.12.0`) and [Neo4J](https://neo4j.com/) along with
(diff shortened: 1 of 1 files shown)


backend/README.md
```diff
@@ -22,7 +22,7 @@ $ docker-compose exec neo4j migrate
 
 {% endtab %}
 
-{% tab title="Without Docker" %}
+{% tab title="Local" %}
```
roschaefer:
```suggestion
{% tab title="Withouth Docker" %}
```
Yes, I was also tempted to call it that way first. Because we call it "local installation" when we mean "without docker", right? People who don't know docker won't understand what it means so I called it simply "without docker"

Tirokk:
Closed it, because it did not show up in GitBook.

roschaefer:
@Tirokk the reason why this branch did not show up on Gitbook is that the pattern is `doc*` not `*doc*`. So, if you would rename this branch to `doc-modi-for-monorepo` it would show up, but you named it `479-doc-modi-for-monorepo`.

Tirokk:
That I remembered after this first try @roschaefer . But why than my next tries `docs-test_modi_for_monorepo` and `docs-modi_for_monorepo_479` did not show up until yet?

PR #484, opened 2019-04-18
Fix? flaky cypress by waiting for <no-ssr> content

roschaefer:
Content menu is a client-side rendered component. Maybe an extra
assertion will wait for it? That's how you can flix flaky cucumbers if
you use capybara.

### Issues

- [X] None

### Checklist

- [X] None

### How2Test

- [X] None

### Todo

- [X] None


ulfgebhardt:
image

I approved this - it is for this test? You do not mention which flanky test you are fixing - from the change it seems to be an menu button -> is it the report user test?

roschaefer:
Yes, it's the `ReportUser.feature` test. Unfortunately, that does not help as you can see. The build server is still failing for that reason.

If my assumption is correct, it might fail because of the `<no-ssr>` around the content menu. Another try would be to check why it's not rendered on the server and to check what is so different for the content menu component on the other pages? I mean, it's always failing for the user, but it does not fail on the post page or the comment page. Could you have a look @ulfgebhardt ?

ulfgebhardt:
I think it is because `ds-button-ghost`. This class seems to be added on runtime, since its not in the code.
image

if i am correct the button will show a bit later on the UI when the rest is loaded. I am not sure tho. This would explain `ds-button-ghost`.
Another possibility is that you want to press the button hidden behind the `dot-menu` - which is not visible till you press the dots. From Travis i cannot tell which button you are trying to click.

If you want the normal user way:
- navigate to correct page
- trigger `dot.menu`
- wait till all entries appeared
- click report
- continue with test...

Or just `force` the click?

roschaefer:
@ulfgebhardt great research! :+1: Using `{ force: true }` unfortunately does not help, I tried that earlier.

Could you see what happens if you remove `<no-ssr>`? Sometimes it's used without any reason.

mattwr18:
@roschaefer @ulfgebhardt, I think I fixed this flaky test by adding a class to the `content-menu`... it would be nice to get it working without, but if this can save us from needing to consistently rerun our builds, I feel like it's a smal price to pay.
I needed to rerun the build three or four times just today to get my feature all green :open_mouth: 
Not sure if this will work every time, but it passed both builds on the first try on this branch

roschaefer:
@mattwr18 adding one more css class is no problem at all. If that solves all our problems, I would be interested in *why* the class fixes the flaky test.. :thinking:  Do you have an explanation?

mattwr18:
no, I don't have an explanation as of yet, I kinda stumbled into this, and I'm not sure if it actually solves the issue... I'm hesitant to declare it a win yet.... As the build for the CreateComment has failed again cause of this, I have pulled in the latest changes and will see if it passes the first try.

I got to this by adding a bunch of `get` and `contains` that would wait till the post info had been brought in after reading cypress issues to do with the page refreshing while the cypress test was trying to execute the failing step. This caused the issue that there were two `.content-menu-trigger` classes, one for the user, one for the post. When I added the class it started reliably passing locally, which is when I started removing code to see if it would still pass without the extra `get` `contains`.

PR #489, opened 2019-04-19
[TryOut] First approach of mutation with filter parameter

Tirokk:
## Pullrequest
Dont Merge, just a study !!!

### Issues

- [X] None

### Checklist

- [X] None

### How2Test

- [X] None

### Todo

- [X] None

### DIFF
--- a/backend/src/schema.graphql
+++ b/backend/src/schema.graphql
@@ -176,10 +176,13 @@ type Post {
   image: String
   visibility: VisibilityEnum
   deleted: Boolean
-  disabled: Boolean
+#  disabled: Boolean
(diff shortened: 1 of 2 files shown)


backend/src/schema.graphql
```diff
@@ -207,6 +210,10 @@ type Post {
   )
 }
 
+input _PostFilter {
```
roschaefer: We don't have to redefine the `_PostFilter` on our side, that's done by the library `neo4j-graphql-js` when we augment the schema.

roschaefer: @Tirokk the library will only add types and inputs etc. if they haven't been defined yet, so defining it on our side will effectively disable this behaviour.

roschaefer:
I guess this can be closed after we figured out how to use filters in #754 .

PR #499, opened 2019-04-23
296 image component

### LINKED ISSUE #296: Refactoring: Prefix URLs pointing to the backend
**Describe the bug**
The server side rendered frontend serves our images at `/api/uploads`. All requests to `/api` are proxied to the backend, which serves images at `/uploads/`. Right now, the backend mutates all URLs in the response to be prefixed with `/api`. This is a design flaw as the backend *knows* how the frontend proxies requests.

**Expected behavior**
We should prefix all `` urls **in the frontend** with `/api` so the backend does not need to do it for the frontend.

**Screenshots**
2019-02-05-094935_1920x1080_scrot

roschaefer:
## Pullrequest
This PR adds a custom `` component which we can use throughout the webapp. The webapp proxies all requests to uploads to the backend, therefore we can assume that any relative image src files come from the backend. The component checks if the image `src` attribute is a relative URL and if yes, it prefixes it with `/api` which is the proxied path to the backend.

### Issues

- [X] fixes #296 

### Checklist

- [ ] @tansaku @aonomike could you finish this PR please? You are part of team [volunteers](https://github.com/orgs/Human-Connection/teams/volunteers/members) and you should have write permissions to this branch.

### How2Test

- [ ] `cd webapp && yarn run test components/Image/`

### Todo

- [ ] Implement `imageSrc` computed property
- [ ] Replace all `img` tags with this component


tansaku:
thanks @roschaefer :-)

@aonomike if you have a chance could you pull this branch out onto your machine (since I don't have a local instance working at the moment :-/ ) and see what you get running the new tests that @roschaefer has created?

roschaefer:
@appinteractive and I are in a call and he told me that the removed assets in this PR are newer than those currently in `/backend` folder. Please move the removed assets into folder `/backend`.

roschaefer:
@aonomike @tansaku what's the status on this? It's OK not to contribute on this PR, but please tell us so we can assign it to someone else.

tansaku:
thanks for this @roschaefer - pinpointing the issue with a unit test is very helpful, although I guess we will still not be able to check it is working in the full system without access to some legacy data, although I guess you are confident that if the unit test passes then everything will work at the high level?

webapp/components/Image/index.vue
```diff
@@ -0,0 +1,23 @@
...
+  props: {
+    imageProps: {
+      type: Object,
+      required: true
+    }
+  },
+  computed: {
+    imageSrc() {
+      // TODO implement
+      return 'this should be an image source attribute'
```
tansaku: so here we need something like:

```js
return  src.startsWith('/img') ? src.replace('/img','/api/img') : src
```
assuming that we can refer to the current `src` value here

roschaefer: No, check if it starts with `/`. I believe NodeJS URL module can tell you if it's absolute or not. For the browser, there should be a polyfill. See if it's possible: https://nodejs.org/api/url.html
If not, feel free to assume that a link starting with `/` is relative.

roschaefer:
@tansaku I am very confident. If you don't believe me, you can also place anything in folder `backend/public/` and create an image component in the webapp - somewhere. The backend already serves assets in the public folder.

webapp/components/Image/index.vue
tansaku: @roschaefer but isn't it the case that we only want to rewrite img links?  or you're saying we'd like to prefix *every* relative url with `/api`?  if so fair enough :-)

It doesn't look like NodeJS URL module has a check for that https://nodejs.org/api/url.html some good discussion of alternatives here: https://stackoverflow.com/questions/10687099/how-to-test-if-a-url-string-is-absolute-or-relative  

looks like [urijs](http://medialize.github.io/URI.js/docs.html#is) provides the functionality you're suggesting, and that's not in the project yet (although could be a good add), and that would give us code like:

```js
var uri = new URI(src);
return uri.is("relative") ? src.replace('^/','/api/') : src
```
@aonomike hope you're following all this :-)

roschaefer: @tansaku the component we're talking about is called `ImageComponent` so a developer who would use this component will use it only for images (hopefully :disappointed: ). We can assume that we have *all* our images at some `uploads` service (currently the backend serving `public`).

You *could* add another npm package to check absolute/relative but I think this is quite an overkill.  

@tansaku it seems like you understand the problem but don't want to provide the fix yourself, is that correct? Were you successful setting up the project? @aonomike is not that responsive these days so I would welcome if you could provide the commit yourself.

Issue #296 was created on February 5th and if this conversation takes any longer I'll probably take care of it myself.

tansaku: @roschaefer based on your great input I could now easily provide the fix - as you have inferred I'm holding off because I'm trying to provide a learning opportunity for @aonomike (and/or others) - hopefully I'll see him in our AV scrum today and we can take care of it together.

aonomike:
Hey @roschaefer sorry I have been missing these conversations on this platform. We have synched with @tansaku and @mattwr18 to make the tests pass. Was a great learning experience. Kindly let me know if there are other test cases we need to take into consideration. Thanks!

tansaku:
@aonomike looks like the build has failed with a linting error?

```
yarn run v1.13.0
$ eslint --ext .js,.vue .
/nitro-web/components/Image/index.vue
  19:52  error  Insert `·`  prettier/prettier
✖ 1 problem (1 error, 0 warnings)
  1 error and 0 warnings potentially fixable with the `--fix` option.
error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
The command "docker-compose exec webapp yarn run lint" exited with 1.
11.04s$ docker-compose exec webapp yarn run test --ci --verbose=false
```

@mattwr18 @roschaefer can you confirm my interpretation of the build failure?

Might it be an idea to have linting issues autofixed when tests run?

aonomike:
Thanks @tansaku I had not seen the build failure. I have fixed that and hopefully it passes 🤞

aonomike:
@roschaefer is replacing all  tags with this component in the scope of this PR?

webapp/components/Image/index.vue
```diff
@@ -0,0 +1,23 @@
...
+  props: {
+    imageProps: {
+      type: Object,
+      required: true
+    }
+  },
+  computed: {
+    imageSrc() {
+      const src = this.imageProps.src
+      return src.startsWith('/') ? src.replace('/', '/api/') : src
```
roschaefer: Excellent! :+1:

webapp/components/Badges.vue
```diff
@@ -10,22 +10,31 @@
       :key="badge.key"
       class="hc-badge-container"
     >
-      <img
+      <hc-image
         :title="badge.key"
-        :src="badge.icon"
+        :image-props="imageProps(badge.icon)"
```
roschaefer:
```suggestion
        :image-props="{src: badge.icon}"
```
@aonomike that does not work?

mattwr18: when I was coding with him @roschaefer this wasn't working because it is calling `this.imageProps` and that was throwing an error, but I haven't pulled it down and played around with it, so I could be off base

webapp/pages/logout.vue
```diff
@@ -36,8 +36,17 @@
...
 
 <script>
+import HcImage from '~/components/Image'
 export default {
+  components: {
+    HcImage
+  },
   layout: 'blank',
+  computed: {
+    imageProps() {
```
roschaefer: @aonomike this method got overlooked? Or is there a reason for keeping it in.

mattwr18:
the `console.error` you saw in `yarn test` was not introduced by this `PR`, has been discussed in others as well, and has been fixed(?) by https://github.com/Human-Connection/Human-Connection/pull/587

I also experienced the weird sizes yesterday while working on :point_up: and I believed talked about it in the `PR` asking for clarification from @appinteractive

ulfgebhardt:
The avatar thing is not introduced in this PR - for all I know - therefore I gonna merge this

roschaefer:
@ulfgebhardt please delete branches when you merge PRs.

PR #535, opened 2019-05-02
Delete SocialMedia

### LINKED ISSUE #463: 🚀 Delete SocialMedia
## :rocket: Feature

As a User
I`d like to be able to delete my social media accounts
In order to manage my social media account

### Is your feature request related to a problem? Please describe.

I believe this is needed to make us GDPR compliant(?)
This is a basic feature that if you can add something to you profile, you should be able to remove it.
### Describe the preferred solution and alternatives you've considered

I believe we don't have other alternatives.

### Design & Layout

A button on `webapp/pages/settings/my-social-media.spec.js` maybe that would send a message to the backend.

### Additional context

Add a mutation to `backend/src/resolvers/socialMedia.js`

Tirokk:
## Pullrequest

Social Media links have to be deleted sometimes.

### Issues

- fixes #463

### Checklist

Backend:
- [x] GraphQL mutation with tests
Webapp:
- [x] Trash symbol
- Modal dialog (opened a new issue #571 because of foreseen code usage)
- [x] Delete the Social Media with mutation

#### Backend

Run …
```
$ docker-compose exec backend yarn run test:jest backend/src/resolvers/socialMedia.spec.js
```

#### Webapp

- [x] Test the Webapp yourself
- [x] Write Frontend test


backend/src/resolvers/socialMedia.js
```diff
@@ -15,6 +18,11 @@ export default {
       )
       session.close()
 
+      return socialMedia
+    },
+    DeleteSocialMedia: async (object, params, context, resolveInfo) => {
+      const socialMedia = await neo4jgraphql(object, params, context, resolveInfo, false)
```
mattwr18: doesn't this create another `SocialMedia` node @Tirokk, or is there some magic that sees that you have called a delete action?

backend/src/resolvers/socialMedia.spec.js
```diff
@@ -30,20 +39,63 @@ describe('CreateSocialMedia', () => {
...
+      ).resolves.toEqual(expect.objectContaining({
+        CreateSocialMedia: {
+          id: expect.any(String),
+          url: 'http://nsosp.org'
+        }
+      }))
+    })
+
+    it('deletes social media', async () => {
+      const variablesC = { url: 'http://nsosp.org' }
```
mattwr18: maybe, we could find a more descriptive name? 
what about `deletionVariables`

Tirokk: Thank you @mattwr18 for the hint. I have done.

backend/src/resolvers/socialMedia.js
Tirokk: There seems to be a magic that deletes the node and with it the relation. @mattwr18 
Because it’s an autogenerated mutation.

ulfgebhardt:
@Tirokk I approved this - the coverage test on backend is still failing with is probably due to the jest setup at backend - if you cannot raise coverage i suggest to make the backend coverage test optional for the time beeing @roschaefer

backend/src/resolvers/socialMedia.js
roschaefer: @mattwr18 you could inspect the generated cypher query of `neo4jgraphql` to get evidence, but I assume it's using `DELETE DETACH`. https://www.graphgrid.com/understanding-detach-delete-in-cypher/

Quite a final solution for a node and it's relationship. Comparable to Rails' `:dependent => :destroy`.

PR #537, opened 2019-05-02
[WIP] Propose a hack to fix performance issue

roschaefer:
@appinteractive the solution is to move the `ORDER BY` statement right
after the first match and before the `RETURN`. We can chain cypher
statements with `WITH`.

I guess this is sth. we should tell the maintainers of
`neo4j-graphql-js`, no?

### Issues

- [X] None

### Checklist

- [X] None

### How2Test

- [X] None

### Todo

- [X] None

### DIFF
--- a/backend/src/resolvers/posts.js
+++ b/backend/src/resolvers/posts.js
@@ -1,6 +1,43 @@
-import { neo4jgraphql } from 'neo4j-graphql-js'
+import { neo4jgraphql, cypherQuery } from 'neo4j-graphql-js'
+import { v1 as neo4j } from 'neo4j-driver'
+import _ from 'lodash'
+
+export function extractQueryResult ({ records }, returnType) {
+  const variableName = 'post'
+  let result = null
+  result = records.map(record => record.get(variableName))
+  // handle Integer fields
+  result = _.cloneDeepWith(result, field => {
+    if (neo4j.isInt(field)) {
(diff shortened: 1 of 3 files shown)


backend/src/resolvers/posts.js
```diff
@@ -1,6 +1,43 @@
-import { neo4jgraphql } from 'neo4j-graphql-js'
+import { neo4jgraphql, cypherQuery } from 'neo4j-graphql-js'
+import { v1 as neo4j } from 'neo4j-driver'
+import _ from 'lodash'
+
+export function extractQueryResult ({ records }, returnType) {
```
roschaefer: All of this is copy+paste from `neo4j-graphql-js`.

backend/src/resolvers/posts.js
```diff
@@ -1,6 +1,43 @@
...
+  return result
+}
 
 export default {
+  Query: {
+    Post: async (object, params, context, resolveInfo) => {
+      let [query, cypherParams] = cypherQuery(params, context, resolveInfo)
+
+      // HACK-ATTACK
+      query = query.replace('RETURN', 'WITH `post` ORDER BY `post`.createdAt DESC RETURN')
```
roschaefer: Here's the hack :neckbeard:

appinteractive: Let’s add a `// TODO: unshack hack-attack` 🦹🏼‍♂️🤣

PR #541, opened 2019-05-03
Ap refactoring

### LINKED ISSUE #221: Use correct Id's
Now the activity id's are used wrong, so that the url is pointing to nowhere.

In the ActivityPub middleware (exactly the CreatePost Resolver) change the activityId and objectId of the created  activity to use the generated ids's directly.

**Describe alternatives you've considered**
My opinion is to create the activity after the call of resolve, because the resolve mutates the database. If that throws an error the activity do not need to be created (Performance).

### LINKED ISSUE #398: 🐛 False negative cucumber test: CreatePost not authenticated
## 🐛 Bugreport
The current implementation of `activityPub.handleCreate` is using a [hard-coded token for authentication](https://github.com/Human-Connection/Human-Connection/blob/master/backend/src/activitypub/NitroDataSource.js#L38). Apparently it's a token for a user who is in the seeds.

The problem with this is:
* All posts will be authored by that user
* If there is no user for that token in the DB, creating posts will fail because it's not authenticated

The latter problem is the reason why `yarn run test:cucumber` fails if you don't have seeds in the database.

1. `yarn run db:reset`
2. `yarn run test:cucumber`
3. Fails with "✖ And the post with id "kljsdfg9843jknsdf" to be created"

### Expected behavior
Cucumber tests should ass

### Additional context
```
2) Scenario: Send an article to a user inbox and make sure it's added to the inbox # features/object-article.feature:11
   ✔ Given our own server runs at "http://localhost:4123" # features/support/steps.js:23
   ✔ And we have the following users in our database: # features/support/steps.js:27
       | Slug   |
       | marvin |
       | max    |
   ✔ When I send a POST request with the following activity to "/activitypub/users/max/inbox": # features/support/steps.js:41
       """
       {
         "@context": "https://www.w3.org/ns/activitystreams",
         "id": "https://aronda.org/users/marvin/status/lka7dfzkjn2398hsfd",
         "type": "Create",
         "actor": "https://aronda.org/users/marvin",
         "object": {
             "id": "https://aronda.org/users/marvin/status/kljsdfg9843jknsdf",
             "type": "Article",
             "published": "2019-02-07T19:37:55.002Z",
             "attributedTo": "https://aronda.org/users/marvin",
             "content": "Hi Max, how are you?",
             "to": "as:Public"
         }
       }
       """
   ✔ Then I expect the status code to be 200 # features/support/steps.js:91
   ✖ And the post with id "kljsdfg9843jknsdf" to be created # features/support/steps.js:113
       Error: function timed out, ensure the promise resolves within 5000 milliseconds
           at Timeout._time.default.setTimeout [as _onTimeout] (/home/robert/Development/human-connection/Human-Connection/backend/node_modules/cucumber/lib/user_code_runner.js:81:20)
           at ontimeout (timers.js:436:11)
           at tryOnTimeout (timers.js:300:5)
           at listOnTimeout (timers.js:263:5)
           at Timer.processTimers (timers.js:223:10)

```

### LINKED ISSUE #413: 🚀 Use token on demand
## 🚀 Feature
Use a token on demand per user and request instead of one token for all request.

### Describe the prefered solution and alternatives you've considered
1. Create a user directly with a cypher query instead of using the GraphQL API
-> add setToken function (NitroDataSource)
2. Generate a token for each request
-> use this token to create a post

Mastercuber:
## Pullrequest
This Pullrequest refactors some files of the ActivityPub component. In addition it add the token on demand funtionality, fixes the problem with using the wrong id and also fixes the false negative creation of a post object test.

### Issues
- fixes #221
- fixes #398 
- closes #413

### Checklist
- [X] None

### How2Test
- [X] None

### Todo
- [X] None


Mastercuber:
Sorry for, again, a big PR. I will reduce the size in the future. And push things up at the end of the day.

backend/test/features/activity-delete.feature
```diff
@@ -5,51 +5,42 @@ Feature: Delete an object
...
     """
 
   Scenario: Deleting a post (Article Object)
-    When I send a POST request with the following activity to "/activitypub/users/bernd-das-brot/inbox":
+    When I send a POST request with the following activity to "/inbox":
     """
     {
       "@context": "https://www.w3.org/ns/activitystreams",
-      "id": "http://localhost:4123/activitypub/users/karl-heinz/status/a4DJ2afdg323v32641vna42lkj685kasd2",
+      "id": "https://localhost:4123/api/users/spongebob/status/kljsddgnffgdgsx9bv8adg43jfdgkgnsdf234s",
```
roschaefer: Why `https` here? @Mastercuber :thinking:

backend/test/features/activity-delete.feature
```diff
@@ -5,51 +5,42 @@ Feature: Delete an object
     Given our own server runs at "http://localhost:4123"
     And we have the following users in our database:
       | Slug          |
-      | bernd-das-brot|
-    And I send a POST request with the following activity to "/activitypub/users/bernd-das-brot/inbox":
+      | spongebob     |
+    And I send a POST request with the following activity to "/inbox":
```
roschaefer:
```suggestion
    And we have the following posts in our database:
```
Create a separate step definition for test setup. In this scenario I think only the `id` is relevant, no? Maybe content, too, but certainly not the entire activity json.

backend/test/features/activity-like.feature
```diff
@@ -1,42 +1,62 @@
...
+      "actor": "http://localhost:4123/users/karl-heinz",
+      "object": "http://localhost:4123/api/users/theodor/status/dkaaadsfljsdfaaaffg9843jknsdf"
     }
     """
     Then I expect the status code to be 200
-    And the post with id "dkasfljsdfaafg9843jknsdf" has been liked by "peter-lustiger"
+    And the post with id "dkaaadsfljsdfaaaffg9843jknsdf" has been liked by "karl-heinz"
+
+  Scenario: Send an Undo of a previous like activity to a users inbox and make sure the object has been disliked
+    When I send a POST request with the following activity to "/users/theodor/inbox":
```
roschaefer:
```suggestion
    Given the activity "http://localhost:4123/api/users/theodor/status/dkaaadsfljsdfaaaffg9843jknsdf" has been liked by "karl-heinz"
    When I send a POST request with the following activity to "/users/theodor/inbox":
```

In this scenario, you never made sure that the activity was liked. Please also use better ids instead of "dkaaadsfljsdfaaaffg9843jknsdf". You could also refactor the `Background` in this scenario.

backend/test/features/activity-like.feature
```diff
@@ -1,42 +1,62 @@
...
     """
     {
       "@context": "https://www.w3.org/ns/activitystreams",
-      "id": "http://localhost:4123/activitypub/users/peter-lustiger/status/83J23549sda1k72fsa4567na42312455kad83",
+      "id": "http://localhost:4123/api/users/karl-heinz/status/83J23549sda1k72fsa4567na42312455kad83",
       "type": "Like",
-      "actor": "http://localhost:4123/activitypub/users/peter-lustiger",
-      "object": "http://localhost:4123/activitypub/users/karl-heinz/status/dkasfljsdfaafg9843jknsdf"
+      "actor": "http://localhost:4123/users/karl-heinz",
+      "object": "http://localhost:4123/api/users/theodor/status/dkaaadsfljsdfaaaffg9843jknsdf"
```
roschaefer: @Mastercuber how come that sometimes you're routing to `http://localhost:4123/` and sometimes to `http://localhost:4123/api/`?

backend/test/features/object-article.feature
```diff
@@ -5,26 +5,96 @@ Feature: Send and receive Articles
...
+        "content": "Content for article object creation!",
+        "to": "as:Public"
       }
     }
     """
     Then I expect the status code to be 200
-    And the post with id "kljsdfg9843jknsdf" to be created
+    And the post with id "kljsasddfg9843jknsdfasd" to be created
+
+  Scenario: Send an note to the shared inbox and make sure it's added to the correct inbox
```
roschaefer: Do we map a `Note` and an `Article` to a Post?

backend/test/features/support/steps.js
```diff
@@ -1,25 +1,37 @@
...
+      name: slug,
+      email: 'example@test.org',
+      password: '1234',
+      actorId: `http://localhost:3000/api/users/${slug}`
+    })
+  }
   // await login({ email: 'example@test.org', password: '1234' })
 }
 
+BeforeAll('Clean Database for fresh test setup', async function () {
```
roschaefer: Great! I did not know cucumber has `BeforeAll` similar to `Given` and `When` I thought we have to use hooks in `features/support/`.

backend/src/resolvers/posts.js
```diff
@@ -16,6 +16,26 @@ export default {
...
+      // first delete WROTE relationship and after the update add it again
+      // this is needed because otherwise there would be 2 WROTE relations in the database
+      await session.run(
+        'MATCH (u:User)-[rel:WROTE]->(p:Post) WHERE p.id = $id delete rel', {
+          id: params.id
+        }
+      )
+      const result = await neo4jgraphql(object, params, context, resolveInfo, false)
+      await session.run(
+        'MATCH (u:User), (p:Post) WHERE p.id = $id AND u.slug = $slug CREATE (u)-[:WROTE]->(p)', {
```
roschaefer: Three things:
1. If you use `MERGE` instead of `CREATE` that would not create duplicate relationships
2. Use `user.id` not `user.slug` if possible
3. If you setup relationships after you already created the object, use the object. In this case, use `result.id` not `params.id`

backend/src/middleware/slugify/uniqueSlug.js
```diff
@@ -13,3 +13,17 @@ export default async function uniqueSlug (string, isUnique) {
   } while (!await isUnique(uniqueSlug))
   return uniqueSlug
 }
+
+export function isUniqueFor (context, type) {
```
roschaefer: This is just a refactoring? :thinking: or do you use it in another place, too?

backend/src/middleware/fixImageUrlsMiddleware.spec.js
```diff
@@ -20,11 +20,16 @@ describe('fixImageURLs', () => {
...
+    })
   })
 
   describe('some string', () => {
-    it('returns untouched', () => {})
-    const string = 'Yeah I\'m a String'
-    expect(fixImageURLs(string)).toEqual(string)
+    it('returns untouched', () => {
+      const string = 'Yeah I\'m a String'
+      expect(fixImageURLs(string)).toEqual(string)
```
roschaefer: Great catch! @Mastercuber :+1:

backend/test/features/activity-like.feature
Mastercuber: I've tried to fit the frontend url expectations (with `/api` prefixed), though the backend is tested. It's more for code understandability from the view of the client.

backend/src/activitypub/ActivityPub.js
```diff
@@ -12,102 +12,99 @@ import request from 'request'
...
+      return
     }
+
+    // start standalone express server
+    dotenv.config()
+
+    const app = express()
+    app.set('ap', activityPub)
+    app.use(router)
+    app.listen(process.env.ACTIVITYPUB_PORT || 4010, () => {
```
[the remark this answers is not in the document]
Mastercuber: problems while trying out something. Would change it back to 4000 if necessary.

backend/test/features/activity-delete.feature
Mastercuber: I used https where I can. For some tests the service sends a request where the actual path, instead of the frontend prefix, is used. Also the response of the server is created with the `/api` prefix and the https protocol. To use the frontend port in the tests would need changes just for testing.

backend/test/features/object-article.feature
Mastercuber: Yes both are mapped to a Post.

backend/test/features/support/steps.js
Mastercuber: Yes, it's very handy.

backend/test/features/activity-delete.feature
roschaefer: Can you use `http` every time? We can assume that all our local web servers run locally, no `https` required. For testing, this is :ok:

roschaefer:
@mastercuber could you merge `origin/master` and resolve conflicts? I tried to do that and it's not obvious to me at all how to do the resolution. That would be great :+1: I think we can merge the PR next sprint, once the conflicts are gone.

roschaefer:
@Mastercuber we decided to have a different way to setup data in the backend: Instead of relying on `neo4j-graphql-js` which is hard to customize, we implement our own mutation resolvers and use [neode](https://github.com/adam-cowley/neode/tree/master/src) for making database calls and for data validations. So I guess it would be really a lot of work to resolve the merge conflicts.

Nevertheless I would like to start over with you, are you available at some point? We can make a pair-programming if you like. I am still very interested to get the activity pub implementation done.

Issue #410, opened 2019-04-04
🚀 [Feature] UserSettings - Security - Password Strength


ulfgebhardt:
### UserSettings - Security - Password Strength
![image](https://user-images.githubusercontent.com/1238238/55571293-8fc4c080-5705-11e9-9959-bf209d2fbf96.png)

- [x] Display Password strength
- [x] Check wether password is strong enought @roschaefer @appinteractive Rule?

### Design & Layout - Alpha

![image](https://user-images.githubusercontent.com/1238238/55571298-90f5ed80-5705-11e9-8929-9ea798274876.png)

### Design & Layout - Nitro

![image](https://user-images.githubusercontent.com/1238238/55571264-7cb1f080-5705-11e9-9296-f86bd88234c9.png)

Issue #247, opened 2019-03-25
[User Profile >> Settings >> Security] Change Password Cross-validations


roschaefer:
**Is your feature request related to a problem? Please describe.**
When your new password and the password confirmation field mismatch, you don't see an error. Also, the validation error messages are not translated.

**Describe the solution you'd like**
Once https://github.com/Human-Connection/Nitro-Styleguide/issues/46 is fixed, we can do cross-validations between form input fields. We even have a component tests in place already [here](https://github.com/Human-Connection/Human-Connection/blob/master/webapp/components/ChangePassword.spec.js#L64) describing the desired behaviour. These tests are currently `skipped`.

**Describe alternatives you've considered**
We could alternatively remove the confirm password field and wait for the backend to send validation errors to the client. But the user experience is better if we validate on the client.

**Additional context**
Here is a deep link: https://nitro-staging.human-connection.org/settings/security

![Screenshot (44)](https://user-images.githubusercontent.com/2110676/54938624-fe946380-4f26-11e9-9e2f-ffd809dc3abd.png)

PR #551, opened 2019-05-06
2019/kw15/change password strength

ulfgebhardt:
## Pullrequest

image

image

image

image

image

### Issues

- [X] requires https://github.com/Human-Connection/Nitro-Styleguide/issues/46
- [X] fixes https://github.com/Human-Connection/Human-Connection/issues/410
- [X] fixes https://github.com/Human-Connection/Human-Connection/issues/247

### Todo

- [x] There is console errors since i updated the styleguide on unrelated components (avatar) - those components are also displayed in a strange way. Please someone find out whats the Problem there
- [x] Is this the fault of this PR? Please confirm on current staging

image

image

image


webapp/components/Password/Change.spec.js
```diff
@@ -140,11 +142,16 @@ describe('ChangePassword.vue', () => {
...
 
-          it('displays error message', () => {
+          it('displays error message', async () => {
+            await wrapper.find('form').trigger('submit')
+            await wrapper.find('form').trigger('submit')
+            await mocks.$apollo.mutate
+
             expect(mocks.$toast.error).toHaveBeenCalledWith('Ouch!')
           })
         })
```
mattwr18: I updated this test as it started failing @roschaefer, but I am not happy with this cause it feels hacky
I was unable to get it to call the `mocks.$toast.error` without 2 submits and a wait for the `mocks.$apollo.mutate`
I think we should remove this test as it is flaky, and test it with cypress
Unlike the `CommentForm` component, I think this would be easy to test because all one needs to do is like I did here where you try to set you newPassword to the same as your oldPassword and expect to see a error toaster.
What do you'all think?

webapp/components/Password/Strength.vue
```diff
@@ -0,0 +1,139 @@
...
+    strengthClass() {
+      return `strength-${this.passwordStrength}`
+    }
+  },
+  watch: {
+    /**
+     * passwordStrength is the score calculated by zxcvbn
+     * @return {Number} Password Strength Score
+     */
+    passwordStrength() {
```
mattwr18: I moved this @ulfgebhardt because it was failing the linting
please see my commit message with a link to the SO post where I found the suggestion to move it to `watch`

ulfgebhardt: I Agree 100% - thanks for fixing this. As you can see i am still new to vue <3

webapp/components/Password/Change.spec.js
ulfgebhardt: I agree that calling `submit` twice is not the way we should solve this. But I certainly would like to have it in place. We certainly should not set the same password in the test since that is a valid option. (How would the server know its the same - its the users fault - we could certainly make a client-side check)

ulfgebhardt:
> this is great @ulfgebhardt
> do you think there is any place for cypress tests?
> I'm unsure how I feel about it

thanks @mattwr18 for doing this - much appreciated!
If you remove the same-password-test and its still green feel free to merge it. Else I will make sure to have more coverage.

ulfgebhardt:
@mattwr18  I removed the test case - but there is still problems with the UI. visible as console errors in the tests
image
It might be due to my updating the styleguide library - but i did not change anything regarding the component `ds-avatar`

mattwr18:
@ulfgebhardt, I have added an issue for this, as it is not part of the scope of this `PR`
and have found(?) the source of the issue and put in a `PR`
https://github.com/Human-Connection/Human-Connection/pull/587

PR #561, opened 2019-05-07
[WIP] [Don't Merge] Add tests to expose bug

### LINKED ISSUE #542: 🐛 Rapidly double clicking on `Comment` button causes two database writes
## :bug: Bugreport

At the moment, if you rapidly click on the `Comment` button, it is possible to create multiple comments with the same content.

Currently, in the `handleSubmit` method, the button is disabled and set to loading, but it is still possible to fire two events before the button is disabled and apollo is sending two requests.
Screenshot at 2019-05-05 12:33:22

1. visit `/`
2. click on a post
3. write a comment
4. click rapidly twice on the `Comment` button

### Expected behavior

There should be some way of handling this situation, if a user wants to comment the same thing twice, for whatever reason, they should need to type the same message/copy && paste it.

More often than not, this will happen by accident, and therefore we should have some protection against this.

have found this https://github.com/Akryum/vue-apollo/issues/185
 which suggests that adding `optimisticResponse` prevents two database writes https://github.com/Akryum/vue-apollo/issues/185#issuecomment-364696038

I've yet to successfully test this in favor of opening a new issue to explore options

### Version & Environment
 Type: []      
 - OS: []      
 - Browser: [] 
 - Version []  
 - Device: []

mattwr18:
- clicking on the "Comment" button twice

### Issues

- fixes #542

### DIFF
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -95,7 +95,7 @@ coverage:
         #  - master
         #flags:
         #  - integration
-        paths: 
+        paths:
           - backend/  # only include coverage in "backend/" folder
       webapp:  # declare a new status context "frontend"
         against: parent
@@ -127,7 +127,7 @@ coverage:
     #      - integration
     #    paths:
     #      - folder
-    
+
     #changes:
     #  default:
     #    against: parent
@@ -150,7 +150,7 @@ coverage:
 
   #ignore: # files and folders for processing
   #  - tests/*
-  
+
   #fixes:
   #  - "old_path::new_path"
 
--- a/backend/package.json
+++ b/backend/package.json
@@ -88,7 +88,7 @@
     "@babel/plugin-proposal-throw-expressions": "^7.2.0",
     "@babel/preset-env": "~7.4.4",
     "@babel/register": "~7.4.4",
-    "apollo-server-testing": "~2.4.8",
+    "apollo-server-testing": "~2.5.0",
     "babel-core": "~7.0.0-0",
     "babel-eslint": "~10.0.1",
     "babel-jest": "~24.8.0",
--- a/backend/package.json
+++ b/backend/package.json
@@ -58,7 +58,7 @@
     "dotenv": "~8.0.0",
     "express": "~4.16.4",
     "faker": "~4.1.0",
-    "graphql": "~14.2.1",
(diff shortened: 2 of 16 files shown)


mattwr18:
@ulfgebhardt @Tirokk @roschaefer 
In regards to this `PR`, I was trying to replicate the double creation of comments bug on https://nitro-staging.human-connection.org and was unable to.

I wonder if more time should be spent trying to fix this bug, if it appears to be only a problem in a dev env(?). If you all still thinks it's worth spending time on, I think I have fixed the bug by using lodash's `debounce`, but the unit tests are now broken, and need to be debugged.

So, repeating my question from `Discord`, how do people feel about the fix?

mattwr18:
ok, so then are you suggesting to debug the component tests and how to get them passing with lodash's `debounce`, or should we be only dealing with bugs reproducible on `nitro-staging`/`nitro-master`?

roschaefer:
@mattwr18 I don't even believe that this error only happens on `development`. What I mean is: I don't really like this workaround but it is :ok: for me. It feels like treating symptoms instead of treating the cause of the disease.

roschaefer:
@mattwr18 any updates on this?

mattwr18:
the update is that I am still unclear whether we should be adding code to deal with a bug that we haven't been able to reproduce in a deployed environment.
I guess maybe what you are saying is just merge in the latest changes and let's get it merged in anyways?

roschaefer:
@mattwr18 no, I would never merge in code without knowing what the effect is. E.g. I refer to unnecessary if-cases as "Angstcode" (fearful code). Code that seems to have a meaning but has no effect. It will confuse other developers and create anxiety in the team.

Can we simply close this PR then? Sorry for the time spent :slightly_frowning_face:

mattwr18:
that is what I would be in favor of @roschaefer... the work won't be lost if we ever need this solution, we can always reopen, or copy and paste the solution... it's just that I think I jumped the gun and created an issue that wasn't actually an issue for our user base.

what do you think @Tirokk?

mattwr18:
I believe we are all more or less in agreement on this @Tirokk @roschaefer 
and it's not that the decision to close this now will mean we can never use this code if we find that I am wrong and the user base does experience this issue...

PR #563, opened 2019-05-07
[WIP] Propose a hack to fix performance

roschaefer:
@appinteractive the solution is to move the `ORDER BY` statement right
after the first match and before the `RETURN`. We can chain cypher
statements with `WITH`.

I guess this is sth. we should tell the maintainers of
`neo4j-graphql-js`, no?

### Issues

- [X] None

### Checklist

- [X] None

### How2Test

- [X] None

### Todo

- [X] None

### DIFF
--- a/backend/src/graphql-schema.js
+++ b/backend/src/graphql-schema.js
@@ -24,7 +24,8 @@ export const resolvers = {
     ...statistics.Query,
     ...userManagement.Query,
     ...notifications.Query,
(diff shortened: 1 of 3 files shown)


roschaefer:
Now pointing towards `master` <<

roschaefer:
closed in favour of https://github.com/Human-Connection/Human-Connection/pull/684

PR #575, opened 2019-05-08
Revert tiptap upgrade

### LINKED ISSUE #572: Tiptap bug for 1.8.0
## :bug: Bugreport

upgrading tiptap introduces a bug, which has been reported here https://github.com/scrumpy/tiptap/issues/188

let's downgrade and follow the discussion there

1. $ cd `/webapp`
2.  run the tests with the `yarn test`

### Expected behavior

we should be able to have tests that interact with tiptap editor

mattwr18:
- it introduced a bug that caused failing tests
- I have found an open issue on the repo and commented what I experienced here https://github.com/scrumpy/tiptap/issues/188#issuecomment-490532736

### Issues

- fixes #572

### DIFF
--- a/webapp/package.json
+++ b/webapp/package.json
@@ -18,7 +18,10 @@
   "jest": {
     "verbose": true,
     "collectCoverage": true,
-    "coverageReporters": ["text", "lcov"],
+    "coverageReporters": [
+      "text",
+      "lcov"
+    ],
     "moduleFileExtensions": [
       "js",
       "json",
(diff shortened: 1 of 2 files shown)


mattwr18:
@roschaefer to downgrade it successfully, I needed(?) to run `rm yarn.lock`, but I know you had told me there is a better way to do this. 

I think in that context what I had done was to run `git checkout origin/master -- yarn.lock`, but in this context since the `yarn.lock` had a newer version it wasn't downgrading it.

The way I did it however has caused upgrades to several packages not covered in this `PR`, so if you have a better idea how I can just downgrade `tiptap` and `tiptap-extensions` in the yarn.lock, I can fix this `PR` and push it back up

roschaefer:
@mattwr18 no, I don't know a better way of downgrading an npm package. I would say `rm yarn.lock` is just fine.

I mean: It should work, right? In an ideal world, transitive dependencies should not have any effect :laughing:

PR #585, opened 2019-05-09
More descriptive minikube setup instructions

daumie:
## Pullrequest

Updates instructions on setting a local Kubernetes cluster with minikube.

### Issues

- [X] None

### Checklist

- [X] Provide clear instruction on minikube setup.

### How2Test

- [X] You should be able to easily setup   a local Kubernetes instance by following the [minikube setup](https://github.com/Human-Connection/Human-Connection/tree/master/deployment/minikube)

### Todo

- [X] Fix failing `nitro-neo4j` container.

### DIFF
--- a/deployment/human-connection/README.md
+++ b/deployment/human-connection/README.md
@@ -6,21 +6,21 @@ just apply our provided configuration files to your cluster.
 
 ## Configuration
 
-Copy our provided templates:
+Change into the `./deployment` directory and copy our provided templates:
 
 ```bash
 $ cp secrets.template.yaml human-connection/secrets.yaml
 $ cp configmap.template.yaml human-connection/configmap.yaml
 ```
 
-Change the `configmap.yaml` as needed, all variables will be available as
+Change the `configmap.yaml` in the `./deployment/human-connection` directory as needed, all variables will be available as
 environment variables in your deployed kubernetes pods.
 
 If you want to edit secrets, you have to `base64` encode them. See [kubernetes documentation](https://kubernetes.io/docs/concepts/configuration/secret/#creating-a-secret-manually).
 
 ```bash
 # example how to base64 a string:
-$ echo -n 'admin' | base64 --wrap 0
+$ echo -n 'admin' | base64
 YWRtaW4=
 ```
 


mattwr18:
I think we need to configure jest not to collect coverage from markdown files, cause we shouldn't have a drop in coverage because of doc updates
@ulfgebhardt

mattwr18:
Actually, I misdiagnosed the reason there is a drop in coverage @ulfgebhardt 
It seems like it is saying the coverage dropped for files not touched by this `PR`
:woman_shrugging:

deployment/human-connection/README.md
```diff
@@ -6,21 +6,21 @@ just apply our provided configuration files to your cluster.
...
 
-Change the `configmap.yaml` as needed, all variables will be available as
+Change the `configmap.yaml` in the `./deployment/human-connection` directory as needed, all variables will be available as
 environment variables in your deployed kubernetes pods.
 
 If you want to edit secrets, you have to `base64` encode them. See [kubernetes documentation](https://kubernetes.io/docs/concepts/configuration/secret/#creating-a-secret-manually).
 
 ```bash
 # example how to base64 a string:
-$ echo -n 'admin' | base64 --wrap 0
```
roschaefer: @daumie well, OK. It might be unnecessary for short strings, but `base64` will wrap long strings automatically. This might be undesired for certain use case, when a newline would break the integrity of, say, the content of a file.

roschaefer:
@ulfgebhardt could we merge this PR somehow?

deployment/human-connection/README.md
daumie: @roschaefer I tested this on Ubuntu 18.04 and it works. 
However, the option `wrap` is not available on MacOS.
Maybe we could be more explicit by specifying that this works on Ubuntu/Linux as opposed to MacOs.

daumie:
> @ulfgebhardt could we merge this PR somehow?

Yeah  sure, no reason to hold this up.

ulfgebhardt:
@daumie @roschaefer I will take care of this today

PR #594, opened 2019-05-10
Revert logout.vue

roschaefer:
For any static image URL we don't need to use `<hc-image>`. The assets
will be served from the frontend container. I did not anticipate that we
have a lot of static images in the frontend when I said "You can replace
every image with the `<hc-image>` component".

@aonomike: In the future you can make some screenshots and post them in the PR
and ask for help.

@ulfgebhardt: Yes, it would have been safer not to merge the PR. We
might run into the danger of forgetting things like this.

### Issues

- [X] None

### Checklist

- [X] None

### How2Test

- [X] None

### Todo

- [X] None

### DIFF
--- a/webapp/pages/logout.vue
+++ b/webapp/pages/logout.vue
@@ -11,11 +11,11 @@
           margin-bottom="xxx-small"
           centered
         >
-          <hc-image
+          <img
             style="width: 200px;"
-            :image-props="imageProps"
+            src="/img/sign-up/onourjourney.png"
             alt="Human Connection"
-          />
+          >
         </ds-space>
         <ds-space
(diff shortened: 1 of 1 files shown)


mattwr18:
Is there no way to take a screenshot of this @roschaefer?
It seems to me that this `webapp/pages/logout.vue` gets called when someone visits `/logout`, or clicks on the `Logout` link in the `Dropdown`, but the image appears so quickly before being redirected to `/login` that it's difficult to take a screen shot....

Actually, I just commented out `this.$router.replace('/')` from line 42 of logout.vue and caught this beautiful shot
Screenshot at 2019-05-10 14:05:49

PR #478, opened 2019-04-17
2019/kw15/user can change its username to emptystring

ulfgebhardt:
## Pullrequest

image

In the case the user manages to change its Username to `empty` the string `Anonymous` is displayed
image
image

### Issues

- fixes https://github.com/Human-Connection/Human-Connection/issues/376

### Checklist

- [X] None

### How2Test

- Try to change the username to smth with less then 3 characters

### Todo

- [x] I have the `userName()` function three times in the code in different modules - where is the correct place for this? @roschaefer @appinteractive [>a<](https://github.com/Human-Connection/Human-Connection/pull/478/files#diff-29d12fa60fafe1db78d5fe10f8ab1ecdR138) [>b<](https://github.com/Human-Connection/Human-Connection/pull/478/files#diff-270c3eb3fd96a99d6f50904782643074R171) [>c<](https://github.com/Human-Connection/Human-Connection/pull/478/files#diff-779a284a195624fd1835008c38982ac6R346)
- [x] is there a central place for serverside settings? @roschaefer @mattwr18 see [here](https://github.com/Human-Connection/Human-Connection/pull/478/files#diff-a58752d5a7674a873dcbb599d985dc07R16)

PR #596, opened 2019-05-10
Refactoring validation middleware

roschaefer:
## Pullrequest
Two things:

* Fix `UpdateUser` test block (was not testing anything).
* Put all validations in one place

@mattwr18 @ulfgebhardt please read my commit messages in this PR.
### Issues
This is a follow-up on #478 

### Checklist

- [X] None

### How2Test

- [X] None

### Todo

- [X] None


backend/src/resolvers/users.spec.js
```diff
@@ -5,15 +5,15 @@ import Factory from '../seed/factories'
 const factory = Factory()
 let client
 
-afterAll(async () => {
+afterEach(async () => {
```
roschaefer: @ulfgebhardt `afterEach` means: Run this after each test case. `afterAll` means: Run this after the entire test **suite**. `afterAll` is great for performance tweaks, but you have to tear down the created data yourself. Read: https://jestjs.io/docs/en/setup-teardown

backend/src/middleware/validation/index.js
```diff
@@ -0,0 +1,31 @@
...
+    /* eslint-disable-next-line no-return-await */
+    return await resolve(root, args, context, info)
+  } else {
+    throw new UserInputError('Input is not a URL')
+  }
+}
+
+export default {
+  Mutation: {
+    CreateUser: validateUsername,
```
roschaefer: @ulfgebhardt we should validate `username` also on creation, no? Without it, the validation would be kind of pointless.

... Except, the `CreateUser` will be called on registration and we supply no username at all. But we'll see. For now, I think it's safer to have it checked on Create+Update.

backend/src/resolvers/users.spec.js
```diff
@@ -35,34 +36,39 @@ describe('userMiddleware', () => {
     })
   })
 
-  describe('update User', () => {
+  describe('UpdateUser', () => {
+    beforeEach(async () => {
+      await factory.create('User', { id: 'u47', name: 'John Doe' })
```
roschaefer: @ulfgebhardt you tried to update a user that did not exist.

roschaefer:
Me: Adding/fixing tests
Codecov: Test coverage dropped by 0.01%.

:laughing:

backend/src/middleware/validation/index.js
```diff
@@ -0,0 +1,31 @@
+import { UserInputError } from 'apollo-server'
+
+const USERNAME_MIN_LENGTH = 3
+
+const validateUsername = async (resolve, root, args, context, info) => {
+  if (args.name && args.name.length >= USERNAME_MIN_LENGTH) {
+    /* eslint-disable-next-line no-return-await */
+    return await resolve(root, args, context, info)
```
mattwr18: are we updating the user twice here? once in the validation middleware and once on line 15 of the `userMiddleware`?

backend/src/middleware/validation/index.js
```diff
@@ -0,0 +1,31 @@
...
+    throw new UserInputError(`Username must be at least ${USERNAME_MIN_LENGTH} characters long!`)
+  }
+}
+
+const validateUrl = async (resolve, root, args, context, info) => {
+  const { url } = args
+  const isValid = url.match(/^(?:https?:\/\/)(?:[^@\n])?(?:www\.)?([^:/\n?]+)/g)
+  if (isValid) {
+    /* eslint-disable-next-line no-return-await */
+    return await resolve(root, args, context, info)
```
mattwr18: same as :point_up:

backend/src/middleware/validation/index.js
roschaefer: No, calling `resolve` will continue the chain of middlewares until you hit the final resolver.

backend/src/resolvers/users.spec.js
```diff
@@ -71,7 +77,8 @@ describe('userMiddleware', () => {
 
     it('with too short name', async () => {
       const variables = {
-        id: 'u1'
+        id: 'u47',
```
roschaefer: I think I have to give myself a "Request Changes" review here. I told @ulfgebhardt not to create test cases (== specification) for behaviour that we actually don't want to support, e.g. passing `id` from the client. I'll update that as soon as possible.

ulfgebhardt:
> Me: Adding/fixing tests
> Codecov: Test coverage dropped by 0.01%.
> 
> 😆

The Backend Test Setup needs to be fixed to comply with standard test runs of jest in oder to collect coverage properly.

roschaefer:
@ulfgebhardt did you see this? Check commit messages and "Files Changed" tab. If yes, you can merge this PR.

PR #690, opened 2019-05-27
Fix edit field render bug(CommentForm)

### LINKED ISSUE #582: 🐛 No Edit Field displayed in CommentForm
## :bug: Bugreport

There is sometimes no edit field displayed in the CommentForm …

First - Comment Form has no Edit Field

If you refresh the page the edit field is displayed …

Refresh Page - Comment Form has Edit Field

1. If you go to the profile of a user …
2. … and click on a post …
3. … and on this post scroll down to the bottom …
4. ... the CommentForm has no edit field

### Expected behavior

There should be an edit field displayed in the CommentForm.

### Version & Environment
 Type: Desktop      
 - OS: macOS      
 - Browser: Safari, FireFox

mattwr18:
- remove no-ssr, which was not necessary and causing the edit field not to appear the majority of the times visiting a Post.
- this was really bad user experience since a user would need to refresh the page to comment.
- removed args in refetchPostComments as there are no params passed in when it is called anymore
- needed to add an if statement since if there are no comments on a Post, then this.$apollo.queries.Post is undefined and it errors out trying to call refetch()
- update test to remove no-ssr

Co-authored-by: Mike Aono <aonomike@gmail.com>

### Issues

- fixes #582

### DIFF
--- a/webapp/components/comments/CommentForm/index.vue
+++ b/webapp/components/comments/CommentForm/index.vue
@@ -5,14 +5,12 @@
   >
     <template slot-scope="{ errors }">
       <ds-card>
-        <no-ssr>
-          <hc-editor
-            ref="editor"
-            :users="users"
-            :value="form.content"
-            @input="updateEditorContent"
-          />
-        </no-ssr>
+        <hc-editor
+          ref="editor"
+          :users="users"
+          :value="form.content"
+          @input="updateEditorContent"
+        />
         <ds-space />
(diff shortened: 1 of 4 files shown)


mattwr18:
Screenshot at 2019-05-27 17:07:18


This is what it now looks like when you click on any Post at anytime!!

mattwr18:
:rofl:

PR #695, opened 2019-05-28
Fix placeholder text not translating bug

### LINKED ISSUE #583: 🐛 Placeholder in CommentForm does not change Localisation dynamically
## :bug: Bugreport

If you change the language the `placeholder` in CommentForm does not change localisation.

1. If you go to the profile of a user …
2. … and click on a post …
3. … and on this post scroll down to the bottom to the CommentForm
4. change language
5. you can see that the `placeholder` text in the CommentForm edit field does not change localisation

### Expected behavior

There should be an edit field displayed in the CommentForm.
The CommentForm edit field `placeholder` text should change the localisation.

### Version & Environment
 Type: Desktop      
 - OS: macOS      
 - Browser: FireFox 

### Additional Context
This problem is related to the use of `tiptap` editor.

mattwr18:
- tiptap-extensions are not reactive
- we must call a method to change them dynamically
- emit changeLanguage event, which is listened for by the Editor component, which then calls a method to set the emptyNodeText of tiptap

### Issues

- fixes #583

### DIFF
--- a/webapp/components/Editor/index.vue
+++ b/webapp/components/Editor/index.vue
@@ -327,10 +327,18 @@ export default {
       },
     },
   },
+  mounted() {
+    this.$root.$on('changeLanguage', () => {
+      this.changePlaceHolderText()
+    })
+  },
   beforeDestroy() {
     this.editor.destroy()
   },
   methods: {
+    changePlaceHolderText() {
+      this.editor.extensions.options.placeholder.emptyNodeText = this.$t('editor.placeholder')
+    },
     // navigate to the previous item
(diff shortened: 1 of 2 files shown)


mattwr18:
@roschaefer, maybe?? we could not have placeholder text for example??
or there might be another way
One of the maintainers just responded to my approach on the issue I shared, saying |

https://github.com/scrumpy/tiptap/issues/17
```
@mattwr18 That should be fine. Would have done that in a similar way.

I should mention that this won't work in IE 11 without polyfill for Proxys
```

Issue #692, opened 2019-05-28
🚀 [Feature] Migrate all available Data from Alpha to Nitro


ulfgebhardt:
### Migrate all available Data from Alpha to Nitro

- [x] Modify import scripts to work with windows
- [x] Make a List of all available Data
- [x] Check off which Data is already transferred
- [ ] Transfere all Data remaining / List Todo's

roschaefer:
* Invites cannot be imported at all
* We import candos later, because we don't know how to model the data before we implement the feature

PR #697, opened 2019-05-28
🍰 2019/kw22/alpha_data_import

ulfgebhardt:
## 🍰 Pullrequest

fixed data import from alpha data:
- include all collections (commented out)
- refactored neo4j import script
- use of .env file for (additional) configurations / configuration overrides
- lots of fiddeling with neo4j cql files and cypher shell

### Issues

- relates https://github.com/Human-Connection/Human-Connection/issues/692

### Todo

- [ ] catalogue all the data to be imported (will not be done in this PR)
- [ ] import the data (will not be done in this PR)


deployment/legacy-migration/maintenance-worker/migration/neo4j/badges.cql
```diff
@@ -1,4 +1,4 @@
-CALL apoc.load.json('file:/tmp/mongo-export/splits/current-chunk.json') YIELD value as badge
+CALL apoc.load.json("file:${IMPORT_CHUNK_PATH_CQL}") YIELD value as badge
```
roschaefer: I don't think you can expand variables like this. It's a `.cql` file, not a bash script.

deployment/legacy-migration/maintenance-worker/migration/neo4j/import.sh
```diff
@@ -1,17 +1,48 @@
...
-do
-  for chunk in /tmp/mongo-export/splits/$collection/*
+# Import collection function defintion
+function import_collection () {
+  for chunk in ${IMPORT_PATH}splits/$1/*
   do
-    mv $chunk /tmp/mongo-export/splits/current-chunk.json
-    echo "Import ${chunk}" && cypher-shell < $SCRIPT_DIRECTORY/$collection.cql
+    mv $chunk ${IMPORT_CHUNK_PATH}
+    NEO4J_COMMAND="$(envsubst '${IMPORT_CHUNK_PATH_CQL}' < $(dirname "$0")/$1.cql)"
```
roschaefer: Or is `envsubst` the command to replace `$VARIABLE` in arbitrary files?

deployment/legacy-migration/maintenance-worker/migration/neo4j/badges.cql
ulfgebhardt: You can if you use magic like I do here: https://github.com/Human-Connection/Human-Connection/pull/697/files#diff-c6d49070ce711f563a75e723b1e8682cR14

deployment/legacy-migration/maintenance-worker/migration/neo4j/import.sh
ulfgebhardt: yes it is

ulfgebhardt:
I would like to merge this with commented out imports to start the next step to analyse which parts are already correctly imported and which are missing - if you agree on that

deployment/legacy-migration/maintenance-worker/migration/neo4j/import.sh
```diff
@@ -1,17 +1,48 @@
...
+#import_collection "organizations"
+#import_collection "pages"
+#import_collection "projects"
+#import_collection "settings"
+#import_collection "shouts"
+#import_collection "status"
+#import_collection "systemnotifications"
+#import_collection "users"
+#import_collection "userscandos"
+#import_collection "usersettings"
```
roschaefer: I cannot approve these changes. I would like to preserve the status quo, which is a lot more than just badges.

PR #699, opened 2019-05-28
269 filter by followed users

### LINKED ISSUE #269: Filter by followed users
```gherkin
As a user of the network
I want to have a filter (possibly as a default filter) for followed users and friends only
So I just read and see posts of users that I know and that I'm interested in
```

Screenshot - 2019-05-31T213522 073

Screenshot - 2019-05-31T213754 862

roschaefer:
## 🍰 Pullrequest
Screenshot - 2019-06-05T223845 651

### Issues

- close #269 

### Todo

- [X] None

### DIFF
--- a/cypress/integration/common/steps.js
+++ b/cypress/integration/common/steps.js
@@ -230,7 +230,7 @@ When('I type in the following text:', text => {
 
 Then('the post shows up on the landing page at position {int}', index => {
   cy.openPage('landing')
-  const selector = `:nth-child(${index}) > .ds-card > .ds-card-content`
+  const selector = `.post-card:nth-child(${index}) > .ds-card-content`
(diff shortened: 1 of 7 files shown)


ulfgebhardt:
@roschaefer this is out of scope for the current release if i am not mistaken ?!

roschaefer:
@ulfgebhardt as discussed in the last sprint planning, this feature is "implemented" in the alpha. It's just implemented in another way: You get a notification if
1. one of your followed users writes a post
2. sb. commented on your post
3. sb. mentions you

We have currently case 3 implemented. We will not implement case 1. Instead, there will be a filter where you can see only posts of your followed users. This *is* part of the release, because it's a supported feature of the alpha and following users would be very pointless if you cannot make use of it.

PR #701, opened 2019-05-29
🍰 Fix update incompatibility

roschaefer:
### Fix update incompatibility
It seems that prettier supports vue in version `v1.15` and above.
See: https://github.com/prettier/prettier/releases/tag/1.15.0

So I guess that interferes with the old vue linter rules in place. I
graduated the vue rule from `recommended` to `essential` and now
running `yarn run lint --fix` is doing fine.

### Issues

- None

### Todo

- [X] Could you please check out this branch and run `yarn run lint --fix`?
- [X] Could we agree if we want to adopt the new style rules?
- [ ] Create Followup Issue on vscode formating problems


roschaefer:
@Tirokk @ogerly @ulfgebhardt @appinteractive @Lulalaby @Mastercuber someone has to run `yarn run lint --fix`. Whoever that might be is going to "own" a lot of code. So, if sb. wants to jump up in the contributors list at once - go ahead.

ulfgebhardt:
@mattwr18 says vscode has formating issues - I would agree to merge if we make an followup issue @roschaefer @mattwr18

ulfgebhardt:
And please @roschaefer assign all the values to your PR, make issues for them and use the PR template properly - you will save me some time <3

roschaefer:
@ulfgebhardt OK! I will do my best regarding zenhub.

@ulfgebhardt I'll merge this, OK? I'm not using vscode so I have a hard time to check if there is any incompatibility...

PR #706, opened 2019-05-30
Add logo as default avatar background image

### LINKED ISSUE #686: 🐛 [Bug] Dummy placeholder for empty user avatars
### Dummy placeholder for empty user avatars

There should be a default placeholder for the user avatar when none is set (default).

mattwr18:
### Issues

- fixes #686 
Screenshot at 2019-05-30 16:09:33

### DIFF
--- a/webapp/components/Upload/index.vue
+++ b/webapp/components/Upload/index.vue
@@ -35,7 +35,9 @@ export default {
   },
   computed: {
     backgroundImage() {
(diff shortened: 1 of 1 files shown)


mattwr18:
I was thinking it would be nice to have like a series of avatars that could be assigned to new user's in sequence or randomly  until they choose their own, but this was the easiest quickest solution...
happy to work on a different solution in the future though :)

PR #715, opened 2019-05-31
206 setup neo4j db with dump

### LINKED ISSUE #206: Neo4J DB initialization
For performance, we would like to have a reproducible set of indices in our Neo4J database.

Write a script, e.g. `migrate` or `db:setup` to create indices in Neo4J. This script will be run when a container starts and it should run after the default `entrypoint.sh`.

**Additional context**:
[Read about how to create an index in Neo4J](https://www.quackit.com/neo4j/tutorial/neo4j_create_an_index_using_cypher.cfm)

We could also add an initial user with admin privileges. This could replace the need for an unauthenticated server for seeding during our test suite.

roschaefer:
## 🍰 Pullrequest
Yet another reason to use docker: You don't have to worry about the unique constraints anymore.

### Issues

- fixes #206 

### Todo

- [X] None


roschaefer:
This is odd. The constraints won't get imported. I tried to dump my host machine database instance. The neo4j constraints get imported when importing back to my host machine, but they do not get imported when importing it in the container.

Our build server fails because of missing constraints.

roschaefer:
closed in favour of #739

PR #724, opened 2019-06-03
🍰 2019/kw22/refactor_backend_structure

ulfgebhardt:
### 2019/kw22/refactor_backend_structure
- refactored backend structure
- use and reference global config

### Issues

- relates https://github.com/Human-Connection/Human-Connection/issues/692

### Todo

- [ ] Middleware refactoring - not done in this PR


backend/package.json
```diff
@@ -11,7 +11,7 @@
     "lint": "eslint src --config .eslintrc.js",
     "test": "run-s test:jest test:cucumber",
     "test:before:server": "cross-env GRAPHQL_URI=http://localhost:4123 GRAPHQL_PORT=4123 yarn run dev 2> /dev/null",
-    "test:before:seeder": "cross-env GRAPHQL_URI=http://localhost:4001 GRAPHQL_PORT=4001 DISABLED_MIDDLEWARES=permissions,activityPub yarn run dev 2> /dev/null",
+    "test:before:seeder": "cross-env GRAPHQL_URI=http://localhost:4001 GRAPHQL_PORT=4001 DEBUG=true DISABLED_MIDDLEWARES=permissions,activityPub yarn run dev 2> /dev/null",
```
roschaefer: Remove `DEBUG`. Maybe I'm missing sth, but I don't see the reason for it.

ulfgebhardt: i enforced this in oder to remove the whole NODE_ENV check. Not the check is for DEBUG and DEBUG can only be set in a non production environment
`DEBUG: process.env.NODE_ENV !== 'production' && process.env.DEBUG === 'true'`

PR #730, opened 2019-06-03
269 filter by followed users backend part

roschaefer:
## 🍰 Pullrequest
This is the backend part for #269. It can be merged independently.

With this PR the backend accepts an optional input parameter for `Post` query: It's called `filterBubble` and has currently two options: ~`{ author: followed }`~ `{ author: following }` or `{ author: all }`.

Screenshot - 2019-06-03T195824 383

versus

Screenshot - 2019-06-03T195903 512

I'm doing my best in the test setup to minimize those tests that do HTTP calls.

### Issues

- None

### Todo

- [X] None

### DIFF
--- a/backend/src/middleware/filterBubble/replaceParams.spec.js
+++ b/backend/src/middleware/filterBubble/replaceParams.spec.js
@@ -75,7 +75,7 @@ describe('replaceParams', () => {
 
         it('makes database calls', async () => {
           await action()
-          expect(run).toHaveBeenCalled()
+          expect(run).toHaveBeenCalledTimes(1)
         })
 
         describe('given any additional filter args', () => {
--- a/backend/src/middleware/filterBubble/replaceParams.spec.js
+++ b/backend/src/middleware/filterBubble/replaceParams.spec.js
@@ -38,9 +38,9 @@ describe('replaceParams', () => {
         context.user = null
       })
 
-      describe('{ filterBubble: { author: followed } }', () => {
+      describe('{ filterBubble: { author: following } }', () => {
         it('throws error', async () => {
-          args = { filterBubble: { author: 'followed' } }
+          args = { filterBubble: { author: 'following' } }
           await expect(action()).rejects.toThrow('You are unauthenticated')
         })
       })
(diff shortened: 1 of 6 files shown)


mattwr18:
currently failing the build with this output
Screenshot at 2019-06-03 22:17:49

backend/src/middleware/filterBubble/replaceParams.spec.js
```diff
@@ -0,0 +1,129 @@
...
+        })
+
+        it('returns args object with resolved ids of followed users', async () => {
+          const expected = { filter: { author: { id_in: [1, 2, 3] } } }
+          await expect(action()).resolves.toEqual(expected)
+        })
+
+        it('makes database calls', async () => {
+          await action()
+          expect(run).toHaveBeenCalled()
```
mattwr18: Is there any benefit here of using `toHaveBeenCalledTimes(1)` or `toHaveBeenCalledWith('whatever args')`
I typically try to be more specific because I want the test to fail if it is called more than once, or with the wrong args

One use cause for this that I saw with @aonomike, is that if you don't properly `clearAllMocks()` in between tests, you can get false positives.

roschaefer: @mattwr18 I usually don't try to be overly specific, so you don't need to refactor the tests if you refactor the code. What about `clearAllMocks`? I didn't know that. What is the exact problem? Does it happen even if you re-initialize the variables in a `beforeEach` block?

mattwr18: The problem is that unless you clear the mocks' calls one of your previous tests can call your mock function and then you get a false positive cause you think the it block that you are trying to test has called the function, but it's really a previous test.

I would favor using `toHaveBeenCalledTimes(1)` because there's not really a use case I can think of where you would want it to be called multiple times, but if you prefer to leave it as `toHaveBeenCalled()` or either way really, I think it's better to have an afterEach, which calls `jest.clearAllMocks()` to ensure you are testing what you think you are.

roschaefer: done!

roschaefer:
@mattwr18 I rebased this PR and removed all the logic to patch the augmented schema retrospectively. This can be merged if I get an approving review.

PR #732, opened 2019-06-03
Hover effect user image upload

mattwr18:
## 🍰 Pullrequest

This `PR` introduces more than just a hover effect because adding a hover effect to the functionality in `master` made the `vue-dropzone` non-clickable and so broke the functionality.

I've refactored it to remove a fair bit of code and get it working with custom slots, which was quite difficult since it also didn't work like the examples. The avatar was non-clickable and I finally got it working by changing the class to `dz-message`. 

Using custom slot allows us to use the default functionality of a avatar with a user's initials as a default
Screenshot at 2019-06-03 17:03:46

This next `gif` shows its functionality
2019-06-04-hover-effect-user-profile-image-upload

There is a bug where the user's avatar in the navbar is not reactive, but this is a bug on `master` as well.
Mayb
Screenshot at 2019-06-04 07:59:44
e we can open a separate ticket for it?


Tirokk:
Because of the pairing session I find not time gap !
Well done @mattwr18 !!!

giphy-4

PR #738, opened 2019-06-04
Refetch post after toggle disable patch

mattwr18:
## 🍰 Pullrequest

This `PR` is a patch to get rid of `console.errors` in our tests and to avoid use of `location.reload` to reload the entire page, instead it emits an event, which refetches the post, since we are using `asyncData` for server-side rendering

### DIFF
--- a/webapp/components/Modal/DisableModal.vue
+++ b/webapp/components/Modal/DisableModal.vue
@@ -69,9 +69,6 @@ export default {
         setTimeout(() => {
           this.$emit('close')
         }, 1000)
-        setTimeout(() => {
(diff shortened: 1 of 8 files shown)


ogerly:
nice work mat !!! ReleaseModal.vue is ok. but at Disable.Modal.vue I will rebuild it.

mattwr18:
Closing this as we are favoring another solution, which removes the reloads altogether and doesn't refetch the post.

This is inferior from the user experience, but as this is a feature just for moderators, we are choosing for less complexity over UX

PR #740, opened 2019-06-05
Delete my User Account functionality

### LINKED ISSUE #404: 🚀 [Feature] UserSettings - Delete Account & Data
### UserSettings - Delete Account & Data
image

- [ ] Confirm slider
- [ ] Delete Button + Functionality

Note: It could be sufficient to send an email if the delete functionality is provided to an admin

### Userstory
```
Feature: Delete own account and content
  As a user
  I'd like to delete all my data, ie. posts, comments, shouts, follows
  Because I have a "right to be forgotten"
```

### Design & Layout - Nitro

Not implemented:
image

mattwr18:
### Issues

- fixes #404


mattwr18:
progress so far
Screenshot at 2019-06-04 21:22:57

we don't have any checkmarks as part of our styleguide :woman_shrugging: so I have resorted to a `select`, but open to other ideas... maybe a modal to ask for confirmation(?)

mattwr18:
Here are some gifs with the functionality:
Delete user, but not posts or comments:
2019-06-07-delete-user-account-not-comments-or-posts
Delete user, with posts, but not comments:
2019-06-07-delete-user-account-and-posts
Delete user, with posts and comments:
2019-06-07-delete-user-account-posts-and-comments
I changed the styling slightly, which can be seen in the screenshots :point_down: 
Screenshot at 2019-06-07 18:38:08

Mobile Responsiveness:
Screenshot at 2019-06-07 18:40:07
Screenshot at 2019-06-07 18:40:37
Screenshot at 2019-06-07 18:38:35

backend/src/middleware/permissionsMiddleware.js
```diff
@@ -80,6 +80,12 @@ const isAuthor = rule({
   return authorId === user.id
 })
 
+const isDeletingOwnAccount = rule({
```
roschaefer: Oh yes, this is quite important!

backend/src/schema/resolvers/users.js
```diff
@@ -11,5 +11,27 @@ export default {
...
+    DeleteUser: async (object, params, context, resolveInfo) => {
+      const { resource } = params
+      const session = context.driver.session()
+
+      if (resource && resource.length) {
+        await Promise.all(
+          resource.map(async node => {
+            await session.run(
+              `
+            MATCH (resource:${node})<-[:WROTE]-(author:User {id: $userId})
```
roschaefer: A couple of thoughts here: First of all, I believe, you can do this in one statement. I would try sth. like:
```
MATCH (author:User {id: $userId}), (post:Post)<-[:WROTE]-(author), (comment:Comment)<[:WROTE]-(author)
DETACH DELETE author, post, comment
```
I don't know if it works like that but I guess you get the point.

backend/src/schema/resolvers/users.js
```diff
@@ -11,5 +11,27 @@ export default {
...
+            DETACH DELETE resource
+            RETURN author`,
+              {
+                userId: context.user.id,
+              },
+            )
+          }),
+        )
+        session.close()
```
roschaefer: Most importantly is that if you delete the nodes, let's say you delete a post, then you won't delete all the comments on that post, because they belong to a different user. So my suggestion is not to delete a post but delete **all attributes** of that post.

I would say it's good practice to have something like a tombstone to maintain the context and the associations with the data that is already there. This is even more important with a comment. If you have a conversation with deleted comments and you cannot see anymore where a comment has been deleted, the responses to that comment might look awkward and out of context.

webapp/components/DeleteAccount/DeleteAccount.spec.js
```diff
@@ -0,0 +1,168 @@
...
+    beforeEach(() => {
+      wrapper = Wrapper()
+    })
+
+    afterEach(() => {
+      jest.clearAllMocks()
+    })
+
+    it('defaults to deleteContributions to false', () => {
+      expect(wrapper.vm.deleteContributions).toEqual(false)
```
roschaefer: I think since the bug in #785 I'm reluctant to `wrapper.vm`. Couldn't we refactor the test here to actually do sth. with the component or see sth. in the html to check it?

webapp/package.json
```diff
@@ -56,7 +56,8 @@
     "@nuxtjs/style-resources": "~0.1.2",
     "accounting": "~0.4.1",
     "apollo-cache-inmemory": "~1.5.1",
-    "apollo-client": "~2.6.2",
+    "apollo-client": "~2.6.1",
+    "buefy": "^0.7.7",
```
roschaefer: Ohhhh, @appinteractive will not like this line of code! :laughing:

webapp/locales/en.json
```diff
@@ -74,7 +74,12 @@
       "name": "Download Data"
     },
     "delete": {
-      "name": "Delete Account"
+      "name": "Delete my User Account",
+      "countPosts": "Delete my {count} posts",
+      "countComments": "Delete my {count} comments",
+      "accountDescription": "Be aware that your Post and Comments are important to our community. If you still choose to delete those you have to mark them below.",
+      "accountWarning": "You <b>CAN'T MANAGE</b> and <b>CAN'T RECOVER</b> your Account, Posts... after deleting your account!",
+      "success": "Account successfully deleted"
```
roschaefer: Use https://www.deepl.com/translator for German translations and you will get an extra translation check from me for free :wink:

backend/src/schema/resolvers/users.js
mattwr18: the point is they wouldn't have the choice of deleting just their account, or just their account and posts, just their account and comments, etc...

webapp/package.json
mattwr18: it's what gives us our switch and checkboxes, so unless there is a plan to add them to the styleguide/ a better library, we're kinda stuck with them...

roschaefer: We had bulma in the past and @appinteractive started to hate it. I guess, the best option is to maintain the styleguide now. Ie. adding features to it :slightly_frowning_face:

backend/src/schema/resolvers/users.js
roschaefer: Oh, right, true! We cannot customize `neo4j-graphql-js`.

webapp/components/DeleteAccount/DeleteAccount.spec.js
mattwr18: I tried to click on the switch, and on the check boxes, but they aren't available... at least that I could find

mattwr18:
I have changed up the implementation, and followed some other of your ideas @roschaefer... here is the new functionality

Deleting only my account:
2019-06-10 -delete-user-account-not-comments-or-posts

Deleting my account, and posts:
2019-06-10-delete-user-account-and-posts

Deleting my account, and comments:
2019-06-10-delete-user-account-and-comments

Deleting my account, posts, and comments:
2019-06-10 -delete-user-account-posts-and-comments

New look,
Screenshot at 2019-06-10 14:07:12

In German,
Screenshot at 2019-06-10 14:07:38


Mobile Responsiveness:
Screenshot at 2019-06-10 14:08:34
Screenshot at 2019-06-10 14:08:47
Screenshot at 2019-06-10 14:09:00
Screenshot at 2019-06-10 14:09:11

backend/src/schema/resolvers/users.js
mattwr18: hey @roschaefer, I think I followed all of your advice, or at least tried to...
I agree that it would be good to have some "tombstone", if a user deletes their comment, either by deleting their account, with all its comments, or by deleting a single comment.
It feels to me that this functionality is a separate ticket though, don't you agree? 
The reason I feel like that is cause the simple action of deleting one comment doesn't support it, much less deleting all comments.

mattwr18: Maybe when a user deletes a comment they have made, instead of deleting the node, we could delete all attributes, except the `id`, and change the content to `this comment has been deleted by the user`, plus add a `deletedOn` date?

ulfgebhardt:
@mattwr18 Can we name it "Delete Account" for a shorter name in the menu? <3

backend/src/schema/resolvers/users.js
roschaefer: Yes, that is the implementation of the alpha! We currently save a boolean flag on nodes `deleted = true`. Setting a date would be better, but we have to change the implementation of all the counters etc. I would suggest to open another issue for refactoring the `deleted` and `disabled` flags from type Boolean to type DateTime.

webapp/locales/de.json
```diff
@@ -82,8 +73,14 @@
     "download": {
       "name": "Daten herunterladen"
     },
-    "delete": {
-      "name": "Konto löschen"
+    "deleteUserAccount": {
+      "name": "Mein Benutzerkonto löschen",
+      "contributionsCount": "Meine {count} Beiträge löschen",
+      "commentsCount": "Meine {count} Kommentare löschen",
+      "accountDescription": "Seien Sie sich bewusst, dass Ihre Beiträge und Kommentare für unsere Community wichtig sind. Wenn du sie trotzdem löschen möchtest, musst du dies unten auswählen.",
```
roschaefer:
```suggestion
      "accountDescription": "Sie dir bewusst, dass deine Beiträge und Kommentare für unsere Community wichtig sind. Wenn du sie trotzdem löschen möchtest, musst du dies unten auswählen.",
```

webapp/locales/de.json
```diff
@@ -82,8 +73,14 @@
...
+      "accountWarning": "Dein Benutzerkonto, die Beiträge, etc. kannst Du nach dem Löschen <b>WEDER VERWALTEN NOCH WIEDERHERSTELLEN!</b>",
+      "success": "Konto erfolgreich gelöscht",
+      "pleaseConfirm": "<b class='is-danger'>Zerstörerische Aktion!</b> Geben Sie <b>{confirm}</b> ein, um zu bestätigen."
```
roschaefer:
```suggestion
      "pleaseConfirm": "<b class='is-danger'>Zerstörerische Aktion!</b> Gib <b>{confirm}</b> ein, um zu bestätigen."
```

backend/src/schema/resolvers/users.js
mattwr18: So, you suggesting that instead of deleting the comment, we should be just changing the attribute `delete = true`?

And for `posts`, you would expect that we not delete them completely because there are other user's comments on them? just delete the user relationship and it's content??

appinteractive:
Hey guys, nice progress! Just the thing that you want the user to type that for every option is a bit of overkill from my perspective! Also does it not encourage anymore to leave the content in my opinion. I would advocate for using the checkboxes for the option again.

backend/src/schema/resolvers/users.js
roschaefer: Regarding the comments, yes. Regarding the posts: Up to you. You won't see deleted posts in your newsfeed. So basically removing the posts will certainly make the posts along with the comments unlocatable. There will be probably no difference from the end user perspective, however we administrators could make use of the data if the relationships are still there at least.

mattwr18: taken care of in the latest commit

mattwr18:
Screenshot at 2019-06-12 21:48:42
Screenshot at 2019-06-12 21:48:24
Screenshot at 2019-06-12 21:49:02
Screenshot at 2019-06-12 21:49:47
Screenshot at 2019-06-12 21:49:57

webapp/locales/de.json
```diff
@@ -25,16 +25,7 @@
     "shouted": "Empfohlen",
     "commented": "Kommentiert",
     "userAnonym": "Anonymus",
-    "socialMedia": "Wo sonst finde ich",
-    "network": {
```
roschaefer: @mattwr18 why did you delete the translations? :thinking:

webapp/pages/profile/_id/_slug.vue
```diff
@@ -69,13 +69,11 @@
         </ds-card>
         <ds-space />
         <ds-heading tag="h3" soft style="text-align: center; margin-bottom: 10px;">
-          {{ $t('profile.network.title') }}
+          Netzwerk
```
roschaefer: Please don't untranslate :heart:

webapp/locales/de.json
mattwr18: I didn't! well, I must have worked through a merge conflict incorrectly... I'll add it back
weird thing is that trying to pull in the latest from master says it's up to date :woman_shrugging:

webapp/pages/profile/_id/_slug.vue
mattwr18: Also, don't know what happened here
:woman_shrugging:

PR #753, opened 2019-06-05
Disable automatically generated queries - Mutations are still enabled

roschaefer:
Ok, it does not work as smoothly as I hoped: When youd disable the
generated queries and hard-code the definition into a file, then
graphql-validation will complain about undefined input parameter types.
Ie. all the *Filter and *Order types. So you have copy+paste those, too.

This PR is still a discussion. So everybody feel free to give your input on this. You can see what definitions need to be maintained when you disable the `query: false` in the `neo4j-graphql-js` configuration.

### Issues

- None

### Todo

- [X] None


roschaefer:
I think we're not going down that road in the near future

PR #754, opened 2019-06-05
402 userprofil tab click function

### LINKED ISSUE #402: 🚀 [Feature] UserProfile Shouted Tab - Display Posts User has shouted
### UserProfile Shouted Tab - Display Posts User has shouted
image

- [ ] Allow Shouted-Tab-Selection in User Profile
- [ ] Display Posts Profile-User has shouted

### Design & Layout - Alpha

image

### Design & Layout - Nitro

Not implemented:
image

ogerly:
## 🍰 Pullrequest
the tab menu in the profile, now has click events that work

ezgif com-video-to-gif (4)

### Issues
 
- fixes #402 
 

### Todo
- [ x ] css change for active TAB dynamically
- [  ]  load data :)


webapp/pages/profile/_id/_slug.vue
```diff
@@ -144,16 +144,17 @@
                     delay: { show: 500 },
                   }"
                   class="ds-tab-nav-item pointer ds-tab-nav-item-active"
-                  @click="tabActivity('posts', $event)"
                 >
-                  <ds-space margin="small">
-                    <!-- TODO: find better solution for rendering errors -->
-                    <no-ssr>
```
roschaefer: @ogerly could you check if `no-ssr` is really necessary here?

webapp/pages/profile/_id/_slug.vue
```diff
@@ -183,14 +185,16 @@
...
-                      <ds-number :label="$t('profile.shouted')">
-                        <hc-count-to slot="count" :end-val="user.shoutedCount" />
-                      </ds-number>
-                    </no-ssr>
-                  </ds-space>
+                  <a @click="handleTab('shouted', $event)">
```
roschaefer:
```suggestion
                  <a @click="handleShoutedTab">
```

webapp/pages/profile/_id/_slug.vue
```diff
@@ -317,6 +321,19 @@ export default {
     },
   },
   methods: {
+    handleTab(tab) {
```
roschaefer:
```suggestion
    handleShoutedTab() {},
    handleCommentedTab() {},
    handlePostTab() {}


```

webapp/pages/profile/_id/_slug.vue
ogerly: > no-ssr

@roschaefer  I was just reading. 
no-ssr is practically not necessary.

ogerly:
ezgif com-video-to-gif (5)

i've chosen another menu. it's all too complicated just for me. and it's annoying when you can simply work fluidly ... because the simplest things still cause problems. 
that seems to go according to my ideas now ... 
TODO: tooltip comes back. or is that really necessary? 
@roschaefer @Tirokk @mattwr18 

but the active problem is solved with it. somehow it just seems to me too much code ... for such an unimportant thing ... the thing of the exam hasn't quite worked out in me yet either

ogerly:
ok now it fits :) 

ezgif com-video-to-gif (6)

ogerly:
i have a pull from master gemacht 

> @ogerly what happened to this PR? Again, the PR has suddenly a lot of unrelated commits. Please regularly `git checkout master` and `git pull` to get the latest changes. Then `git checkout -b your_feature_branch` from there. If you want to update your feature branch with the latest changes, simply do a `git fetch` and `git merge origin/master`.
> 
> what are you doing

that i learned today why the pull is so important .. otherwise you make yourself noble because of mistakes which are actually no longer there ... 
is now conscious to me.
but that works now ? and is clean? 
it is done again with the latest master version ...
zack it became green.

webapp/pages/profile/_id/_slug.vue
```diff
@@ -317,6 +328,11 @@ export default {
...
+      this.tabActive = tab
+      this.Post = null
```
roschaefer: This is probably not very intuitive: Setting the `Post` to null ensures that you don't see two loading spinners at the same time. (We have two for different purposes, because we might switch tabs or we fetch more contributions).

backend/src/schema/types/type/User.gql
```diff
@@ -56,14 +56,14 @@ type User {
   contributionsCount: Int! @cypher(
     statement: """
       MATCH (this)-[:WROTE]->(r:Post)
-      WHERE (NOT exists(r.deleted) OR r.deleted = false)
-      AND (NOT exists(r.disabled) OR r.disabled = false)
+      WHERE NOT r.deleted = true AND NOT r.disabled = true
```
roschaefer: @ulfgebhardt there was a strange bug that the pagination button ("Fetch more") would not go away even though there are no records left. This is the reason why:

Screenshot - 2019-06-07T014505 476

Pagination on this page takes into account the total number of `contributionsCount` which you've overwritten in 6f0447515. It was incorrectly more than expected, the condition `records.length == total` never met and pagination wouldn't disappear.

webapp/pages/profile/_id/_slug.vue
ogerly: ok, the exact dependencies are not quite familiar to me yet. because the first time you call up a profile / your own profile, the first TAB is displayed. and i had the problem with the dynamic classes. 
Case: if I set an active CLASS statically to the postTAB then it is strangely double there. 
which can also be a bug in the devconsole in chrome. however it deletes the CLASS active cleanly. but with postTAB it is set twice! 
which should not be allowed. but dynamic CLASS behave differently than addClass. :class is apparently very stupid. so in the sense that it works like a hammer. but I got that out soon

roschaefer:
@mattwr18 @appinteractive @ulfgebhardt @Tirokk could you add your review here? I added a couple of commits and this is ready to merge from my side.

If there is anything that you would like to see tested, feel free to tell @ogerly and me.

PR #755, opened 2019-06-05
[WIP] Try setting up vuex for disabled posts

mattwr18:
### Todo

- [X] None


mattwr18:
Closing this as we are favoring another solution, which removes the reloads altogether and doesn't refetch the post.

This is inferior from the user experience, but as this is a feature just for moderators, we are choosing for less complexity over UX

PR #762, opened 2019-06-06
Refactor filter bubble

roschaefer:
## 🍰 Pullrequest
Refactored the backend and threw away a lot of custom code. What I'm trying to do is possible with `neo4j-graphql-js` filter input params. I could see the performance slightly deteriorate but I'm in favour of fixing performance problems if really requested and go for the smaller complexity if possible.

### Issues

- None

### Todo

- [X] None


webapp/components/FilterMenu/FilterMenu.spec.js
```diff
@@ -9,45 +9,60 @@ localVue.use(Styleguide)
...
 
-      it('makes button primary', () => {
-        wrapper.find({ name: 'filter-by-followed-authors-only' }).trigger('click')
-        expect(
-          wrapper.find({ name: 'filter-by-followed-authors-only' }).classes('ds-button-primary'),
-        ).toBe(true)
+      describe('click "filter-by-followed-authors-only" button', () => {
+        it('emits filterBubble object', () => {
+          wrapper.find({ name: 'filter-by-followed-authors-only' }).trigger('click')
+          expect(wrapper.emitted('changeFilterBubble')).toBeTruthy()
```
mattwr18: At the risk of you getting annoyed, and not saying I wouldn't be happy to merge as is, I am not a fan of these tests that check if something is `toBeTruthy()`. I feel like these are flaky. 

I am aware that in the documentation they use them https://vue-test-utils.vuejs.org/api/wrapper/#emitted
so maybe it's something I need to get over, but I personally would write it `expect(wrapper.emitted().changeFilterBubble.length).toBe(1)`, for what it's worth

though I see below that you are testing what the emit event is called with, which I really like :smile_cat:

roschaefer: ah, yeah, that test is just to check that it emits anything after all..

mattwr18: yes, I got that, and I guess it would never be Truthy unless it actually emits something, which is probably why they have it in the examples in the docs... just makes me a lil' nervous hehe

PR #786, opened 2019-06-07
[Bug] Untranslated Texts in Post Page #764

### LINKED ISSUE #764: 🐛 [Bug] Untranslated Texts in Post Page
### [Bug] Untranslated Texts in Post Page

image

There is several Texts not translated correctly (we are not talking about switching language)

### Expected behavior

All Texts should be translated/translatable

ogerly:
## 🍰 Pullrequest
Los paquetes de idioma español y francés están actualizados. 

### Issues
 
- fixes #764 
 
  

### Todo
Which language do we use as a reference? 
I would suggest German.

### DIFF
--- a/webapp/locales/de.json
+++ b/webapp/locales/de.json
@@ -19,9 +19,9 @@
   "profile": {
     "name": "Mein Profil",
     "memberSince": "Mitglied seit",
-    "follow": "Folgen",
-    "followers": "Folgen",
-    "following": "Folgt",
+    "follow": "abonnieren",
+    "followers": "Abonnenten",
+    "following": "abonniert",
     "shouted": "Empfohlen",
     "commented": "Kommentiert",
     "userAnonym": "Anonymus",
@@ -91,9 +91,9 @@
     }
   },
   "admin": {
-    "name": "Systemverwaltung",
+    "name": "Admin",
     "dashboard": {
-      "name": "Startzentrale",
+      "name": "Dashboard",
       "users": "Benutzer",
       "posts": "Beiträge",
       "comments": "Kommentare",
@@ -102,7 +102,7 @@
       "projects": "Projekte",
       "invites": "Einladungen",
       "follows": "Folgen",
-      "shouts": "Shouts"
+      "shouts": "Empfehlungen"
     },
     "organizations": {
       "name": "Organisationen"
@@ -261,7 +261,7 @@
   },
   "followButton": {
     "follow": "Folgen",
-    "following": "Folge Ich"
+    "following": "Folge ich"
   },
   "shoutButton": {
     "shouted": "empfohlen"
--- a/webapp/locales/en.json
+++ b/webapp/locales/en.json
@@ -14,7 +14,7 @@
     "hello": "Hello"
   },
   "editor": {
-    "placeholder": "Write something inspiring..."
+    "placeholder": "Leave your inspirational thoughts..."
   },
   "profile": {
     "name": "My Profile",
--- a/webapp/locales/en.json
+++ b/webapp/locales/en.json
@@ -14,7 +14,7 @@
     "hello": "Hello"
   },
   "editor": {
-    "placeholder": "Leave your inspirational thoughts..."
+    "placeholder": "Write something inspiring..."
   },
   "profile": {
     "name": "My Profile",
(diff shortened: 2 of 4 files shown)


webapp/locales/en.json
```diff
@@ -14,22 +14,22 @@
     "hello": "Hello"
   },
   "editor": {
-    "placeholder": "Leave your inspirational thoughts..."
+    "placeholder": "Write something inspiring..."
```
mattwr18: Could you justify this change? I prefer the way it is

webapp/locales/en.json
```diff
@@ -14,22 +14,22 @@
...
-    "shouted": "Shouted",
+    "shouted": "Recommended",
     "commented": "Commented",
     "userAnonym": "Anonymous",
     "socialMedia": "Where else can I find"
   },
   "notifications": {
     "menu": {
-      "mentioned": "has mentioned you in a post"
+      "mentioned": "mentioned you in a post"
```
mattwr18: It is better English to say `has mentioned` because it is not specific when the mention happened

webapp/locales/en.json
```diff
@@ -102,7 +102,7 @@
       "projects": "Projects",
       "invites": "Invites",
       "follows": "Follows",
-      "shouts": "Shouts"
+      "shouts": "Recommended"
```
mattwr18: I don't think we need to change this in English

webapp/locales/de.json
```diff
@@ -102,7 +102,7 @@
       "projects": "Projekte",
       "invites": "Einladungen",
       "follows": "Folgen",
-      "shouts": "Shouts"
+      "shouts": "Empfehlungen"
```
roschaefer: Actually, I think I :heart it. Indeed, shouting might have a negative connotation in English.

Shouting in Human Connection means to increase the reach of the post. Let more people see it. So I kind of like your translation of it! :+1:

webapp/locales/en.json
roschaefer: Agree with @mattwr18 here

webapp/locales/en.json
roschaefer: @mattwr18 do you think "Shouting" could be misunderstood in a negative way? As I pointed out in the comment above :point_up:?

webapp/locales/es.json
```diff
@@ -119,6 +173,113 @@
...
+    "success": "Discapacitado con éxito",
+    "user": {
+      "title": "Desactivar usuario",
+      "type": "Usuario",
+      "message": "¿Realmente quieres deshabilitar el usuario \"<b>{name}</b>\"?"
+    },
+    "contribution": {
+      "title": "Deshabilitar contribución",
+      "type": "Contribución",
+      "message": "¿Realmente quieres deshabilitar la contribución \"<b>{name}</b>\"?"
```
roschaefer: Whaaaat, this is possible? I didn't know you can add variables to a translation.

webapp/locales/de.json
```diff
@@ -91,9 +91,9 @@
     }
   },
   "admin": {
-    "name": "Systemverwaltung",
+    "name": "Admin",
     "dashboard": {
-      "name": "Startzentrale",
+      "name": "Dashboard",
```
roschaefer: Yes I think most Germans understand the meaning of "Dashboard" by now. Literally translated it means "Armaturenbrett".

webapp/locales/es.json
ogerly: ahhh ... there is a mistake ... the B element closes but it does not open ... 
an error which e.g. is not recognized!

webapp/locales/de.json
ogerly: certain words just don't work in other languages. we are forced to find our own terms. 
in the admin area it has more to do with people who can use the word dashboard. and expect something. 
linguistics is definitely an important branch and our cultural burden in europe :) 
i think it's important that we develop our own style. one world .. you can think if people are one step ahead.

roschaefer:
@ogerly could you add @mattwr18 suggestions here? :slightly_smiling_face:

webapp/locales/en.json
mattwr18: sorry, @roschaefer, I saw this, but didn't reply...
I don't think so, but maybe shout-out would be better... like to give someone or something a shout-out... just that I think Instagram already uses that

webapp/locales/de.json
mattwr18: yes, we have a lot of expressions in English like `to shout something from the rooftop`, `to give a shout-out`, etc..

Screenshot at 2019-06-13 21:02:41
Screenshot at 2019-06-13 21:02:29

webapp/locales/es.json
mattwr18: what do you mean @ogerly... I needed to add this `<b></b>` to a `v-html` for it to get converted to bold, instead of it being read literally, but maybe that is not what you mean at all.

@roschaefer, I was also pleasantly surprised to see this!! and it's super handy when you need to pass in computed values, like the `commentsCount`, `contributionsCount`

PR #795, opened 2019-06-10
Create Posts with a specified language, or fallback

### LINKED ISSUE #781: 🚀 [Feature] Language of Contribution
## :rocket: Feature

The User should be able to select the Language of the Contribution.

### Design & Layout

Alpha - Create Post - Language - Bildschirmfoto 2019-06-06 um 09 13 14

mattwr18:
### Issues

- fixes #781 

### Todo

- [ ] Write component tests for languages


mattwr18:
Screenshot at 2019-06-12 19:40:21
Screenshot at 2019-06-12 19:40:21
Screenshot at 2019-06-12 19:42:19

mattwr18:
Screenshot at 2019-06-13 17:27:40

this is what you would expect @Tirokk??

webapp/components/ContributionForm/index.vue
```diff
@@ -137,7 +138,10 @@ export default {
       this.$refs.contributionForm.update('content', value)
     },
     returnLocaleName(locale) {
-      if (this.$i18n.locale() === locale.code) {
+      if (
+        (this.contribution && this.contribution.language === locale.code) ||
+        (!this.contribution && this.$i18n.locale() === locale.code)
+      ) {
```
Tirokk: What happens in the else case?

mattwr18: good question @Tirokk, I didn't think there was a case where `this.$i8n.locale()` is undefined :thinking: 
what would you expect to happen if there is no locale?

Tirokk: Just use english as the world language @mattwr18 , I think. Or?
If the else case is not possible, we don’t need an ìf` here?!

mattwr18: Addressed in our latest pairing session @Tirokk... unfortunately, I forgot to add you as a co-author, but I think you know I appreciate you!!

PR #821, opened 2019-06-13
Implement prefix of image urls with a filter

### LINKED ISSUE #820: 🐛 [Bug] Posts teaser images are broken
## :bug: Bugreport
Posts teaser images are broken after #809 and #810 

1. Visit https://nitro-production.human-connection.org/
2. Login
3. Open Javascript Console
4. See all the 404 error messages

### Expected behavior
Posts images should be visible

### Additional Context

Screenshot - 2019-06-13T155537 115

roschaefer:
Ok, so after I would have to use the same method in three different
locations (`<ds-card>` expects an `image` attribute but cannot render
entire components) I decided to implement the prefix of image urls with
a filter rather than a component. The downside of this is that we have
to add the filter on a lot of component tests. The benefit is less
components and hopefully less complexity.

### Issues

- Fix #820


webapp/components/Avatar/Avatar.vue
```diff
@@ -1,5 +1,10 @@
 <template>
-  <ds-avatar :image="avatarUrl" :name="userName" class="avatar" :size="size" />
+  <ds-avatar
+    :image="user && user.avatar | proxyApiUrl"
```
mattwr18: I needed to check what the `|` means in this case @roschaefer 
it's a bitwise OR, correct?
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Bitwise_Operators#.7c_%28Bitwise_OR%29
can you explain why you used it in this case?

roschaefer: No, this is not a bitwise OR. It's the use of a vue filter in a template. See https://vuejs.org/v2/guide/filters.html

roschaefer:
@mattwr18 the only way I tested this was to check the 404 errors in the console. I couldn't find any 404 error with a URL not having the `/api` prefix, so I believe this should be OK. The dump that was used is the production data with URLs pointing to the uploads folder.

PR #823, opened 2019-06-13
Whitelist queries/mutations, fix tests

### LINKED ISSUE #343: 🐛 Whitelist all auto-generated Mutations/Queries
**Describe the bug**
Disable all Mutations/Queries via middleware and recheck which can be enabled under which circumstance.

We use [neo4j-graphql-js](https://github.com/neo4j-graphql/neo4j-graphql-js) as a query builder. None of the generated mutations/queries has authentication/authorization by default. Therefore we need to white-list only those mutations that we need and which are tested on security we before migrating our [production server](https://alpha.human-connection.org/) to the new nitro code base.

**To Reproduce**
1. Go to 'http://localhost:4000'
2. Create e.g. a comment:
```graphql
mutation {
  CreateComment(content: "I am a comment") {
    id
  }
}
```
3. Click submit
3. You just created a comment without an author.

**Expected behavior**
No private data should be exposed, ie. emails of users. No disabled or deleted content should be visible. No data should be created without authorship of the current user.

roschaefer:
## 🍰 Pullrequest

This Pull Request throw an authorization error on all Queries/Mutations by default, unless we explicitly specifiy how authorization should behave.
The types on the other hand will be allowed by default. The types are checked *after* the resolver of the query/mutation has been called, the db requests have resolved and the response is on its way out. We e.g. have some extra checks on the `email`, `password`, and `privateKey` field of the user, so those attributes can never be read.

@Mastercuber there are two failing cucumber scenarios now. I disabled these cucumber scenarios for reasons I explained in the commit message.

### Issues

- close #343 

### Todo

- [ ] Refactor ActivityPub middleware (will be done when we have a look into #541)


backend/src/middleware/permissionsMiddleware.js
```diff
@@ -48,6 +54,13 @@ const belongsToMe = rule({
   return Boolean(notification)
 })
 
+/* TODO: decide if we want to remove this check: the check
+ * `onlyEnabledContent` throws authorization errors only if you have
+ * arguments for `disabled` or `deleted` assuming these are filter
+ * parameters. Soft-delete middleware obfuscates data on its way out
+ * anyways. Furthermore, `neo4j-graphql-js` offers many ways to filter for
+ * data so I believe, this is not a good check anyways.
+ */
```
roschaefer: What do we do about `enabledContent`? Shall I remove it?

roschaefer:
@ogerly this PR is not related to images :thinking:

backend/src/middleware/permissionsMiddleware.js
ulfgebhardt: This clearly depends on who does the filtering - if its the server filtering stuff its fine to remove the check. If the filter is defined by the client and not enforced by the server, we need to make sure only content is delivered which the requesting user is allowed to see.

roschaefer: The 100% security that the users see only what he is allowed to see is **not** handled by this check.

The check does sth. else: If you filter for posts that are deleted **only**, you will get an authentication error right away in case you are not an administrator. But if we remove this check, then you would get a result set with all deleted posts obfuscated. Same effect but different behaviour.

The obfuscation is the 100% security, because it's implemented on the post itself, not the top-level query. So even if you request nested posts (e.g. all posts of a user) you will see the obfuscated data.

PR #831, opened 2019-06-14
Turn off $root listeners in beforeDestroy

### LINKED ISSUE #804: 🐛 [Bug] $root.$on with no $root.$off
## :bug: Bugreport

While we are moving away from using `$root.$on` to listen for root emitted events, in the short term we should at least follow good practice and have a `$root.$off` in a `beforeDestroy` lifecycle hook

mattwr18:
### Issues

- fixes #804 

long-term we want to avoid these types of events, but in the short-term, or where necessary, we should be turning these off in a `beforeDestroy` lifecycle method

### DIFF
--- a/webapp/components/Editor/index.vue
+++ b/webapp/components/Editor/index.vue
@@ -323,6 +323,7 @@ export default {
     })
   },
   beforeDestroy() {
+    this.$root.$off('changeLanguage')
     this.editor.destroy()
   },
   methods: {
--- a/webapp/components/comments/CommentList/index.vue
+++ b/webapp/components/comments/CommentList/index.vue
@@ -54,6 +54,9 @@ export default {
       this.refetchPostComments()
     })
   },
+  beforeDestroy() {
+    this.$root.$off('refetchPostComments')
+  },
   methods: {
     refetchPostComments() {
       if (this.$apollo.queries.Post) {


roschaefer:
@mattwr18 can you give me and others a little more context why this is necessary and what it changes? Maybe with some links into documentation?

mattwr18:
sure @roschaefer here are the docs https://vuejs.org/v2/api/#vm-off

and in this medium blog it says `Also, you need to make sure that the event listener is turned off before you destroy the component.`

https://medium.com/vuejobs/create-a-global-event-bus-in-vue-js-838a5d9ab03a
although, doesn't say why it's important

this blog says you need to do it as well https://alligator.io/vuejs/global-event-bus/

Finally, I saw that @Tirokk was getting error messages when switching his locale `this.docView is null`, which led me to this issue https://discuss.prosemirror.net/t/what-could-create-a-null-docview/1830, where one of the people responds `Calling .destroy() on an editor does that.`

there is only one place in our code base where we call `this.editor.destroy()`, https://github.com/Human-Connection/Human-Connection/blob/master/webapp/components/Editor/index.vue#L326, which happens to be where we are setting up one of these event listeners on line `321`

mattwr18:
yes, this is my understanding as well, and I agree... I don't feel great about this because ultimately I want to replace the two instances of these `$root.$emit/$root.$on` in the code base, but it might take slightly longer, so I thought I'd put in a temp fix....

I think actually the placeholder, we can use Vuex since updating the value of it will cause the placeholder to be reactive.

the other instance, I think updating the apollo cache is the answer. I'll merge this temporarily and work on removing them.

PR #834, opened 2019-06-15
801 reset password

### LINKED ISSUE #801: 🚀 [Feature] Reset password
## :rocket: Feature

Reset your password and get a mail to login again.

#### User story
```gherkin
As a user
I want to have a button to send me a reset password mail
To be able to login again in case I forgot my password
```

### Design & Layout

Screenshot - 2019-06-12T120331 736

roschaefer:
## 🍰 Pullrequest
Implement a basic password reset flow. I did not follow a security best practice here: You should ask the user for her username. That way, the account is safe even if an attacker has access to the email inbox. Reason for not implementing was that our users cannot choose their own slug at the moment. Therefore I believe we would make the whole flow pretty unusable if we ask for sth. that the user never changed and probably does not remember.

### Issues

- Close #801

### Todo

- [X] None


roschaefer:
This is still a little bit work-in-progress but I think this is the moment when I can ask you for a first review.

webapp/locales/de.json
```diff
@@ -8,11 +8,32 @@
...
+      "submitted": "Eine E-Mail mit weiteren Instruktionen wurde verschickt an <b>{email}</b>"
+    }
+  },
+  "verify-code": {
+    "form": {
+      "code": "Code eingeben",
+      "description": "Öffne Deine E-Mail Postfach und gib den Code ein, den wir geschickt haben.",
+      "next": "Weiter",
+      "change-password":{
+        "success": "Änderung des Passworts war erfolgreich",
```
Tirokk: `… war erfolgreich!` we usually put a `!` behind success texts I have the impression.

webapp/locales/en.json
```diff
@@ -8,11 +8,32 @@
...
+      "submitted": "A mail with further instruction has been sent to <b>{email}</b>"
+    }
+  },
+  "verify-code": {
+    "form": {
+      "code": "Enter your code",
+      "description": "Open your inbox and enter the code that we've sent to you.",
+      "next": "Continue",
+      "change-password": {
+        "success": "Changing your password was successful",
```
Tirokk: `… successful!` we usually put a ! behind success texts I have the impression.

backend/src/schema/resolvers/passwordReset.js
```diff
@@ -0,0 +1,72 @@
...
+      RETURN r
+      `
+      let transactionRes = await session.run(cypher, { stillValid, email, code, newHashedPassword })
+      const [reset] = transactionRes.records.map(record => record.get('r'))
+      const result = !!(reset && reset.properties.usedAt)
+      session.close()
+      return result
+    },
+  },
+}
```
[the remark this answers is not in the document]
roschaefer: yes

backend/src/schema/resolvers/passwordReset.spec.js
```diff
@@ -0,0 +1,180 @@
...
+      password: '1234',
+    })
+  })
+
+  afterEach(async () => {
+    await factory.cleanDatabase()
+  })
+
+  describe('requestPasswordReset', () => {
+    const mutation = `mutation($email: String!) { requestPasswordReset(email: $email) }`
```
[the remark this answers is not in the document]
roschaefer: In tests? I usually don't do that - by looking nicer you mean the editor is doing syntax highlighting?

backend/src/schema/resolvers/passwordReset/emailTemplates.js
```diff
@@ -0,0 +1,85 @@
...
+  actionUrl.searchParams.set('email', email)
+
+  return {
+    to: email,
+    subject,
+    text: `
+Hi ${name}!
+
+You recently requested to reset your password for your Human Connection account.
+Use the button below to reset it. This password reset is only valid for the next
```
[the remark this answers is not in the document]
roschaefer: Yes, you're right

webapp/locales/de.json
roschaefer: :+1:

roschaefer:
@Tirokk does `yarn install --ignore-engines` resolve the problem? I have never seen that output on my machine and believe it's unrelated.

backend/src/schema/resolvers/passwordReset.js
```diff
@@ -0,0 +1,72 @@
...
+    },
+    resetPassword: async (_, { email, code, newPassword }, { driver }) => {
+      const session = driver.session()
+      const stillValid = new Date()
+      stillValid.setDate(stillValid.getDate() - 1)
+      const newHashedPassword = await bcrypt.hashSync(newPassword, 10)
+      const cypher = `
+      MATCH (pr:PasswordReset {code: $code})
+      MATCH (u:User {email: $email})-[:REQUESTED]->(pr)
+      WHERE duration.between(pr.issuedAt, datetime()).days <= 0 AND pr.usedAt IS NULL
```
mattwr18: Maybe I'm missing something... the duration between when the password reset was issued and now in days is `less` than or equal to 0? shouldn't it be 1??

roschaefer: The duration will be a full day. So `<=1` would mean 1 day + 23 hours + 59 minutes + 59 seconds.

backend/src/schema/resolvers/passwordReset.spec.js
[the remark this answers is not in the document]
Tirokk: In tests! Yes!
Test should be good readable as well, or?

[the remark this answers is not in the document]
mattwr18: I think the point is that @roschaefer doesn't use VS code

backend/src/schema/resolvers/passwordReset.js
mattwr18: Sorry, maybe I am being dense, but how is less than or equal to 1 mean 1+?

PR #836, opened 2019-06-15
Set up editor placeholder to use Vuex

### LINKED ISSUE #835: 🐛 [Bug] Use Vuex to make editor placeholder reactive
## :bug: Bugreport

Currently, we have a hack where we do a `$root.$emit` to change the placeholder text.
We want to avoid these event listeners in the code base, one way we can do that is using Vuex.

mattwr18:
- to make reactive
- safer than using global event buses ($root.$emit)

Co-authored-by: Wolfgang Huß <wolle.huss@pjannto.com>
Co-authored-by: Mike Aono <aonomike@gmail.com>

### Issues

- fixes #835 

We did this `PR` as part of the new `VueJS Mob` pair programming session, `Saturdays 3pm UTC`


mattwr18:
thanks @Tirokk 
@roschaefer I wrote a component test for `LocaleSwitch`, but I wanted to add a test that the `editor/SET_PLACEHOLDER_TEXT` is called when a user changes the locale, but I can't check that it's called unless it's a mock or a spy...
If I set it up as a mock, I get this console.error `[vuex] unknown mutation type: editor/SET_PLACEHOLDER_TEXT`

I could get rid of this error by importing the mutations from `editor/SET_PLACEHOLDER_TEXT`, but then I can't check to see that it's called.

Maybe, it's something simple, but it's not obvious to me, so I thought I'd work on something else and see what others thought... this `PR` has been approved and I think ups the test coverage, so maybe we merge, but I thought I'd check to see what you think...

PR #864, opened 2019-06-19
🐛 [Bug] :  long comments can be folded up and down

### LINKED ISSUE #844: 🐛 [Bug] Comments cannot be expanded if they are too long
### Comments cannot be expanded if they are too long

image

Comments are cut off with `...` but cannot be expanded to read the full comment.

1. Visit Post with Comments
2. See dots on long comments
3. Unable to expand
4. ...
5. Profit

### Expected behavior

Comments should be expandable to read them in full or be displayed in full to begin with.

### Version & Environment
 Type: [Desktop]      
 - OS: [?]      
 - Browser: [Chrome] 

### Additional context

Example: https://nitro-production.human-connection.org/post/5d0107b1cbac47002b176166/prof-rainer-mausfeld-gibt-es-hoffnung-fur-die-lammer

ogerly:
## 🍰 Pullrequest
long comments can be folded up and down

ezgif com-gif-maker

### Issues
 
- fixes #844 
 
- None

### Todo
- [X] None


webapp/components/Comment.vue
```diff
@@ -23,10 +23,27 @@
...
+    <div v-show="comment.content !== comment.contentExcerpt">
+      <span
+        v-if="!isActive"
+        style="padding-left: 40px; color: #17b53f; cursor:pointer"
+        @click="isActive = !isActive"
+      >
+        {{ $t('comment.show.more') }}
+      </span>
+    </div>
+    <div v-if="isActive" v-html="comment.content" style="padding-left: 40px;" />
```
roschaefer:
```suggestion
    <div v-else v-html="comment.content" style="padding-left: 40px;" />
```

webapp/components/Comment.vue
```diff
@@ -23,10 +23,27 @@
     <!-- eslint-disable vue/no-v-html -->
     <!-- TODO: replace editor content with tiptap render view -->
     <ds-space margin-bottom="small" />
-    <div style="padding-left: 40px;" v-html="comment.contentExcerpt" />
-    <!-- eslint-enable vue/no-v-html -->
+    <div v-if="!isActive" v-html="comment.contentExcerpt" style="padding-left: 40px;" />
```
roschaefer: The logic implemented in this template is calling for it's own component. How about `<collapsible>`?

webapp/components/Comment.vue
```diff
@@ -23,10 +23,27 @@
...
+        {{ $t('comment.show.more') }}
+      </span>
+    </div>
+    <div v-if="isActive" v-html="comment.content" style="padding-left: 40px;" />
+    <span
+      v-if="isActive"
+      @click="isActive = !isActive"
+      style="padding-left: 40px; color: #17b53f; cursor:pointer"
+    >
+      {{ $t('comment.show.less') }}
```
roschaefer: Could you DRY this out? (DRY == don't repeat yourself, here is a [link](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself)). I think if you put the `isCollapsed` property to the root node, that should be it.

webapp/components/Comment.vue
```diff
@@ -23,10 +23,27 @@
     <!-- eslint-disable vue/no-v-html -->
     <!-- TODO: replace editor content with tiptap render view -->
     <ds-space margin-bottom="small" />
-    <div style="padding-left: 40px;" v-html="comment.contentExcerpt" />
-    <!-- eslint-enable vue/no-v-html -->
+    <div v-if="!isActive" v-html="comment.contentExcerpt" style="padding-left: 40px;" />
+    <div v-show="comment.content !== comment.contentExcerpt">
+      <span
+        v-if="!isActive"
```
roschaefer: Call the property `isCollapsed`. From the name `active` would expect only one comment to be active at a time.But `isActive` refers only to sth. visible here.

backend/src/seed/factories/reports.js
```diff
@@ -8,10 +8,12 @@ export default function create(params) {
       mutation($id: ID!, $description: String!) {
         report(description: $description, id: $id) {
           id
-          createdAt
```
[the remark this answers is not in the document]
ogerly: that always caused errors when saving.  ... I made the mistake at the place and took it out ...

mattwr18:
should we pair on testing this @ogerly

webapp/components/Comment.vue
Tirokk: Is done.

Tirokk: There is a show in-between. So this easy and quick solution is not working.

webapp/components/Comment.vue
Tirokk: We have to refactor this anyway by using the editor library to display the content very soon. We should do this than together …

webapp/components/Comment.vue
Tirokk: Same same … as said before. Not just this …

webapp/components/Comment.vue
Tirokk: Done

Tirokk:
Please pull in the master, than I will merge …

PR #879, opened 2019-06-21
polish locales

fatherJS:
pl.json file corrected


roschaefer:
ah right! Travis never runs build for the branch of forks.. I'l merge manually

PR #892, opened 2019-06-24
🍰 2019/kw25/split_graphql_schema

ulfgebhardt:
## 🍰 Pullrequest

split graphql schema

### Issues

- None

### Todo

- [X] None


ulfgebhardt:
I was undecided about this aswell - I decided to split it as much as possible - it is still reversable ;-)

appinteractive:
My approach was to split it by model/topic not by crud. So in theory there should be less files. It’s a more modular approach. I think.

ulfgebhardt:
image
thats my View on this ;-)

ulfgebhardt:
feel free to use this PR to refactor as needed

roschaefer:
Can we close this? I also prefer the CRUDs in one file.

I think I'm gonna close this. If we really need this, we can still re-open.

PR #893, opened 2019-06-24
Deploy development mailserver

roschaefer:
## 🍰 Pullrequest
Why not deploying a dummy mailserver? It's already deployed [here](https://nitro-mailserver.human-connection.org/#/) this PR will add documentation.

### Issues

- None

### Todo

- [X] None


roschaefer:
@mattwr18 since it is already deployed, the log output is expected. Good that you tried it out though! :+1:

PR #894, opened 2019-06-24
Spike use of casual as seed data generator

mattwr18:
this is a spike related to an ongoing issue with lorempixel.com being unavailable

see what you think @roschaefer @ulfgebhardt @Tirokk @ogerly


mattwr18:
closing this as we got `unsplash` working with faker

PR #935, opened 2019-06-29
Patch cypher injection vulnerability

roschaefer:
@mattwr18 this patches the vulnerability. However we should never do
string interpolation with user input.

### Issues

- None

### Todo

- [X] None


Tirokk:
Good that you have seen this vulnerability @roschaefer !!! 😍

The code was not working as I tried a delete of my account. There was a `String` not replaced with `Deletable`. Hope I have not missed anything.
I just pushed that change up.

PR #944, opened 2019-07-01
🍰 2019/kw27/data_import_emotions

ulfgebhardt:
## 🍰 Pullrequest

Import all Emotions on legacy migration

### Issues

- relates #692 

### Todo

- [X] None


deployment/legacy-migration/maintenance-worker/migration/neo4j/emotions/emotions.cql
```diff
@@ -5,31 +5,54 @@
...
-CALL apoc.load.json("file:${IMPORT_CHUNK_PATH_CQL_FILE}") YIELD value as emotion;
+CALL apoc.load.json("file:${IMPORT_CHUNK_PATH_CQL_FILE}") YIELD value as emotion
+MATCH (u:User {id: emotion.userId}),
+      (c:Post {id: emotion.contributionId})
+MERGE  (u)-[e:EMOTED {
+          id:        emotion._id["$oid"],
+          emotion:   emotion.rated,
+          createdAt: datetime(emotion.createdAt.`$date`),
+          updatedAt: datetime(emotion.updatedAt.`$date`)
+        }]->(c)
```
roschaefer: `neo4j-graphql-js` does not support model relationship properties I guess we want to display those emotions in a query, so I think you have to model each `emotion` with a separate relationship type.

roschaefer: How about `REACTED_HAPPILY`, `REACTED_SADLY`, etc. ?

ulfgebhardt: No thats not a proper solution - we don' t need support of neo4j-graphql-js from this, since we can query manually

// Queries
  // user sets an emotion emotion:
  // MERGE (u)-[e:EMOTED {id: ..., emotion: "funny", createdAt: ..., updatedAt: ...}]->(c)
  // user removes emotion
  // MATCH  (u)-[e:EMOTED]->(c) DELETE e
  // contribution distributions over every `emotion` property value for one post
  // MATCH (u:User)-[e:EMOTED]->(c:Post {id: "5a70bbc8508f5b000b443b1a"}) RETURN e.emotion,COUNT(e.emotion)
  // contribution distributions over every `emotion` property value for one user (advanced - "whats the primary emotion used by the user?")
  // MATCH (u:User{id:"5a663b1ac64291000bf302a1"})-[e:EMOTED]->(c:Post) RETURN e.emotion,COUNT(e.emotion)
  // contribution distributions over every `emotion` property value for all posts created by one user (advanced - "how do others react to my contributions?")
  // MATCH (u:User)-[e:EMOTED]->(c:Post)<-[w:WROTE]-(a:User{id:"5a663b1ac64291000bf302a1"}) RETURN e.emotion,COUNT(e.emotion)
  // if we can filter the above an a variable timescale that would be great (should be possible on createdAt and updatedAt fields)

ulfgebhardt: The given solution can fullfill all our needs of querying as shown above

ulfgebhardt: Also the documentation references it - therefore it should be possible
image

backend/src/schema/types/type/EMOTED.gql
```diff
@@ -0,0 +1,10 @@
+type EMOTED @relation(name: "EMOTED") {
+  from: User
+  to: Post
+
+  emotion: String
```
roschaefer:
```suggestion
  emotion: EmotionEnum
```

PR #945, opened 2019-07-01
943 bring user mutation under control

### LINKED ISSUE #943: 🐛 [Bug] Filter feature exposing email
## :bug: Bugreport
If you use `neo4j-graphlq-js` feature to filter attributes in e.g. a `email_contains` in a `User` query, you can basically make an email listing. This is also possible for password hashes. 

1. Go to https://nitro-staging.human-connection.org/api
2. Enter the following query:
```graphql
{
  User(filter: { email_contains: "user" }) {
    name
  }
}

```
3. See all users matching the query
4. Repeat until you deciphered the whole email address

### Expected behavior
It should not be possible to filter for emails unless you have admin privileges *and* you know the email address.

roschaefer:
### Issues

- None

### Todo

- [X] None


backend/src/schema/resolvers/registration.spec.js
```diff
@@ -0,0 +1,395 @@
...
+        CreateInvitationCode: { token: expect.any(String) },
+      })
+    })
+
+    it('creates an InvitationCode with a `createdAt` attribute', async () => {
+      await action()
+      const codes = await instance.all('InvitationCode')
+      const invitation = await codes.first().toJson()
+      expect(invitation.createdAt).toBeTruthy()
+      expect(Date.parse(invitation.createdAt)).toEqual(expect.any(Number))
```
mattwr18: why do we need two expectations here?

backend/src/schema/resolvers/registration.spec.js
```diff
@@ -0,0 +1,395 @@
...
+              return client.request(mutation)
+            }),
+          )
+          return client.request(mutation, variables)
+        }
+      })
+
+      describe('as ordinary `user`', () => {
+        it('throws `Not Authorised` because of maximum number of invitations', async () => {
+          await expect(action()).rejects.toThrow('Not Authorised')
```
mattwr18: Is there not a way we can customize this message? it's not clear why they wouldn't be able to...

backend/src/schema/resolvers/registration.spec.js
```diff
@@ -0,0 +1,395 @@
...
+        })
+
+        it('creates no additional invitation codes', async done => {
+          try {
+            await action()
+          } catch (e) {
+            const invitationCodes = await instance.all('InvitationCode')
+            await expect(invitationCodes.toJson()).resolves.toHaveLength(3)
```
mattwr18: why is this expectation in catch?

backend/src/schema/resolvers/registration.spec.js
```diff
@@ -0,0 +1,395 @@
...
+
+  it.todo('throws Authorization error')
+
+  describe('with invalid InvitationCode', () => {
+    beforeEach(() => {
+      variables.token = 'wut?'
+    })
+
+    it.todo('throws UserInputError')
+  })
```
mattwr18: I like that you have added these as todos, but it's not clear to me what is keeping you from writing the tests now

backend/src/schema/resolvers/registration.spec.js
```diff
@@ -0,0 +1,395 @@
...
+      })
+    })
+
+    describe('given a valid email', () => {
+      beforeEach(() => {
+        variables.email = 'someUser@example.org'
+      })
+
+      it('resolves', async () => {
+        await expect(action()).resolves.toEqual({
```
mattwr18: I don't want to be pedantic, and I acknowledge that we each have different styles, but it takes me a little longer to verify what is happening here because the variables are all named the same `action` in this case does what? I need to scroll up to check.
consider naming it `signupByInvitationAction`, etc...

backend/src/schema/resolvers/registration.spec.js
```diff
@@ -0,0 +1,395 @@
...
+          const emailAddress = await emailAddresses.first().toJson()
+          expect(emailAddress.createdAt).toBeTruthy()
+          expect(Date.parse(emailAddress.createdAt)).toEqual(expect.any(Number))
+        })
+
+        it('with a cryptographic `nonce`', async () => {
+          await action()
+          const emailAddresses = await instance.all('EmailAddress')
+          const emailAddress = await emailAddresses.first().toJson()
+          expect(emailAddress.nonce).toEqual(expect.any(String))
```
mattwr18: how does this check that it is `cryptographic`? I guess there isn't a way to check that, so you must just check that it is any string?

backend/src/schema/resolvers/registration.spec.js
```diff
@@ -0,0 +1,395 @@
...
+
+        it('connects inviter through invitation code', async () => {
+          await action()
+          const result = await instance.cypher(
+            'MATCH(inviter:User)-[:GENERATED]->(:InvitationCode)-[:ACTIVATED]->(email:EmailAddress {email: {email}}) RETURN inviter',
+            { email: 'someuser@example.org' },
+          )
+          const inviter = instance.hydrateFirst(result, 'inviter', instance.model('User'))
+          await expect(inviter.toJson()).resolves.toEqual(
+            expect.objectContaining({ name: 'Inviter' }),
```
mattwr18: I really like your name choice for the name of the user who invited the other.... you could have just used random names like Torsten, Wolfgang, etc... but then it would make it more difficult to remember who is who... thank you

backend/src/schema/resolvers/registration.spec.js
```diff
@@ -0,0 +1,395 @@
...
+            await factory.create('User', { email: 'someuser@example.org' })
+          })
+
+          it('throws unique violation error', async () => {
+            await expect(action()).rejects.toThrow('User account with this email already exists.')
+          })
+        })
+
+        describe('if the EmailAddress already exists but without user account', () => {
+          it.todo('decide what to do')
```
mattwr18: how would this happen?

backend/src/schema/resolvers/registration.spec.js
```diff
@@ -0,0 +1,395 @@
...
+            instance
+              .model('User')
+              .create({ name: 'Somebody', password: '1234', email: 'john@example.org' }),
+          ])
+          await emailAddress.relateTo(user, 'belongsTo')
+        })
+
+        describe('sending a valid nonce', () => {
+          beforeEach(() => {
+            variables.nonce = '123456'
```
mattwr18: why do you need to reset the nonce here?

backend/src/schema/resolvers/registration.spec.js
```diff
@@ -0,0 +1,395 @@
...
+        })
+      })
+
+      describe('disconnected EmailAddress exists', () => {
+        beforeEach(async () => {
+          const args = {
+            email: 'john@example.org',
+            nonce: '123456',
+          }
+          await instance.model('EmailAddress').create(args)
```
mattwr18: should we say that an `EmailAddress` must be related to a `User`?

backend/src/schema/resolvers/socialMedia.spec.js
```diff
@@ -98,14 +98,19 @@ describe('SocialMedia', () => {
       const variables = {
         url: '',
       }
-      await expect(client.request(mutationC, variables)).rejects.toThrow('Input is not a URL')
+      await expect(client.request(mutationC, variables)).rejects.toThrow(
```
mattwr18: I know you didn't add this, but I have no idea what `mutationC` is in this case
I would need to open up this file to see, which when reviewing a big `PR` makes it even more time-consuming
Not to harp on the same point, but if we get in the habit of instead naming it something generic like `mutation` or `mutationA` then I think we can help ourselves and others to read the code faster.
Or maybe others don't agree with me, in which case I can make an effort to deal with it and not continuously bring it up.

backend/src/schema/resolvers/user_management.js
```diff
@@ -5,50 +5,39 @@ import { neo4jgraphql } from 'neo4j-graphql-js'
...
   Query: {
-    isLoggedIn: (parent, args, { driver, user }) => {
+    isLoggedIn: (_, args, { driver, user }) => {
       return Boolean(user && user.id)
     },
     currentUser: async (object, params, ctx, resolveInfo) => {
       const { user } = ctx
       if (!user) return null
-      return neo4jgraphql(object, { id: user.id }, ctx, resolveInfo, false)
+      return neo4jgraphql(object, { id: user.id }, ctx, resolveInfo, true)
```
mattwr18: I think we want to change this back to false, no?

backend/src/schema/resolvers/user_management.js
```diff
@@ -5,50 +5,39 @@ import { neo4jgraphql } from 'neo4j-graphql-js'
...
-          'RETURN user {.id, .slug, .name, .avatar, .email, .password, .role, .disabled} as user LIMIT 1',
+          'RETURN user {.id, .slug, .name, .avatar, .email, .encryptedPassword, .role, .disabled} as user LIMIT 1',
         {
           userEmail: email,
         },
       )
-
       session.close()
-      const [currentUser] = await result.records.map(function(record) {
+      const [currentUser] = await result.records.map(record => {
```
mattwr18: much prefer using arrow functions where possible :smile_cat:

backend/src/schema/resolvers/users.spec.js
```diff
@@ -11,51 +11,32 @@ afterEach(async () => {
...
-        it('is allowed to create new users', async () => {
-          const expected = {
-            CreateUser: {
-              id: expect.any(String),
-            },
-          }
-          await expect(client.request(mutation, variables)).resolves.toEqual(expected)
-        })
+      it('is permitted', async () => {
+        await expect(client.request(query, variables)).resolves.toEqual({ User: [] })
```
mattwr18: would it be better to actually create a user for these test cases, or you are happy that it either throws an authorization error, or not?

backend/src/schema/resolvers/users.spec.js
```diff
@@ -119,12 +100,12 @@ describe('users', () => {
         await expect(client.request(mutation, variables)).resolves.toEqual(expected)
```
mattwr18: don't know what this is testing...

backend/src/schema/types/type/User.gql
```diff
@@ -3,20 +3,16 @@ type User {
...
   avatar: String
   coverImg: String
-  avatarUpload: Upload
   deleted: Boolean
   disabled: Boolean
   disabledBy: User @relation(name: "DISABLED", direction: "IN")
   role: UserGroup
   publicKey: String
-  privateKey: String
-
```
mattwr18: this is why you removed the tests that @Tirokk added to check that we can't ask for a user's `privateKey`
meaning if it is removed from the type, we are unable to ask for it.

package.json
```diff
@@ -29,6 +30,8 @@
     "faker": "Marak/faker.js#master",
     "graphql-request": "^1.8.2",
     "neo4j-driver": "^1.7.5",
-    "npm-run-all": "^4.1.5"
+    "neode": "^0.2.16",
+    "npm-run-all": "^4.1.5",
+    "slug": "^1.1.0"
```
mattwr18: great that you only needed to add only two packages to the root-level package.json :smile_cat:

webapp/plugins/vue-filters.js
```diff
@@ -39,15 +38,6 @@ export default ({ app = {} }) => {
...
     },
-    // format seconds or milliseconds to durations HH:mm:ss
-    duration: (value, unit = 's') => {
-      if (unit === 'ms') {
-        value = value / 1000
-      }
-      return value
-        ? format(addSeconds(new Date('2000-01-01 00:00'), value), 'HH:mm:ss')
-        : '00:00:00'
-    },
```
mattwr18: nice that you figured out we don't use this, removing unused dependencies should be an ongoing task

backend/src/middleware/permissionsMiddleware.js
```diff
@@ -70,6 +70,29 @@ const onlyEnabledContent = rule({
...
+      `
+      MATCH (user:User {id:$id})-[:GENERATED]->(i:InvitationCode)
+      RETURN COUNT(i) as count
+      `,
+      { id: user.id },
+    )
+    const [count] = result.records.map(record => {
+      return record.get('count').toNumber()
+    })
+    return count >= 3
```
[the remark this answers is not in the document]
roschaefer: Oh, yes, you're right!

backend/src/middleware/validation/index.js
```diff
@@ -1,16 +1,5 @@
...
-
-const validateUsername = async (resolve, root, args, context, info) => {
-  if (!('name' in args) || (args.name && args.name.length >= USERNAME_MIN_LENGTH)) {
-    /* eslint-disable-next-line no-return-await */
-    return await resolve(root, args, context, info)
-  } else {
-    throw new UserInputError(`Username must be at least ${USERNAME_MIN_LENGTH} characters long!`)
-  }
-}
-
```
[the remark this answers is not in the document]
roschaefer: No, this is basically a refactoring of the existing code base. For some strange reason we want to force our users to type in a 3-letter minimum username.

backend/src/schema/resolvers/registration.js
```diff
@@ -0,0 +1,107 @@
...
+import { neode } from '../../bootstrap/neo4j'
+import fileUpload from './fileUpload'
+import encryptPassword from '../../helpers/encryptPassword'
+
+const instance = neode()
+
+/*
+ * TODO: remove this function as soon type `User` has no `email` property
+ * anymore
+ */
```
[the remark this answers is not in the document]
roschaefer: I made a refactoring to the type `User` and left it at a dangerously in-between state. So, right now, the `User` has an attribute `email`. The idea of the the signup is, however, that you add an `EmailAddress` *without* a user account and then create the user account and connect it with the orphaned email address. This `email` attribute is a left-over. It should be completely removed at some point. But I think it's OK for now to keep it, it doesn't break anything and since we're free to create our own resolvers, we can also have a `User.email` resolver, which basically just `MATCH(u:User)-[:PRIMARY_EMAIL]->(e:EmailAddress)` pulls in the email. This is also a preparation for having multiple email addresses.

backend/src/schema/resolvers/registration.spec.js
roschaefer: I just want to make sure that the string returned from the backend is a valid ISO datestring. Apparently the way to check it is to parse it.

backend/src/schema/resolvers/users.spec.js
roschaefer: yes, correct. It's a refactoring but I think I can easily add more meaning to the test.. hang on

backend/src/schema/resolvers/users.spec.js
roschaefer: I think it is testing that you cannot alter other user accounts, only your own.

backend/src/schema/types/type/User.gql
roschaefer: Exactly, it is absolutely silly to declare it on the `User` type and then to make sure it's never exposed. Only because of `neo4j-graphql-js` it was there.

backend/src/schema/resolvers/registration.spec.js
mattwr18: sure, but then can we remove the `toBeTruthy()` expectation?

roschaefer: no, because `Date.parse(null)` returns a valid date

roschaefer: oh, no, it's `NaN`

roschaefer: new Date(null) is a valid date though

roschaefer: anyhow, I think for safety I keep it there, OK?

backend/src/schema/resolvers/registration.js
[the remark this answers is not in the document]
mattwr18: great, thanks for the explanation

backend/src/schema/resolvers/registration.spec.js
mattwr18: of course!
I know that you avoid multiple expectations in the same it block, just curious why you went against your normal workflow. thanks for the explanation

backend/src/schema/types/type/User.gql
mattwr18: great! another really nice side effect to not using it in this use case :smile_cat:

backend/src/middleware/validation/index.js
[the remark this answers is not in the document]
Tirokk: So we allow one character usernames yet?

backend/src/schema/resolvers/user_management.js
```diff
@@ -5,7 +5,7 @@ import { neo4jgraphql } from 'neo4j-graphql-js'
 
 export default {
   Query: {
-    isLoggedIn: (parent, args, { driver, user }) => {
+    isLoggedIn: (_, args, { driver, user }) => {
```
Tirokk: I would prefer `_parent` here, to keep clear where this parameter stands for. Just my opinion …

backend/src/schema/resolvers/user_management.js
```diff
@@ -15,40 +15,29 @@ export default {
...
-      //   throw new Error('Another User with same email exists.')
-      // }
-      // data[email] = {
-      //   password: await bcrypt.hashSync(password, 10),
-      // }
-
-      return true
-    },
-    login: async (parent, { email, password }, { driver, req, user }) => {
+    login: async (_, { email, password }, { driver, req, user }) => {
```
Tirokk: Same same …

backend/src/schema/resolvers/user_management.spec.js
```diff
@@ -311,121 +310,3 @@ describe('change password', () => {
...
-
-  describe('authenticated query of "privateKey"', () => {
-    it('throws "Not Authorised!"', async () => {
-      await generateUserWithKeys(authenticatedClient)
-      await expect(
-        client.request(queryUserPrivateKey, { queriedUserSlug: 'apfel-strudel' }),
-      ).rejects.toThrow('Not Authorised')
-    })
-  })
-})
```
Tirokk: All my sweat and blood code gets deleted 😱😢😂

backend/src/schema/resolvers/users.js
```diff
@@ -1,15 +1,84 @@
...
+  Mutation: {
+    UpdateUser: async (object, args, context, resolveInfo) => {
+      args = await fileUpload(args, { file: 'avatarUpload', url: 'avatar' })
+      try {
+        let user = await instance.find('User', args.id)
+        if (!user) return null
+        await user.update(args)
+        return user.toJson()
+      } catch (e) {
+        throw new UserInputError(e.message)
```
Tirokk: Same same …

backend/src/schema/resolvers/users.js
```diff
@@ -34,4 +103,43 @@ export default {
...
+    ...count({
+      contributionsCount: '-[:WROTE]->(related:Post)',
+      friendsCount: '<-[:FRIENDS]->(related:User)',
+      followingCount: '-[:FOLLOWS]->(related:User)',
+      followedByCount: '<-[:FOLLOWS]-(related:User)',
+      commentsCount: '-[:WROTE]->(r:Comment)',
+      commentedCount: '-[:WROTE]->(:Comment)-[:COMMENTS]->(related:Post)',
+      shoutedCount: '-[:SHOUTED]->(related:Post)',
+      badgesCount: '<-[:REWARDED]-(related:Badge)',
+    }),
```
Tirokk: It was really interesting and fun to see how you did that stuff @roschaefer 🤗😎

Tirokk:
So we lost Tick, Trick and Track … 😢

mattwr18:
> So we lost Tick, Trick and Track …

Maybe we could add `Huey, Dewey, and Louie`??

backend/src/schema/resolvers/users.js
mattwr18: agreed!

backend/src/schema/resolvers/user_management.spec.js
mattwr18: I thought the same hehe
but from a safety perspective, a big improvement :smile_cat:

backend/src/schema/resolvers/user_management.js
mattwr18: I agree with you, I think it's not such a big deal cause there are other examples in the code base where someone looking for the naming would find it, but if we were to refactor all of them to have just an underscore, it would make it more difficult for people unfamiliar with the name convention to figure out what was going on

backend/src/middleware/permissionsMiddleware.js
```diff
@@ -70,6 +70,29 @@ const onlyEnabledContent = rule({
...
+      RETURN COUNT(i) >= 3 as limitReached
+      `,
+      { id: user.id },
+    )
+    const [limitReached] = result.records.map(record => {
+      return record.get('limitReached')
+    })
+    return limitReached
+  } catch (e) {
+    throw e
```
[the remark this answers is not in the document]
mattwr18: I tend to agree, or even `error`, what would we lose have two extra characters, however, I think people who are somewhat familiar with JavaScript will know what this means.

I feel like there is a fine line between allowing for different coding styles and adopting community standards that we enforce. Maybe we could talk in a `Human Connection Developer's Community Meeting`?

PR #953, opened 2019-07-02
Editor embeds

### LINKED ISSUE #256: [Editor] Embeds
As a user
I would like to see link previews and video embeds
In order to not have to click or leave the page to get the contents
---
We need to render `a` tags with a `data-url-embed` attribute an embed component. As a reference we have to look into the implementation of the todo-item node from the tiptap examples.
https://github.com/scrumpy/tiptap/blob/master/packages/tiptap-extensions/src/nodes/TodoItem.js
https://tiptap.scrumpy.io/todo-list

mattwr18:
### Issues

- fixes #256


mattwr18:
@appinteractive @roschaefer probably not the best solution, but I was struggling to figure out how to maintain the commit history of `Nitro-Embed` and it was getting complicated... what I ended up doing is using `git read-tree --prefix=embed/ embed/master`....
I can close this if we want to explore other options...

mattwr18:
closed in favor of https://github.com/Human-Connection/Human-Connection/pull/960

PR #956, opened 2019-07-02
Maintenance service to support maintenance mode

### LINKED ISSUE #906: 🚀 [Feature] Maintenance Mode
## :rocket: Feature

We need to have a page that user's see when we are doing maintenance on the site

mattwr18:
### Issues

- fixes #906 

@roschaefer, so far this is what I have done... I'm sure I've made some mistakes, but at least this is a start and maybe we can get it figured out together?

following the docs here: https://www.linode.com/docs/applications/containers/how-to-deploy-nginx-on-a-kubernetes-cluster/#deploy-nginx-on-the-kubernetes-cluster

I created a deployment with `kubectl --namespace=human-connection create deployment nginx --image=nginx`

Screenshot at 2019-07-02 19:40:39

Then, I created a service `kubectl --namespace=human-connection create service nodeport nginx --tcp=80:80`

Screenshot at 2019-07-02 19:41:51

it's info:
Screenshot at 2019-07-02 19:42:36

I then copied the `service-web.yaml` and changed the selector like it says here: https://www.nrmitchi.com/2017/11/easy-maintenance-mode-in-kubernetes/, but I noticed that the label doesn't match for example and also maybe the service should be named `maintenance`, as above.

Also, I had a doubt about this line `This isolated service contains only an nginx container, with our generic Maintenance page preloaded.`... 
If I am not mistaken the nginx container exists now, but how can I preload a maintenance page? With `apply` and some config? sshing into the nginx container and creating it?

Just a start...


deployment/human-connection/service-nginx.yaml
```diff
@@ -0,0 +1,14 @@
...
+  namespace: human-connection
+  labels:
+    human-connection.org/selector: deployment-human-connection-web
+spec:
+  ports:
+    - name: web
+      port: 3000
+      targetPort: 3000
+  selector:
+    human-connection.org/selector: nginx
```
roschaefer: I myself don't know exactly what connects the service with the number of deployed pods from your deployment. But by comparing `deployment-web.yaml` and `service-web.yaml` it must be this `spec.selector` which has to match.

deployment/human-connection/service-nginx.yaml
```diff
@@ -0,0 +1,14 @@
+apiVersion: v1
+kind: Service
+metadata:
+  name: nitro-web
```
roschaefer: We have a service with that name already
```suggestion
  name: maintenance
```

deployment/human-connection/service-nginx.yaml
```diff
@@ -0,0 +1,14 @@
...
+  namespace: human-connection
+  labels:
+    human-connection.org/selector: deployment-human-connection-web
+spec:
+  ports:
+    - name: web
+      port: 3000
+      targetPort: 3000
```
roschaefer: Shouldn't really matter if you just serve a single html page. You could choose whatever you want, port 80 sounds good to me in case of an ordinary webserver.

roschaefer:
@mattwr18 oh, I forgot one thing: The nginx should respond with 503!

@mattwr18 do you want to take care of that? And maybe make the page look a bit more fancy? :smile:

mattwr18:
yes, I do @roschaefer 
I'll have a look at the documentation as well, as I understood it instead of editing to add the `command: ["tail", etc...]`, we edit to swap out the name of the service, then change it back

mattwr18:
@roschaefer, I made the page mobile responsive and used the font family and line height from our styleguide... here is what it looks like:
Ipad:
Screenshot at 2019-07-29 09:13:02
Screenshot at 2019-07-29 09:12:47
Iphone:
Screenshot at 2019-07-29 09:12:37
Screenshot at 2019-07-29 09:12:17
Desktop:
Screenshot at 2019-07-29 09:12:07

Probably not award-winning stuff, but I think good enough for the maintenance page, until/if the marketing gets back to us with a different pic, or we'd like to change the message in some way. What do you think?

PR #960, opened 2019-07-03
Editor embeds merge in nitro embed

### LINKED ISSUE #256: [Editor] Embeds

mattwr18:
### Issues

- fixes #256


mattwr18:
does this look ok @roschaefer? I used `tomono`, but because I already had a repo with the name `Human-Connection`, I needed to remove it and recreate it with the merged commit history. This caused some small issues for me locally, but I think others won't face the same.

The files changed and commit history look good to me

roschaefer:
@mattwr18 this looks good, but could sb. explain me why this is a separate service and not just another resolver in our backend?

@appinteractive maybe?

mattwr18:
I think it could be, I thought the same thing myself @roschaefer

backend/src/schema/resolvers/embeds.spec.js
```diff
@@ -0,0 +1,216 @@
...
+              title: 'Baby Loves Cat',
+              author: 'Merkley Family',
+              publisher: 'YouTube',
+              date: expect.any(String),
+              description:
+                'She’s incapable of controlling her limbs when her kitty is around. The obsession grows every day. Ps. That’s a sleep sack she’s in. Not a starfish outfit. Al...',
+              url: 'https://www.youtube.com/watch?v=qkdXAtO40Fo',
+              image: 'https://i.ytimg.com/vi/qkdXAtO40Fo/maxresdefault.jpg',
+              audio: null,
+              video: null,
```
roschaefer: `audio` and `video` was always `null` in all my tests...

backend/src/schema/resolvers/embeds.spec.js
```diff
@@ -0,0 +1,216 @@
...
+    author_url: 'https://www.youtube.com/channel/UC5P8yei950tif7UmdPpkJLQ',
+  }),
+)
+
+describe('Query', () => {
+  describe('embed', () => {
+    let embedAction
+
+    beforeEach(() => {
+      embedAction = async variables => {
```
mattwr18: :heart: the naming convention @roschaefer

backend/src/schema/resolvers/embeds.spec.js
```diff
@@ -0,0 +1,216 @@
...
+
+    describe('given a Facebook link', () => {
+      beforeEach(() => {
+        fetch
+          .mockReturnValueOnce(Promise.resolve(new Response(HumanConnectionOrg)))
+          .mockReturnValueOnce(Promise.resolve('invalid json'))
+        variables = { url: 'https://www.facebook.com/HumanConnectionOrg/' }
+      })
+
+      it('does not crash if embed provider returns invalid JSON', async () => {
```
mattwr18: Any idea when providers return invalid JSONs and why?

backend/src/schema/resolvers/embeds.spec.js
mattwr18: strange, that was what I observed as well

backend/src/schema/resolvers/embeds/scraper.js
```diff
@@ -0,0 +1,102 @@
...
+
+// some providers allow a format parameter
+// we need JSON
+oEmbedProvidersFile = oEmbedProvidersFile.replace('{format}', 'json')
+
+const oEmbedProviders = JSON.parse(oEmbedProvidersFile)
+
+const fetchEmbed = async url => {
+  const provider = oEmbedProviders.find(provider => {
+    return provider.provider_url.includes(url.hostname)
```
mattwr18: can you explain where you got this long list of providers? how does this work actually? we added one of @Tirokk personal sites and it returned the metadata, but not sure how any of the hard-coded providers include his url hostname :thinking:

backend/src/schema/resolvers/embeds/scraper.js
```diff
@@ -0,0 +1,102 @@
...
+  return {
+    sources: ['resource'],
+    ...resource,
+  }
+}
+
+export default async function scrape(url) {
+  url = new URL(url)
+  if (url.hostname === 'youtu.be') {
+    // replace youtu.be to get proper results
```
mattwr18: thank you for your strategic use of comments @roschaefer

backend/src/schema/resolvers/users.js
```diff
@@ -36,16 +37,6 @@ const count = obj => {
...
-const undefinedToNull = list => {
-  const resolvers = {}
-  list.forEach(key => {
-    resolvers[key] = async (parent, params, context, resolveInfo) => {
-      return typeof parent[key] === 'undefined' ? null : parent[key]
-    }
-  })
-  return resolvers
-}
-
```
mattwr18: Nice refactoring :smile_cat:

backend/src/schema/resolvers/embeds.spec.js
appinteractive: Is should be filmed when there is video or audio contend in an article as far as I understand metascraper.

backend/src/schema/resolvers/embeds/scraper.js
appinteractive: oEmbed is a specefication which Sites can use for embeds but don’t have to. It’s a way I enhanced the metascraper results with data directly from the providers. There is an official oEmbed list which contains a lot of sites which supports that. Does that answer your question?

backend/src/schema/resolvers/embeds.spec.js
appinteractive: *filled

PR #969, opened 2019-07-03
Filter posts by category

### LINKED ISSUE #967: 🚀 [Feature] Filter Posts by Category
## :rocket: Feature

As a user
I would like to be able to filter posts by their category(ies)
So that I can see only posts that show categories I'm interested in
### Design & Layout

Screenshot at 2019-07-03 17:15:31

### LINKED ISSUE #975: 🐛 [Bug] search bar in desktop view on normal size
## :bug: search bar in desktop view on normal size 
the search bar should only appear in the mobile view across the whole screen

Bildschirmfoto vom 2019-07-04 08-50-32

 
### Expected behavior
 
 
Bildschirmfoto vom 2019-07-04 08-22-19

mattwr18:
### Issues

- fixes #967 
- fixes #975


mattwr18:
2019-07-09-filter-posts-by-category

webapp/components/FilterPosts/FilterPosts.vue
```diff
@@ -0,0 +1,282 @@
...
+                  <ds-flex-item>
+                    <label>{{ category.name }}</label>
+                  </ds-flex-item>
+                  <ds-space margin-bottom="xx-large" />
+                </ds-flex>
+              </ds-flex>
+            </ds-flex>
+          </ds-flex-item>
+          <ds-flex-item :width="{ base: '50%', sm: '0%', md: '50%', lg: '10%' }">
+            <ds-flex v-for="category in chunk[7]" :key="category.id" class="categories-list">
```
mattwr18: Not happy with how many lines of code this component has just to break up the categories into 8 rows, but try as I might to iterate over the subarrays of the chunk array, I couldn't get it right... hopefully someone will be able to help me reduce this code

@roschaefer @Tirokk @ogerly

webapp/layouts/default.vue
```diff
@@ -3,13 +3,23 @@
...
                 <ds-logo />
-              </nuxt-link>
+              </a>
             </ds-flex-item>
-            <ds-flex-item>
+            <ds-flex-item
+              :width="{ base: '20%', sm: '20%', md: '20%', lg: '0%' }"
+              class="mobile-hamburger-menu"
+            >
+              <ds-button icon="bars" @click="toggleMobileMenuView" right />
```
mattwr18: hope you are okay with me changing your implementation @ogerly... as I added the filter posts dropdown to the navbar, it cluttered and broke the css on mobile, so I just added a hamburger menu as it was always kinda needed.
iPhone:
Screenshot at 2019-07-09 18:43:12
Screenshot at 2019-07-09 18:44:00
Screenshot at 2019-07-09 18:43:22
Ipad:
Screenshot at 2019-07-09 18:43:49
Screenshot at 2019-07-09 18:43:38

webapp/components/FilterPosts/FilterPosts.vue
```diff
@@ -0,0 +1,282 @@
...
+        const last = groupedCategories[groupedCategories.length - 1]
+        if (!last || last.length === 2) {
+          groupedCategories.push([this.categories[i]])
+        } else {
+          last.push(this.categories[i])
+        }
+      }
+      return groupedCategories
+    },
+  },
```
mattwr18: we could use `_.chunk` from lodash for this, if anyone thinks it's better/necessary

webapp/components/FilterPosts/FilterPosts.vue
```diff
@@ -0,0 +1,282 @@
...
+  methods: {
+    ...mapMutations({
+      setPosts: 'posts/SET_POSTS',
+    }),
```
mattwr18: setting `Posts` in `Vuex` allows us to be able to change the `Posts` in the newsfeed without needing to use a child component.

webapp/pages/index.vue
```diff
@@ -10,7 +10,7 @@
         />
       </ds-flex-item>
       <hc-post-card
-        v-for="(post, index) in uniq(Post)"
+        v-for="(post, index) in posts"
```
mattwr18: maybe I need to add this `uniq` back??

roschaefer:
When I click on the logo:

Screenshot - 2019-07-10T013306 302

roschaefer:
What's the meaning of this filter configuration, actually?

2019-07-10-013608_1920x1080_scrot

webapp/components/FilterPosts/FilterPosts.vue
roschaefer: You could wrap the chunks in a separate component and pass `chunk` as a prop and an integer as `key`. Would that help?

webapp/components/FilterPosts/FilterPosts.vue
roschaefer: There is plenty of `lodash` in `yarn.lock`. I wouldn't mind to add it to `package.json`.

webapp/components/FilterPosts/FilterPosts.vue
roschaefer: I don't quite get the reason for using Vuex here, could you explain that in more detail?

webapp/pages/index.vue
roschaefer: I dunno, what could add a duplicate post here?

mattwr18:
> What's the meaning of this filter configuration, actually?
> 
> 2019-07-10-013608_1920x1080_scrot

what do you mean by that? Have you clicked on all the categories? Maybe we could expect that if all the categories are in the filter, it would just send an empty `{}` and the `All`  button would be active, but I think that is a very rare use case, and probably not that important to worry about now. 

Do you disagree, or did I misunderstand your question?

webapp/components/FilterPosts/FilterPosts.vue
mattwr18: Well, I think I can, yes. The first thing I tried is to add the filter to the `FilterPosts.vue` and then update the apollo cache when the query returned successfully, but I had several issues.

One is that when I added `this.$apollo.query` and tried to update the stores cache it didn't update the `Posts` at all. I was able to get it to work with `this.$apollo.mutate`, even though it is actually a query and doesn't mutate anything, but it only worked if the user double-clicked on the category in the dropdown, which sent two network requests, but updated the `Posts` on the second.

I wanted to make use of the `Post` query with built-in filter that I think you added as part of the `Your filter bubble`, but since `FilterPosts` is a child of `default.vue`, and not a child of `webapp/pages/index.vue`, it's not possible to call `changeFilterBubble(filter)` without using global even listeners, which we've decided to refactor out and not use going forward.

I placed the filter posts dropdown in the `FilterMenu.vue`, and got it working fine, but in a pairing session with @alina-beck, she gave the feedback that it is probably better in the navbar, as filtering is a kind of searching, and it might not be clear what it is in the `FilterMenu`. I tend to agree with her.

Enter Vuex, I am quite familiar with using Redux as a way of managing state and passing props from related but not parent-child components, so I refactored the code to introduce the store where we can update the store after filtering for `Posts`, which in turn updates the `UI`.

webapp/components/FilterPosts/FilterPosts.vue
mattwr18: great

webapp/components/FilterPosts/FilterPosts.vue
mattwr18: yes, that would help, I believe

webapp/components/FilterPosts/FilterPosts.vue
```diff
@@ -0,0 +1,69 @@
...
+  },
+}
+</script>
+<style lang="scss">
+#filter-posts-header {
+  display: block;
+}
+.categories-list {
+  text-align: center;
+}
```
mattwr18: this styling should probably go in `FilterPostsMenuItems.vue`... I forgot, but can easily fix that

webapp/components/FilterPosts/FilterPostsMenuItems.vue
```diff
@@ -0,0 +1,221 @@
...
+      } else {
+        this.selectedCategoryIds.push(id)
+      }
+      if (!id) this.selectedCategoryIds = []
+      this.selectedCategoryIds.length ? (this.allCategories = false) : (this.allCategories = true)
+      this.$emit('filterPosts', this.selectedCategoryIds)
+    },
+  },
+}
+</script>
```
mattwr18: maybe this is an improvement @roschaefer @Tirokk @ulfgebhardt @ogerly 
I think it might actually add lines of code, but it extracts the menu items into their own component

webapp/layouts/default.vue
```diff
@@ -206,6 +245,21 @@ export default {
     foldedup: function() {
       document.getElementById('nav-search-box').classList.remove('unfolded')
     },
+    toggleMobileMenuView() {
+      this.toggleMobileMenu = !this.toggleMobileMenu
+    },
+    redirectToRoot() {
+      this.$router.replace('/')
+      this.fetchPosts({ i18n: this.$i18n, filter: {} })
+    },
```
mattwr18: this certainly feels like less than an ideal situation, and I'm open to other ideas how to clear the search queries for `Hashtags` and also refetch all the `Posts` to clear the search for `Hashtags/Categories`
@Tirokk @roschaefer @ogerly @ulfgebhardt 

it works with `location.reload()` as well, but I think that's worse.
it is not great user experience because it's not clear that anything is happening... maybe there could be a spinner/loader, but it's challenging because the logo link lives in `layouts/default.vue`, whereas the `Posts` list lives in `webapp/pages/index.vue`

webapp/store/posts.js
```diff
@@ -0,0 +1,77 @@
...
+          }
+      }`),
+      variables: {
+        filter,
+        first: 12,
+        offset: 0,
+      },
+    })
+    commit('SET_POSTS', Post)
+    return Post
```
mattwr18: would be nice to use `filterPosts` from `PostQuery`, but when I tried it, it's not returning the proper query, but a `Document` object... I think because you need to use `return filterPosts(i18n)`,but not sure how to do that here :thinking:

mattwr18:
new pics of the new and improved styling
Screenshot at 2019-07-12 10:43:36
Screenshot at 2019-07-12 10:45:22
Screenshot at 2019-07-12 10:45:09

there is still a performance issue when a person filters for posts by categories and then click on all to see all the posts... there is a very noticeable delay before the button is set to primary, only after the posts have been updated does the user get some feedback that their click was registered.

I have dropped in some starting, middle, and end times and `console.log`ged the results and it seems the request resolves quite quickly, and sets the posts in the store, but then there is a delay till the UI updates.... ideas, suggestions?? @roschaefer @Tirokk @ogerly @ulfgebhardt ??
Please play around with it and let me know what you think :pray:

mattwr18:
thanks @roschaefer... at which width does it look weird? I checked it out for all the width available on mobile responsive firefox, and it looks good, I think

yes, we talked about opening a separate issue for the performance issue, which @Tirokk remembers having a conversation with @appinteractive about(may or may not be related)

Issue #806, opened 2019-06-12
🐛 [Bug] HcContributionForm submit is not disabled by default


mattwr18:
## :bug: Bugreport

This means a user can try to submit the form without a title, for example.
It will fail the frontend validations, but it would still be better to make it disabled by default so that the button is disabled and only after meeting the validations would the button be disabled.

![ezgif com-video-to-gif (11)](https://user-images.githubusercontent.com/1324583/59379980-5be6c680-8d58-11e9-9a8c-892bac8430f8.gif)

Tirokk:
Solved by #977

PR #977, opened 2019-07-04
HcContributionForm submit is disabled by default

ogerly:
## 🍰 Pullrequest
HcContributionForm submit is disabled by default

ezgif com-video-to-gif (2)

### Issues
- fixes #https://github.com/Human-Connection/Human-Connection/issues/806


webapp/components/ContributionForm/ContributionForm.spec.js
```diff
@@ -201,7 +220,12 @@ describe('ContributionForm.vue', () => {
             content: 'auf Deutsch geschrieben',
             language: 'de',
             image,
-            categories: [{ id: 'cat12', name: 'Democracy & Politics' }],
+            categories: [
+              {
+                id: 'cat12',
+                name: 'Democracy & Politics',
+              },
+            ],
```
roschaefer: @ogerly try not to touch unrelated files. First, it makes revieweing your PR difficult. Second and more important, you will "own" the code. This will make it difficult to understand the code, because `git blame` will tell me that you are the author, but you just changed indentation.

To fix this: Run`git checkout origin/master -- path/to/unrelated/files`.

webapp/components/ContributionForm/ContributionForm.vue
```diff
@@ -169,6 +169,16 @@ export default {
     },
     updateEditorContent(value) {
       // this.form.content = value
+      this.disabled = true
+      if (
+        value.replace(
+          /<p>|<\/p>|<h3>|<\/h3>|<h4>|<\/h4>|<ul>|<\/ul>|<ol>|<\/ol>|<li>|<\/li>|<blockquote>|<\/blockquote>|<strong>|<\/strong>|<em>|<\/em>|<a.*>|<\/a>/gm,
```
roschaefer: Please don't use a regex. If you want to strip HTML from a string, the [internet has the answer](https://stackoverflow.com/questions/5002111/how-to-strip-html-tags-from-string-in-javascript).. Regexes are horrendously hard to maintain. "If you found out you can solve your problem with a regex, you have a new problem."

webapp/components/ContributionForm/ContributionForm.vue
```diff
@@ -169,6 +169,16 @@ export default {
...
+          '',
+        ).length > 3
```
roschaefer: Its' quite obscure what you do here. I guess you want to avoid validation errors in the backend? Do we even have a validation for the length of a `post.content`? @ogerly could you double-check?

Anyways, ideally, you would explain what you do here with a test case :+1:

webapp/components/ContributionForm/ContributionForm.vue
ogerly: to count the absolute characters, a lenght function is not enough. because the editor just inserts these html tags. so i only count the real characters that a user writes. nothing will be deleted or added from the original text. 

since I hold myself back from inserting npm packete. 

or I don't know if this already exists. 

since i love regular expressions that was my solution :)

ogerly: no the number of characters in a content is not limited . i would suggest 2000 characters. maybe 4000. you can say badly .. some like to write whole romanes. maybe there should be a BLOG corner for every user for something like that. 
i think 2000 characters is ok. those are 2 A4 pages

webapp/components/ContributionForm/ContributionForm.vue
ogerly: no the number of characters in a content is not limited . i would suggest 2000 characters. maybe 4000. you can say badly .. some like to write whole romanes. maybe there should be a BLOG corner for every user for something like that. 
i think 2000 characters is ok. those are 2 A4 pages

webapp/components/ContributionForm/ContributionForm.spec.js
ogerly: yes, i'm sorry. 
i have a lint problem in my editor. 
when i edit a page that i later restore to its original state. there is still a change in the git status due to the lint. 
should i just turn it off in my editor ? 
Bildschirmfoto vom 2019-07-10 06-51-51

webapp/components/ContributionForm/ContributionForm.vue
ogerly: no the number of characters in a content is not limited . i would suggest 2000 characters. maybe 4000. you can say badly .. some like to write whole romanes. maybe there should be a BLOG corner for every user for something like that. 
i think 2000 characters is ok. those are 2 A4 pages

webapp/components/ContributionForm/ContributionForm.vue
ogerly: replace("<[^>]*>", "")  is the short code ;)

ogerly: the short code is: .replace(/<\/?[^>]+(>|$)  ;)

ogerly:
maximum number of characters in content without html tags is 2000, characters are counted and number is displayed. regex is strongly shortened.

webapp/components/ContributionForm/ContributionForm.vue
ogerly: FireShot Capture 174 - Human Connection - Human Connection - localhost

roschaefer: But I think we're not talking about maximum but *minimum* characters, right? Or how do you explain what your code does?

webapp/components/ContributionForm/ContributionForm.vue
```diff
@@ -10,13 +10,15 @@
           />
         </hc-teaser-image>
         <ds-input model="title" class="post-title" placeholder="Title" name="title" autofocus />
+        <small style="width:100%;position:relative;left:90%">{{ form.title.length }}/64</small>
```
mattwr18: I prefer to avoid inline styling @ogerly, not sure how others fell about it

mattwr18:
@roschaefer, this is still waiting your approval before getting merged...

webapp/components/ContributionForm/ContributionForm.vue
Tirokk: Would be nice(!) to make a class out of it … @ogerly 
But not necessarily …
Inline styling has the positive side of better analysability, if the same style is not used elsewhere.
Classes (can) have descriptive names telling us what this is about.

webapp/components/ContributionForm/ContributionForm.vue
```diff
@@ -10,13 +10,15 @@
...
         <no-ssr>
           <hc-editor
             :users="users"
             :hashtags="hashtags"
             :value="form.content"
             @input="updateEditorContent"
           />
+          <small style="width:100%;position:relative;left:90%">{{ form.contentLength }}/2000</small>
```
Tirokk: I would like more to have this not a hard code max length and been a speaking and centralised changeable const …

Ah, you have defined it here: `this.formSchema.content.max`

webapp/components/ContributionForm/ContributionForm.vue
```diff
@@ -94,13 +97,14 @@ export default {
       },
       formSchema: {
         title: { required: true, min: 3, max: 64 },
-        content: { required: true, min: 3 },
+        content: { required: true, min: 3, max: 2000 },
```
Tirokk: Same same 👆🏼

webapp/components/ContributionForm/ContributionForm.vue
```diff
@@ -177,6 +181,14 @@ export default {
     updateEditorContent(value) {
       // this.form.content = value
       this.$refs.contributionForm.update('content', value)
+
+      this.disabled = true
+      this.n = value.replace(/<\/?[^>]+(>|$)/gm, '').length
+      this.form.contentLength = this.n
+
+      if (this.n > this.formSchema.content.min && this.n < this.formSchema.content.max) {
+        this.disabled = false
```
Tirokk: `this.n` should be a local `let` variable and not be global to this component …

webapp/layouts/default.vue
```diff
@@ -228,7 +228,7 @@ export default {
 
 .main-container {
   padding-top: 6rem;
-  padding-botton: 5rem;
+  padding-bottom: 5rem;
```
Tirokk: TOP !!! 👍🏼

webapp/components/ContributionForm/ContributionForm.vue
```diff
@@ -177,6 +181,14 @@ export default {
     updateEditorContent(value) {
       // this.form.content = value
       this.$refs.contributionForm.update('content', value)
+
+      this.disabled = true
+      this.n = value.replace(/<\/?[^>]+(>|$)/gm, '').length
+      this.form.contentLength = this.n
+
+      if (this.n > this.formSchema.content.min && this.n < this.formSchema.content.max) {
```
Tirokk: Here you use `this.formSchema.content.max` already! 👋🏼

webapp/components/ContributionForm/ContributionForm.vue
```diff
@@ -177,6 +181,14 @@ export default {
     updateEditorContent(value) {
       // this.form.content = value
```
Tirokk: Remove this comment?

webapp/components/ContributionForm/ContributionForm.vue
ogerly: amended

PR #993, opened 2019-07-08
User role not nullable

roschaefer:
## 🍰 Pullrequest
While implementing the signup I discovered this bug. Fortunately, we don't face it until we start to create user accounts.

Includes a refactoring, too.

### Issues

- None

### Todo

- [X] None

### DIFF
--- a/backend/src/models/User.js
+++ b/backend/src/models/User.js
@@ -11,7 +11,7 @@ module.exports = {
   coverImg: { type: 'string', allow: [null] },
   deleted: { type: 'boolean', default: false },
   disabled: { type: 'boolean', default: false },
-  role: 'string',
+  role: { type: 'string', required: true, default: 'user' },
   publicKey: 'string',
   privateKey: 'string',
   wasInvited: 'boolean',
--- a/backend/src/models/User.js
+++ b/backend/src/models/User.js
@@ -1,0 +1,54 @@
+import uuid from 'uuid/v4'
+
+module.exports = {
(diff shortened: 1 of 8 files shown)


backend/src/models/User.js
```diff
@@ -0,0 +1,54 @@
...
+  actorId: { type: 'string', allow: [null] },
+  name: { type: 'string', min: 3 },
+  email: { type: 'string', lowercase: true, email: true },
+  slug: 'string',
+  encryptedPassword: 'string',
+  avatar: { type: 'string', allow: [null] },
+  coverImg: { type: 'string', allow: [null] },
+  deleted: { type: 'boolean', default: false },
+  disabled: { type: 'boolean', default: false },
+  role: { type: 'string', default: 'user' },
```
mattwr18: so, basically you changed this to have a default of `user`
the other changes to this `PR` are just refactoring to extract out the models into their own files, correct?

backend/src/models/User.spec.js
```diff
@@ -0,0 +1,20 @@
...
+describe('role', () => {
+  it('defaults to `user`', async () => {
+    const user = await instance.create('User', { name: 'John' })
+    await expect(user.toJson()).resolves.toEqual(
+      expect.objectContaining({
+        role: 'user',
+      }),
+    )
+  })
+})
```
mattwr18: #moresoftwaretests
:smile_cat:

PR #1040, opened 2019-07-13
Remove graphql-yoga

roschaefer:
## 🍰 Pullrequest
After we decided to seed our database directly in our tests and not going over HTTP, this PR will remove `graphql-yoga` (insufficiently maintained, 3rd party) for `apollo-server`. This will also free us from the obsolete extra server during backend testing, `apollo-server-testing` works without starting an entire HTTP server. We even might get reasonable code coverage now @ulfgebhardt :wink: 

doge

### Issues

- None

### Todo

- [X] None


backend/src/middleware/handleHtmlContent/handleContentData.spec.js
```diff
@@ -68,7 +67,7 @@ describe('currentUser { notifications }', () => {
           'Hey <a class="mention" href="/profile/you/al-capone">@al-capone</a> how do you do?'
 
         beforeEach(async () => {
-          const createPostMutation = gql`
+          const createPostMutation = `
```
mattwr18: Can you explain why you removed this?
I think some really like the syntax highlighting this provides and while it might not provide any benefit for you using `vim`, I'm curious if you believe there is any benefit to removing it since some actually really like it.

backend/src/server.js
```diff
@@ -20,28 +20,30 @@ const driver = getDriver()
 
 const createServer = options => {
   const defaults = {
-    context: async ({ request }) => {
-      const user = await decode(driver, request.headers.authorization)
+    context: async ({ req }) => {
```
mattwr18: Can you explain this change?
I see cases where I would like to go through the code base and do the opposite where I spell it  out further, but I respect that others like to shorten variables, but if I lengthen and you shorten, we are kinda working against each other, or should we try to decide a standard and all stick by it?

backend/src/middleware/handleHtmlContent/handleContentData.spec.js
Tirokk: Yeah, I as an example … 😫😉

backend/src/server.js
Tirokk: Same, same … 😂

backend/src/middleware/handleHtmlContent/handleContentData.spec.js
roschaefer: I didn't want to remove it. But `gql` is not for syntax highlighting. It is parsing the string and returns an object. If you we use the parsed object in our tests, apollo complains about sending a parsed object instead of a string. Apparently `graphql-yoga` was indifferent about this.

roschaefer: By the way, vim also does the syntax highlighting. So I was benefiting from it, too. :wink:

backend/src/server.js
roschaefer: Again, not my decision, but `apollo`'s decision. I would spell it `request` and so does `graphql-yoga`, but `apollo` uses `req`.

roschaefer: Read [this](https://github.com/prisma/graphql-yoga/issues/449#issuecomment-440595783) for the **why**.

roschaefer:
@mattwr18 @Tirokk ES2015 template literals: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals

backend/src/jest/helpers.js
```diff
@@ -15,3 +15,9 @@ export async function login(variables) {
     authorization: `Bearer ${response.login}`,
   }
 }
+
+//* This is a fake ES2015 template string, just to benefit of syntax
+// highlighting of `gql` template strings in certain editors.
+export function gql(strings) {
+  return strings.join('')
+}
```
mattwr18: I like this solution, but how do you feel about it @roschaefer 
don't want you to feel like you are making too many concessions
I am also ok dealing with no syntax highlighting, not sure how @Tirokk feels.

backend/src/server.js
Tirokk: Thanks for letting us understand. 👍🏼

PR #1041, opened 2019-07-13
Fix #822 by updating the counts of the user

### LINKED ISSUE #822: 🐛 [Bug] load more is sometimes wrong on User Profile in the three Tabs
## :bug: Bugreport

At the Users Profile sometimes the `load more` is shown, even when there is nothing to load!

1. login as Jenny Rostock
2. go to Users Profile
3. delete a Post of yourself
4. switch between Tabs
5. suddenly the `load more` showed up at the Posts and Comments Tab

Bildschirmfoto 2019-06-13 um 18 28 46

Bildschirmfoto 2019-06-13 um 18 28 36

roschaefer:
The load more button depends on the total of the posts available.
Removing posts changes the total counts, so we have to manually update
it.

### Issues

- fixes #822 

### Todo

- [X] None

### DIFF
--- a/webapp/pages/profile/_id/_slug.vue
+++ b/webapp/pages/profile/_id/_slug.vue
@@ -202,11 +202,11 @@
 
           <template v-if="activePosts.length">
             <hc-post-card
-              v-for="(post, index) in activePosts"
+              v-for="post in activePosts"
               :key="post.id"
               :post="post"
               :width="{ base: '100%', md: '100%', xl: '50%' }"
-              @removePostFromList="activePosts.splice(index, 1)"
+              @removePostFromList="removePostFromList"
             />
           </template>
           <template v-else-if="$apollo.loading">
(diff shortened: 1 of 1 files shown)


webapp/pages/profile/_id/_slug.vue
```diff
@@ -331,6 +331,10 @@ export default {
     },
   },
   methods: {
+    removePostFromList(index) {
+      this.activePosts.splice(index, 1)
```
Tirokk: From where you get the `index` parameter, dear? @roschaefer 
After you deleted it from the `for loop` above …

roschaefer: The index parameter is still there! You pass the function `removePostFromList` and it has the signature of expecting one argument. When it's called it gets called with the required index param.

roschaefer: get called with the right arguments

PR #1139, opened 2019-07-30
Refactor social media backend

alina-beck:
## 🍰 Pullrequest
- uses `neode instance` for social media mutations
- uses `apollo test client` for social media backend tests


backend/src/middleware/permissionsMiddleware.js
```diff
@@ -7,7 +10,7 @@ import { rule, shield, deny, allow, and, or, not } from 'graphql-shield'
 const isAuthenticated = rule({
   cache: 'contextual',
 })(async (_parent, _args, ctx, _info) => {
-  return ctx.user !== null
+  return ctx.user != null
```
roschaefer: Why not
```suggestion
  return ctx && ctx.user && ctx.user.id
```
?

backend/src/middleware/permissionsMiddleware.js
```diff
@@ -30,6 +33,14 @@ const isMyOwn = rule({
   return context.user.id === parent.id
 })
 
+const isMySocialMedia = rule({
+  cache: 'no_cache',
+})(async (_, args, { user }) => {
+  let socialMedia = await instance.find('SocialMedia', args.id)
+  socialMedia = await socialMedia.toJson()
+  return socialMedia.ownedBy.node.id === user.id
```
roschaefer: Interesting! What is `socialMedia.ownedBy.node`? There are two different ids in Neo4J: It uses an internal `_id` which is an auto-incrementing number and we have our own custom `id` which is a random `uuid` in production environment.

How does `ownedBy.node` differ from `ownedBy.properties`?

backend/src/schema/resolvers/socialMedia.js
```diff
@@ -1,43 +1,32 @@
...
-          socialMediaUrl: params.url,
-        },
-      )
-      session.close()
+      const socialMedia = await instance.find('SocialMedia', params.id)
+      await socialMedia.update({ url: params.url })
+      const response = await socialMedia.toJson()
 
-      return params
+      return response
```
roschaefer: Returning the actual updated object is so much better!

so much better

backend/src/schema/resolvers/socialMedia.spec.js
```diff
@@ -1,186 +1,247 @@
...
     })
 
     describe('unauthenticated', () => {
       it('throws authorization error', async () => {
-        client = new GraphQLClient(host)
-        await expect(client.request(mutation, variables)).rejects.toThrow('Not Authorised')
+        const user = null
+        const result = await socialMediaAction(user, mutation, variables)
+
+        expect(result.errors[0]).toHaveProperty('message', 'Not Authorised!')
```
roschaefer: :heart_eyes: I learned where to look for the error message with `apollo-server-testing`.

alina-beck:
> ... what about those graphql requests:
> 
> ```graphql
> mutation {
>   CreateSocialMedia(...) {
>     url
>     ownedBy {
>       name
>     }
>   }
> }
> ```
> 
> I guess this will raise an error, because `SocialMedia.ownedBy` returns undefined. But we'll see...

Where does this mutation come from? All I know is:
```
mutation($url: String!) {
  CreateSocialMedia(url: $url) {
    id
    url
  }
}
```

backend/src/middleware/permissionsMiddleware.js
alina-beck: Sorry, I forgot to ask you about this! When we talked you were wondering how to test for `undefined` or `null` – I found out! :wink: But of course we can check for the user id if you think it makes more sense.

backend/src/middleware/permissionsMiddleware.js
alina-beck: Honestly, at that point I was just `console.log`ing the JSON object in search for the right `id` – so I'm not quite sure what `ownedBy.node` is and haven't even heard of `ownedBy.properties`... I will check again!

I couldn't use the internal Neo4J `_id` because the frontend is sending _our_ id in the args and that's the one I need to compare to.

backend/src/schema/resolvers/socialMedia.spec.js
alina-beck: :smile: Better yet are the nested error messages...

By the way, I read some more about apollo error handling yesterday and I think the reason a request resolves with errors is to allow partial failures – e.g. when you request a `User` but for some reason can't fetch his `SocialMedia` you will still get a response with all the other user data, plus an error telling you (hopefully) which part of the request went wrong and why.

alina-beck:
@roschaefer I know we briefly talked about using the `url` as unique identifier instead of an additional `id` – but then the same URL could not be added by two different users, right? And I think there might be cases where this is relevant (e.g. two friends sharing a youtube account). That's why I decided to keep it.

alina-beck:
@roschaefer your commits removed the possibility to `edit`/`Update` social media links from both the backend and frontend – was that on purpose?

roschaefer:
@alina-beck GraphQL solves the N+1 problem of REST. Let's say you query for your own user, all of your posts and then all of the posts' comments respectively. You might have to send `1+postCount+postCount*commentsCount` queries and you have to **wait for the requests over the wire**.
Graphql solves this by a query language where you can specify all related objects. E.g for the query:
```gql
{
  currentUser {
    name
    contributions {
      title
      comments {
        content
      }
    }
  }
}

```
You send one request and get all the data you need.

The downside is that your backend implementation becomes a little more complex because you may reckon that the client also requests the related objects.

Here is the specification where we declare the`ownedBy` on the type `User`:
```gql
type SocialMedia {
  id: ID!
  url: String
  ownedBy: User!
}
```

`ownedBy` returns a non-nullable `User` object. If we don't want to use `neo4j-graphql-js` we have to implement the `SocialMedia.ownedBy` resolver ourselves.

The reason why you don't see so many custom resolvers is that `neo4j-graphql-js` automagically provides all the data for you. However it's sometimes buggy and not very customizable. Therefore we want to write our own resolvers for at least the mutations.

alina-beck:
@roschaefer Ok, so if we implement this `SocialMedia.ownedBy` resolver does it work both ways? Because I reckon we would rather request a `User` with his owned `SocialMedias` than a `SocialMedia` with the `User` that owns it...

So, basically, with the current structure – can we do something like:
```
{
  currentUser {
    name
    ...
    socialMedias {
      url
    }
  }
}
```
and then get all the user's social media urls?

roschaefer:
@alina-beck the resolver of top level query `currentUser` is using `neo4jgraphql`, look [here](https://github.com/Human-Connection/Human-Connection/blob/refactor-social-media-backend/backend/src/schema/resolvers/user_management.js#L17). So if you request `currentUser.socialMedia` the library will make a database call and include the related `SocialMedia` and we don't have to do anything. However, if the backend responds with a type `SocialMedia` you can *always* ask for the related objects, too.

Theoretically you could do sth. silly like this:
```gql
{
  User {
    name
    socialMedia {
      url
      ownedBy {
        name
      }
    }
  }
}
```

By the way, this query results in an error on my machine.

alina-beck:
@roschaefer I just read your reasoning for removing the `update social media` feature. I agree that it makes sense from a database point of view if we use the `url` as the primary key.

From a user perspective I think I would prefer to have the possibility to edit links: In case I misspelled the link or when I change it it would feel better to `edit` than to `delete` and then `add` again.

I don't think it is a huge deal or that people will complain about not having this feature. But since it is already implemented it feels wrong to kick it out again – because I believe it would make life (a wee bit) easier for our users.

@roschaefer @mattwr18 @Tirokk  What do you think?

Also, if we use the `url` as primary key would it still be possible for two users to add the same social media link? This is probably not the most important use case but I do know of some people who share their social media accounts.

roschaefer:
@alina-beck you're correct, making `url` the primary key should also make it unique and *not editable* that's why I removed the update mutation. 

But I'm reconsidering it: What are the use-cases for having shared social media urls?

alina-beck:
@roschaefer I know of couples sharing instagram accounts and friends with shared youtube channels. I'm sure it would be frustrating if only one of the owners could add this account to their Human Connection profile.

Plus, it might be relevant for people who want to associate themselves with their employer (or sports team, etc.) – I've seen this a lot in the non-profit sector. I know these social media links are meant to be your personal accounts but if a proud volunteer wants to link to Human Connection, too, I say why not? :wink:

mattwr18:
Really great stuff here @alina-beck!! I haven't started using neode yet, but reviewing this definitely makes me want to give it a try!!
For what it's worth, I also agree with you and not blocking people from using the same social media account!
Your contributions are much appreciated!! :heart:

PR #1169, opened 2019-08-02
Move filter by followers functionality to filter dropdown

### LINKED ISSUE #964: 🚀 [Feature] Icon for all posts vs. posts of followers
## :rocket: Feature

~~~gherkin
Feature: As a user
I'd like to see a specific icon on my Dashboard
to show if I actually see all posts
or just the posts of people i follow

Scenario: An icon for all posts
Given that a User is logged in
When he switches to all posts
Then the corresponding icon is shown

Scenario: An icon for posts of users followed
Given that a User is logged in
When he switches to see just posts of users he follows
Then the corresponding icon is shown
~~~

### Design & Layout

The actual icon should be replaced by a reasonable one or a good text representation. The text "Your Filter Bubble" shold represent the actual state. E.g. "All Posts" vs. "Posts of People you follow".
feature-followers-icon

#### Design change
@appinteractive stated that there should be no white box behind this filter bubble thing, because it is no content.

### Additional context

It should be 100% clear, if the tooltip and the icon shows the actual state or the change of the state. The term "Filter Bubble" may not be widely known.

mattwr18:
### Issues

- fixes #964


mattwr18:
I have been unable to replicate this bug @roschaefer, are you on the latest code?
this duplicate key must not have something to do with the filterPosts by usersFollowed or byCategories because the duplicate key is a `post.id`, it seems.

I was able to see that if a user filters by categories and/or usersFollowed and then sort, all the posts are fetched instead of just the ones that a user had filtered for

PR #1171, opened 2019-08-02
Fix depreciation warnings

roschaefer:
I couldn't see any difference in using `transition`, `pageTransition` or
neither of those, so I decided to disable the configuration completely.

### Issues

- None

### Todo

- [X] None

### DIFF
--- a/webapp/nuxt.config.js
+++ b/webapp/nuxt.config.js
@@ -21,10 +21,10 @@ module.exports = {
 
   modern: !dev ? 'server' : false,
 
-  transition: {
-    name: 'slide-up',
-    mode: 'out-in',
-  },
+  // transition: {
+  //  name: 'slide-up',
+  //  mode: 'out-in',
+  // },
 
   env: {
     // pages which do NOT require a login
@@ -162,7 +162,9 @@ module.exports = {
   // Give apollo module options
   apollo: {
     tokenName: 'human-connection-token', // optional, default: apollo-token
-    tokenExpires: 3, // optional, default: 7 (days)
+    cookieAttributes: {
+      expires: 3, // optional, default: 7 (days)
+    },
     // includeNodeModules: true, // optional, default: false (this includes graphql-tag for node_modules folder)
 
     // Watch loading state for all queries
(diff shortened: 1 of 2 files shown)


webapp/nuxt.config.js
```diff
@@ -21,10 +21,10 @@ module.exports = {
...
   modern: !dev ? 'server' : false,
 
-  transition: {
-    name: 'slide-up',
-    mode: 'out-in',
-  },
+  // transition: {
+  //  name: 'slide-up',
+  //  mode: 'out-in',
+  // },
```
mattwr18: is there a reason we aren't either deleting this, or changing it to `pageTransition`?

webapp/nuxt.config.js
```diff
@@ -162,7 +162,9 @@ module.exports = {
   // Give apollo module options
   apollo: {
     tokenName: 'human-connection-token', // optional, default: apollo-token
-    tokenExpires: 3, // optional, default: 7 (days)
+    cookieAttributes: {
+      expires: 3, // optional, default: 7 (days)
+    },
```
mattwr18: is this related to a different deprecation warning that I'm not aware of? or how is this related?

roschaefer: Err, yes, if you `yarn run dev` you see a deprecation warning of `nuxt-apollo` about the `tokenExpires`. They just added more options and want to put it into an object

webapp/nuxt.config.js
roschaefer: My motivation to disable, not remove it: I could not see any difference in removing or adding it. However it is a documented feature. Maybe w'ere doing something wrong? Maybe I'm blind?

I can also remove it, if that helps you @mattwr18

webapp/nuxt.config.js
mattwr18: thanks for the explanation :)

webapp/nuxt.config.js
mattwr18: well, I don't think it would help me so much personally, hehe, the reason I ask is because it can be easy to collect comments in the code base and if we don't want/need some code, let's delete it, I think.

I pushed up some changes, if you can review. 

I had a look at the docs, and it seems like this config would be used anytime we have a `slide-up` transition, which we have a handful of, so I think it might be best to leave it, not remove it.
https://nuxtjs.org/api/pages-transition/
https://vuejs.org/v2/guide/transitions.html#Transition-Modes

PR #1174, opened 2019-08-03
Add static page for data privacy

ogerly:
EDIT:

```
    Implement a German-only static data-privacy page

    Okay, so the data privacy section needs to be updated @datenbrei:
    We have two sections mentioning the Vimeo and Youtube plugins. After we
    have the Embeds in place, we have many many more embedded code of 3rd
    parties. Can we simply remove the two sections?

    Also regarding the translations: I would wait until we have functional
    components and we can write translations enriched by markup. Until then
    I think it's sufficient to have German-only data privacy page.
```

### DIFF
--- a/webapp/pages/data-privacy.vue
+++ b/webapp/pages/data-privacy.vue
@@ -5,8 +5,379 @@
     </ds-space>
     <ds-container>
       <ds-space margin-top="large">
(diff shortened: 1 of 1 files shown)


roschaefer:
@datenbrei we created a static page for our data privacy. It's untranslated.
Can this be merged _for now_ :tm:  and we add translations later?

Issue #845, opened 2019-06-17
🚀 [Feature] Automatic loading of more Content when scroll hits load more


ulfgebhardt:
### Automatic loading of more Content when scroll hits load more

The Frontpage should load more content automatically once the `load more` option becomes visible.

The same applies to the profile page.

> wird das noch umgestellt auf automatisches Nachladen der Beiträge? Wäre aus meiner Sicht deutlich smarter, als wenn man ständig da drauf klicken muss. 

### Design & Layout

![image](https://user-images.githubusercontent.com/1238238/59601859-a0ca7e80-9105-11e9-94e2-97713b512c3f.png)

![image](https://user-images.githubusercontent.com/1238238/59601932-cf485980-9105-11e9-8e44-311a3179ae7a.png)

Tirokk:
Hey @MoldaB ,

thanks a lot for your wonderful and important contribution. 🚀💫🚀

There are still build errors with the Cypress tests and we need a moment to have time to get working on them …

Tirokk:
We would be very happy if you could apply this to the users profile page as well @MoldaB .

roschaefer:
closed by #1176

Issue #253, opened 2019-02-26
[Main Page] `↓ load more` button should disappear if there are no more posts to load


Tirokk:
**Describe the bug**
All posts are loaded. The `↓ load more` button underneath is still visible and active.

**To Reproduce**
1. Run the `master` branch. Go to 'http://127.0.0.1:3000/'
2. Scroll down to the last post
3. Click on `↓ load more` button
4. Do this 'til no more posts appear
5. `↓ load more` button is still visible and active

**Expected behavior**
`↓ load more` button disappears or gets inactive.

**Screenshots**
See screenshot:
<img width="1703" alt="bildschirmfoto 2019-02-26 um 17 01 48" src="https://user-images.githubusercontent.com/25344101/53426952-44bddc00-39e8-11e9-9010-c24d0f7cd259.png">

**Desktop (please complete the following information):**
 - OS: iOS
 - Browser safari
 - Version 12.0.2

**Related**
Related to issue #1176 .

PR #1176, opened 2019-08-04
845 load more

MoldaB:
### Issues
related #845 
related #253


MoldaB:
i closed the previous pull request because i was messy there. here is a more elegant solution :)

mattwr18:
hey, this is really great @MoldaB, did you see there are these warnings when running the component tests in `webapp`? `[Vue warn]: Failed to resolve directive: infinite-scroll`
were you able to run the tests locally? get the same output? maybe we can have a pairing session, if you'd like.

roschaefer:
sweet

I like the PR, too! I put in a commit to fix the `console.errors`. From the last build, I believe, the cypress tests are now stalling the build. @mattwr18 let's have a look, if the build server times out again.

ogerly:
i have merge the conflict :)

roschaefer:
@MoldaB thank you for your PR! I investigated why the build server crashes and it seems that the plugin endlessly tries to fetch more posts. This happens especially if there are few results, because infinite scroll will be triggered immediately. Even when `infinite-scroll-immediate-check` is set, it does not stop to call for more posts.

After looking into [vue-infinite-scroll](https://github.com/ElemeFE/vue-infinite-scroll) I was alarmed that the plugin has not seen any commits to the `master` branch [since September 2017](https://github.com/ElemeFE/vue-infinite-scroll/commits/master)! Also it has a long list of confusing, never addressed [issues](https://github.com/ElemeFE/vue-infinite-scroll/issues). It seemed I failed in checking the basics before suggesting you this plugin. Sorry for guiding you in the wrong direction!

So I tried to find a better maintained alternative and found [vue-infinite-loading](https://github.com/PeachScript/vue-infinite-loading). It is quite heavy and also comes with frontend code that we implement ourselves (e.g. spinner icons), but the main argument against it is the problematic API: It has a callback `infiniteHandler` which has a `state` argument and we're supposed to call `state.completed()` once we know there are no more posts. This is not quite what we need: We need to pass a property to the infinite loading component to tell it we're done. So I would not suggest to use it.

Maybe, if we're lucky, we could try to find a better maintained fork of `vue-infinite-scroll`, there are [338](https://github.com/ElemeFE/vue-infinite-scroll/network/members) of them at the moment!

Otherwise I would probably suggest to implement the infinite scroll manually. [It can't be *that* hard](https://www.google.com/search?q=vue+infinite+scrolling+tutorial&oq=vue+infinite+scrolling+tutorial). Maybe we even fork `vue-infinite-scroll` ourselves or copy+paste the useful parts of it?

roschaefer:
@Tirokk, @alina-beck and myself came to the conclusion: Let's add https://github.com/ElemeFE/vue-infinite-scroll and see if we run into any issues with it. So I'm going to merge this PR and say: Thank you @MoldaB for being awesome!

thank you

PR #1209, opened 2019-08-07
🍰 1054 blocked users

### LINKED ISSUE #1054: 🚀 [Feature] Block User
## :rocket: [Feature] Block User

~~~gherkin
Feature: As a user
I'd like to block another user
to hinder him to see and interact with my contributions, profile and comments
and also that I don't see his profile, contributions and comments any longer.
~~~

### Design & Layout

block-user
blocked-users

### Additional context

This is part of Feature Blacklist #279 and a working out of #222. It should also be possible to unblock a user again later the same way.

roschaefer:
### Issues

- fixes #1054 

### Todo

- [X] None


backend/src/schema/resolvers/users/blockedUsers.spec.js
```diff
@@ -0,0 +1,393 @@
...
+const instance = neode()
+
+let currentUser
+let blockedUser
+let authenticatedUser
+let server
+
+beforeEach(() => {
+  authenticatedUser = undefined
+  ;({ server } = createServer({
```
Tirokk: I never saw this notation before.
What does the `;` mean? @roschaefer

webapp/locales/de.json
```diff
@@ -189,6 +189,25 @@
...
+        "intro": "Wenn ein anderer Benutzer von dir blockiert wurde, dann passiert folgendes:",
+        "your-perspective": "In deiner Beitragsübersicht tauchen keine Beiträge der blockierten Person mehr auf.",
+        "their-perspective": "Umgekehrt das gleiche: Die blockierte Person sieht deine Beiträge auch nicht mehr in ihrer Übersicht.",
+        "search": "Die Beiträge von blockierten Personen verschwinden aus deinen Suchergebnissen.",
+        "notifications": "Von dir blockierte Personen erhalten keine Benachrichtigungen mehr, wenn sie in deinen Beiträgen erwähnt werden.",
+        "closing": "Das sollte fürs Erste genügen, damit blockierte Benutzer dich nicht mehr länger belästigen können."
+      },
+      "columns": {
+        "name": "Name",
+        "slug": "Alias"
```
Tirokk: Why here `Alias` and in English `Slug`? @roschaefer 

In the Admins Users list in both languages the slug is name `Slug`.
`Alias`is better understandable for the users for sure!

webapp/locales/en.json
```diff
@@ -189,6 +189,26 @@
...
+        "your-perspective": "The blocked person's posts will no longer appear in your news feed.",
+        "their-perspective": "Vice versa: The blocked person will also no longer see your posts in their news feed.",
+        "search": "Posts of blocked people disappear from your search results.",
+        "notifications": "Blocked users will no longer receive notifications if they are mentioned in your posts.",
+        "closing": "This should be sufficient for now so that blocked users can no longer bother you."
+
+      },
+      "columns": {
+        "name": "Name",
+        "slug": "Slug"
```
Tirokk: See above …

webapp/locales/de.json
roschaefer: I thought that "Slug" is not so common in German for most people to understand. I don't know what it actually translates to, I believe it's "Schnecke" and refers to the pattern of strings that have no witespaces as delimiter (in a URL).

Good luck understanding that one..

backend/src/schema/resolvers/users/blockedUsers.spec.js
roschaefer: It was added by eslint. The problem is that if you have e.g.
```js
const { a } = something
```
is valid and no problem but
```js
let a
{ a } = something
```
then babel gets confused. You need to wrap it with parantheses and apparently it's good practice (?) to add a semicolon as delimiter?

roschaefer:
@Tirokk forgot to add 31299235a3850e47ed885cf71579c3117da52dac could you recheck, please?

backend/src/schema/resolvers/users/blockedUsers.spec.js
Tirokk: Ah, okay.

webapp/locales/de.json
Tirokk: But does an English user understands `Slug` better as a German? @roschaefer

roschaefer: No, I think "Slug" is also a bad name. Shall we use "alias" everywhere?

Tirokk:
DE: Nun funzt es. Wir fliegen zu den Sternen …
EN: Now it’s functioning. We flying to the stars …

Cool thing @roschaefer !!! ❄️💫🚀🚀🚀

Wisdom

Tirokk:
Love to see what @mattwr18 says about the `Slug` …

webapp/locales/de.json
mattwr18: I understand what a slug is, but can't remember when I gained that understanding :woman_shrugging: 
An alias, I guess is better understood by the general user, but it's not really the same, is it?

Issue #372, opened 2019-04-03
🌟 [EPIC] User Dashboard / Home


ulfgebhardt:
### User Dashboard / Home

![image](https://user-images.githubusercontent.com/1238238/55469422-c7e4da00-5605-11e9-93dd-71c0e960a211.png)

Header:
- [X] GoToFeed
- [ ] Search https://github.com/Human-Connection/Human-Connection/issues/308
- [ ] Filter https://github.com/Human-Connection/Human-Connection/issues/374
- [X] Language
- [x] Notifications https://github.com/Human-Connection/Human-Connection/issues/345
- [X] Usermenu
- [ ] Sorting https://github.com/Human-Connection/Human-Connection/issues/375

Post Card:
- [x] Author Bug https://github.com/Human-Connection/Human-Connection/issues/376
- [X] CreatedAt
- [X] Title
- [X] Content Excerpt
- [X] Categories
- [X] Actions
- [x] Design differences https://github.com/Human-Connection/Human-Connection/issues/377
- [ ] Dynamic Card Height https://github.com/Human-Connection/Human-Connection/issues/378

More:
- [ ] New Post https://github.com/Human-Connection/Human-Connection/issues/379 + ~~Bug~~ https://github.com/Human-Connection/Human-Connection/issues/337
- [ ] Footer https://github.com/Human-Connection/Human-Connection/issues/381

### Userstory

```gherkin
Feature: News Feed
  As a user
  I want to see a list of contributions, possibly filtered and sorted,
  In order to get interesting news and ideas and get possibly involved in candos and projects.
```

### Design & Layout - Vision
![image](https://user-images.githubusercontent.com/1238238/55667309-ee4f8300-585a-11e9-945d-ecbdaf47ebca.png)

### Design & Layout - Alpha

![image](https://user-images.githubusercontent.com/1238238/55667323-25be2f80-585b-11e9-92f3-2a86cf685b1c.png)

### Design & Layout - Nitro

![image](https://user-images.githubusercontent.com/1238238/55472317-18f7cc80-560c-11e9-80cf-962c81383c1b.png)

PR #1256, opened 2019-08-13
Add masonry layout grid

### LINKED ISSUE #378: 🐛[Bug] All PostCards have the same height - dynamic height required
### All PostCards have the same height - dynamic height required

max height must be defined

image

image

1. Login
2. See Cards
3. All same Height
4. ...
5. Profit

### Expected behavior

image

### Version & Environment
 Type: [Desktop]      
 - OS: [Windows]      
 - Browser: [Chrome Beta] 
 - Version https://github.com/Human-Connection/Human-Connection/commit/e4cfa2403a457032c10b449abeaf3082e1863111  

### Additional Context

Keyword: masonry grid

alina-beck:
## 🍰 Pullrequest
Use CSS grid to create a masonry layout – allowing posts to have different heights.

So instead of this:
Screenshot 2019-08-13 at 16 42 47

We should now see this:
Screenshot 2019-08-13 at 16 49 48

### Issues
- fixes #378
- relates #372
- waiting for PR: https://github.com/Human-Connection/Nitro-Styleguide/pull/131

### ToDo
- [x] update styleguide dependency
- [x] use styleguide components in project


roschaefer:
@alina-beck I rebased your PR - you will need to `git fetch && git reset origin/add-masonry-grid --hard`

alina-beck:
@roschaefer this is still a work in progress – I need the [styleguide PR](https://github.com/Human-Connection/Nitro-Styleguide/pull/131) merged and released, then I can finish this one. I still have a few plans. :wink: But thank you for your suggestions! The diagram is really helpful. 😍 

May I ask why you are force-pushing and rebasing instead of merging in master? I've only ever done it locally and thought it wouldn't be a good idea to do it on the remote repo... Now I'm curious to hear your reasons because I've seen you do it quite often!

webapp/components/PostCard/index.vue
```diff
@@ -123,6 +132,26 @@ export default {
         this.$toast.error(err.message)
       }
     },
+    calculateItemHeight() {
```
roschaefer: @alina-beck do you still need this? The build server is crashing during frontend testing and it's trying to set some `style` attribute of `null`.

webapp/components/MasonryGrid/MasonryGridItem.vue
```diff
@@ -0,0 +1,38 @@
...
+<script>
+export default {
+  data() {
+    return {
+      rowSpan: 10,
+    }
+  },
+  methods: {
+    calculateItemHeight() {
+      this.$parent.$emit('calculating-item-height')
```
mattwr18: I was curious what the difference between `this.$emit` and `this.$parent.$emit` was, and found this SO post.
As far as I understand it, this is the way to tell the parent component to emit an event to its parent? so like emit an event to a grandparent component.
Is that correct @alina-beck?

webapp/components/MasonryGrid/MasonryGridItem.vue
```diff
@@ -0,0 +1,38 @@
...
+      })
+    },
+  },
+  mounted() {
+    const image = this.$el.querySelector('img')
+    if (image) {
+      image.onload = () => this.calculateItemHeight()
+    } else {
+      // use timeout to make sure layout is set up before calculation
+      setTimeout(() => this.calculateItemHeight(), 0)
```
mattwr18: I like that you added a comment to explain this, but if the timeout is `0`, I'm a little confused what this does

webapp/pages/index.vue
```diff
@@ -1,10 +1,10 @@
 <template>
   <div>
-    <ds-flex :width="{ base: '100%' }" gutter="base">
-      <ds-flex-item>
+    <masonry-grid>
+      <ds-grid-item v-show="hashtag" :row-span="2" column-span="fullWidth">
```
mattwr18: I was wondering where the `hashtag` was :smiley_cat:

webapp/components/MasonryGrid/MasonryGridItem.vue
alina-beck: Yes, and the reason we have to do it here is that this component sits within a slot – I found the solution in [this comment](https://github.com/vuejs/vue/issues/4332#issuecomment-263444492) and it's working! But hard to test... 😕

webapp/components/MasonryGrid/MasonryGridItem.vue
alina-beck: This has to do with the [JavaScript event loop](https://stackoverflow.com/questions/33955650/what-is-settimeout-doing-when-set-to-0-milliseconds/33955673). The timeout – even though it is set to `0` – sends the execution of this function to the back of the queue.

I am not sure which function _exactly_ we are waiting for because I don't know how Vue.js works its magic in the background but I know that without the timeout the layout is messed up. :smile:

alina-beck:
@mattwr18 thanks for the review!

I noticed the hashtag issue – and I believe it has nothing to do with this PR. The first thing I did after Rob rebased my commits was test the hashtag feature (because I needed to toggle the `FilterMenu` and wanted to know what was going on) and already then realised that I could click on a hashtag but wouldn't see all the relevant posts...

alina-beck:
@roschaefer Good catch! Thanks to you I found out that 

> CSS Grid Layout initially defined the `grid-row-gap` property. This prefixed property is being replaced by `row-gap`. (See [MDN](https://developer.mozilla.org/en-US/docs/Web/CSS/row-gap))

and am now using the new property (with the old one as a fallback). 😎

webapp/components/MasonryGrid/MasonryGridItem.vue
roschaefer: You could emit again in the parent component
(child)-emit->(parent)-emit->(granddaddy)

webapp/components/MasonryGrid/MasonryGrid.vue
```diff
@@ -0,0 +1,35 @@
+<template>
+  <ds-grid
+    :min-column-width="300"
+    v-on:calculating-item-height="startCalculation"
```
roschaefer:
```suggestion
    @calculating-item-height="startCalculation"
```

PR #1261, opened 2019-08-14
Localise categories

### LINKED ISSUE #1201: 🐛 [Bug] Categories are not localised
## :bug: Bugreport

We are currently storing the Categories as English words in the database, which means they are not localised anywhere is the front end.

A user should be able to see the categories in their preferred language.

1. Change the language from the `LocaleSwitch` dropdown to `DE`
2.  Click on the filterPosts dropdown in the navbar on the homepage
3. Notice the Categories are still in English 
4. ...
5. Profit

### Additional context

There are several possible solutions:
here are some articles about the subject
https://www.vertabelo.com/blog/technical-articles/data-modeling-for-multiple-languages-how-to-design-a-localization-ready-system
https://www.smashingmagazine.com/2014/06/css-driven-internationalization-in-javascript/

mattwr18:
### Issues

- fixes #1201


deployment/legacy-migration/maintenance-worker/migration/neo4j/users/users.cql
```diff
@@ -117,7 +117,7 @@ MERGE (e:EmailAddress {
   verifiedAt: toString(datetime())
 })
 MERGE (e)-[:BELONGS_TO]->(u)
-MERGE (u)<-[:PRIMARY_EMAIL]-(e)
+MERGE (u)-[:PRIMARY_EMAIL]->(e)
```
mattwr18: is this code creep?
I realised that this was incorrect during the last deploy to production, but never put in a `PR` to fix it... 
I can separate it into another `PR`, if y'all prefer

roschaefer: Yes, it is unrelated! However it's surprising to me that this was not found earlier.

PR #1266, opened 2019-08-15
🍰 registration button displayed when terms and conditions confirmed

ogerly:
## 🍰 Pullrequest
registration button displayed when terms and conditions confirmed

ezgif com-video-to-gif

### DIFF
--- a/webapp/components/Registration/CreateUserAccount.vue
+++ b/webapp/components/Registration/CreateUserAccount.vue
@@ -3,11 +3,9 @@
     <ds-card v-if="success" class="success">
       <ds-space>
         <sweetalert-icon icon="success" />
-        <ds-text
-          align="center"
-          bold
-          color="success"
-        >{{ $t('registration.create-user-account.success') }}</ds-text>
+        <ds-text align="center" bold color="success">
+          {{ $t('registration.create-user-account.success') }}
+        </ds-text>
       </ds-space>
     </ds-card>
     <ds-form
@@ -68,7 +66,9 @@
               :loading="$apollo.loading"
               :disabled="!checkedConfimed"
               primary
-            >{{ $t('actions.save') }}</ds-button>
+            >
+              {{ $t('actions.save') }}
+            </ds-button>
           </template>
         </ds-card>
       </template>
--- a/webapp/components/Registration/CreateUserAccount.vue
+++ b/webapp/components/Registration/CreateUserAccount.vue
@@ -3,9 +3,11 @@
     <ds-card v-if="success" class="success">
       <ds-space>
         <sweetalert-icon icon="success" />
-        <ds-text align="center" bold color="success">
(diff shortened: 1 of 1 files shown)


mattwr18:
hey @ogerly, I see that you requested  my review, but it still has the `[WIP]` tag and it has merge conflicts, can you merge in the latest from `master`, work through the merge conflicts and remove the `[WIP]` tag if it's ready?
or are you still working on this?

roschaefer:
Was this PR created by accident? I had a look into the registration form yesterday and I was satisfied with it. We could improve the validation error messages by implementing a "<ds-checkmark>` component but that's a change in the styleguide.

Can we close this?

ogerly:
@roschaefer  no please do not close!
That's an extension from yesterday. 
so that the button can be clicked again even if the password is changed. 
the check mark at AGB will be deleted if you set it and then enter your time password or change it. 
the hack will be deleted again. 
i have to deal more with the form shema. the checkbox should be connected to the formschema.

roschaefer:
@ogerly I cannot confirm this. On my laptop and on the current `master` branch, the registration form works just fine. If I change my password, the button gets disabled, if I uncheck the terms and conditions, the button get's disabled. If and only if both passwords match **and** the checkmark is checked only then the submit button is enabled. :+1: 

I made a comment about integrating the checkmark into the form schema but that's currently not possible because there are no checkmarks in the [Styleguide](https://styleguide.human-connection.org/data-input/dsform). @alina-beck wants to write us a checkmark component which we could use here and on the "Delete User Account" page. But that's for another day.

I think it's good for now, isn't it?

ogerly:
@roschaefer  das problem ist das der checkbox button nicht zum formshema gehört. 
das bedeutet das ich es extra prüfen muss. 
mit den variablen alleine hat es bei mir nicht funktioniert

roschaefer:
@ogerly das habe ich ja gesehen - ist ja nicht schlimm, wenn du das extra prüfen musst! Du kannst es nicht zum `formSchema` hinzufügen, weil es noch keine Checkmark-Komponente gibt im Styleguide. Das formSchema ist eine property, die du der `ds-form` Komponente gibst. Die `ds-form` Komponente kommt aus dem Styleguide und funkioniert mit den anderen Eingabe Komponenten `ds-input`, `ds-select` oder `ds-radio`. Es gibt aber eben noch nicht `ds-checkmark`. Ohne Checkmark Komponente wird das validieren nicht funktionieren.

Vielleicht schreibt uns die @alina-beck noch eine Checkmark Komponente. Bis dahin können wir aber die Registrierungsseite so lassen wie sie ist :+1:

Tirokk:
Page to go to in state logged out:
http://127.0.0.1:3000/registration/create-user-account

ogerly:
close

PR #1268, opened 2019-08-15
Validate that a post cannot be created without categories/too many categories

### LINKED ISSUE #1222: 💬 [Question] Does a Post have to have a Category?
## :speech_balloon: Question

A user can save a post even if there is no category chosen.
In the Alpha there had to be at least one category selected to save the post.

Bildschirmfoto 2019-08-08 um 16 20 46

mattwr18:
### Issues

- fixes #1222 

### Todo

- [X] None


backend/src/middleware/handleHtmlContent/handleContentData.spec.js
```diff
@@ -359,25 +354,17 @@ describe('Hashtags', () => {
...
           await mutate({
             mutation: updatePostMutation,
             variables: {
-              postId,
-              postTitle,
-              updatedPostContent,
+              id,
+              title,
+              content,
+              categoryIds,
```
mattwr18: @Tirokk, I changed the variable names for all these to reuse the mutation and variables and DRY it out a bit
I think it's clear that it's a post id, title, etc... since we are using the `CreatePost`/`UpdatePost` mutation/variables, but I could change this, if you think it's better.

backend/src/schema/resolvers/notifications.spec.js
```diff
@@ -1,26 +1,34 @@
...
+    icon: 'university',
+  })
 })
 
 afterEach(async () => {
   await factory.cleanDatabase()
 })
 
-describe('query for notification', () => {
+describe('Notification', () => {
```
mattwr18: this was an unintended change @Tirokk when dealing with a merge conflict, I can revert it, if you want

backend/src/models/Category.js
```diff
@@ -0,0 +1,21 @@
...
+    required: true,
+    default: () => new Date().toISOString(),
+  },
+  post: {
+    type: 'relationship',
+    relationship: 'CATEGORIZED',
+    target: 'Post',
+    direction: 'in',
+  },
+}
```
roschaefer: Love it :heart_eyes:

PR #1270, opened 2019-08-15
🍰 1062 notification about comment on post

### LINKED ISSUE #1062: 🚀 [Feature] Notification about Comment to the Users Post
## :rocket: Feature

I like to get a Notification in case an other User writes a Comment to my Post.

### Design & Layout

If I click on the Notification I will see this Comment by getting directed to my Post and scrolling to the right Comment.

### Additional context

One can use the functionality which is or will be implemented for mentioning a User in a Post, where he/she gets a Notification. See issue #1017.

Tirokk:
## 🍰 Pullrequest

Create Notifications if a User writes a Comment to your Post.

### Issues

- fixes #1062

### Todo

- [x] write more backend tests
- [x] write frontend tests


backend/src/middleware/handleNotifications/handleNotificationsMiddleware.js
```diff
@@ -0,0 +1,156 @@
...
+    (label === 'Post' && reason !== 'mentioned_in_post') ||
+    (label === 'Comment' && !['mentioned_in_comment', 'comment_on_post'].includes(reason))
+  ) {
+    throw new UserInputError('Notification fits not to reason!')
+  }
+
+  const session = context.driver.session()
+  const createdAt = new Date().toISOString()
+  let cypher
+  switch (reason) {
```
roschaefer: Now, this cypher statement is getting messy. Shall I jump in an try to refactor?

mattwr18: that would be great @roschaefer, but @Tirokk has changes to the branch to push up, could you wait some hours?

Tirokk: No the code is pushed up. @roschaefer 
I and @mattwr18 would like to have pairing with you about this.
May @ogerly and @alina-beck like to join it as well?

backend/src/middleware/handleNotifications/handleNotificationsMiddleware.js
```diff
@@ -0,0 +1,156 @@
+import { UserInputError } from 'apollo-server'
+import extractMentionedUsers from './notifications/extractMentionedUsers'
+import extractHashtags from './hashtags/extractHashtags'
+
+const notifyUsers = async (label, id, idsOfUsers, reason, context) => {
+  if (!idsOfUsers.length) return
+
+  // Done here, because Neode validation is not working.
```
mattwr18: I think the neode validations only work if you use neode
Like, if you use it in the resolver than it should work, but I think we are not, so it doesn't work.
If you create a test using neode and have defined that an attribute is required and don't use it, the validations should work.

backend/src/middleware/handleNotifications/handleNotificationsMiddleware.js
```diff
@@ -0,0 +1,156 @@
...
+  const reasonsAllowed = ['mentioned_in_post', 'mentioned_in_comment', 'comment_on_post']
+  if (!reasonsAllowed.includes(reason)) {
+    throw new UserInputError('Notification reason is not allowed!')
+  }
+  if (
+    (label === 'Post' && reason !== 'mentioned_in_post') ||
+    (label === 'Comment' && !['mentioned_in_comment', 'comment_on_post'].includes(reason))
+  ) {
+    throw new UserInputError('Notification fits not to reason!')
```
mattwr18: this would be better `Notification does not fit the reason` and this would not be a `UserInputError`, right?
for this to happen, it would not be the user doing something incorrectly, or is this for someone hitting the backend directly and trying to create a notification without the correct reasons?

backend/src/middleware/handleNotifications/handleNotificationsMiddleware.js
```diff
@@ -0,0 +1,156 @@
...
+    case 'comment_on_post': {
+      cypher = `
+        MATCH (postAuthor: User)-[:WROTE]->(post: Post)<-[:COMMENTS]-(comment: Comment { id: $id })<-[:WROTE]-(author: User)
+        MATCH (user: User)
+        WHERE user.id in $idsOfUsers
+        AND NOT (user)<-[:BLOCKED]-(author)
+        AND NOT (author)<-[:BLOCKED]-(user)
+        CREATE (notification: Notification {id: apoc.create.uuid(), read: false, reason: $reason, createdAt: $createdAt })
+        MERGE (comment)-[:NOTIFIED]->(notification)-[:NOTIFIED]->(user)
+      `
```
mattwr18: Are the `mentioned_in_comment` and `comment_on_post` cypher statements the same, or am I missing something?
the `$reason` would be different, right? but since it's passed in as a variable that would be ok to use the same cypher statement, no?

mattwr18: oh, this is different, no
```
AND NOT (user)<-[:BLOCKED]-(author)
AND NOT (author)<-[:BLOCKED]-(user)
```
right?

backend/src/middleware/handleNotifications/handleNotificationsMiddleware.js
Tirokk: I changed the comment to
`// Checked here, because it does not go through GraphQL checks at all in this file.`
@mattwr18 

Or

backend/src/middleware/handleNotifications/handleNotificationsMiddleware.js
Tirokk: Changed the translation.
Thanks for the a better one! 😍 @mattwr18 

Yap, you are absolutely right. It is just an `Error`. 👍🏼

backend/src/middleware/handleNotifications/handleNotificationsMiddleware.js
Tirokk: Yes, this is different. 😉

backend/src/middleware/handleNotifications/handleNotificationsMiddleware.js
```diff
@@ -0,0 +1,155 @@
...
+    `
+    const result = await session.run(cypherFindUser, {
+      commentId: comment.id,
+    })
+    session.close()
+    const [postAuthor] = await result.records.map(record => {
+      return record.get('user')
+    })
+    if (context.user.id !== postAuthor.id) {
+      await notifyUsers('Comment', comment.id, [postAuthor.id], 'comment_on_post', context)
```
roschaefer: Is it possible that you cannot mention another user when you write a post?

backend/src/middleware/handleNotifications/handleNotificationsMiddleware.js
```diff
@@ -0,0 +1,155 @@
...
+  if (comment) {
+    await notifyUsers('Comment', comment.id, idsOfUsers, 'mentioned_in_comment', context)
+  }
+
+  return comment
+}
+
+const handleCreateComment = async (resolve, root, args, context, resolveInfo) => {
+  // removes classes from the content
+  const comment = await handleContentDataOfComment(resolve, root, args, context, resolveInfo)
```
roschaefer: just saw that.. striked my comment below :point_down: 

I think the `handleHtmlContent` middleware could be separated into a `hashtagMiddleware` and `notficationMiddleware`. Those two are conceptionally completely different. It would make it a lot easier to understand what's going on here..

Tirokk: I have refactored it after your suggestion @roschaefer .
Thanks for your hint !!! 👍🏼

PR #1271, opened 2019-08-15
Fix user mention menu on production data

roschaefer:
Apparently some of our users in production have no slug, this will
assign an (hopefully unique) random slug to them.

### Issues

- None

### Todo

- [X] None

### DIFF
--- a/deployment/legacy-migration/maintenance-worker/migration/neo4j/users/users.cql
+++ b/deployment/legacy-migration/maintenance-worker/migration/neo4j/users/users.cql
@@ -94,12 +94,11 @@
       }
     }
 */
-
 CALL apoc.load.json("file:${IMPORT_CHUNK_PATH_CQL_FILE}") YIELD value as user
 MERGE(u:User {id: user._id["$oid"]})
 ON CREATE SET
 u.name       = user.name,
-u.slug       = user.slug,
+u.slug       = COALESCE(user.slug, apoc.text.random(20, "[A-Za-z]")),
 u.email      = user.email,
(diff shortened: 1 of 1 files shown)


deployment/legacy-migration/maintenance-worker/migration/neo4j/users/users.cql
```diff
@@ -94,12 +94,11 @@
       }
     }
 */
-
 CALL apoc.load.json("file:${IMPORT_CHUNK_PATH_CQL_FILE}") YIELD value as user
 MERGE(u:User {id: user._id["$oid"]})
 ON CREATE SET
 u.name       = user.name,
-u.slug       = user.slug,
+u.slug       = COALESCE(user.slug, apoc.text.random(20, "[A-Za-z]")),
```
mattwr18: from the docs `coalesce() returns the first non-null value in the given list of expressions.`
since `apoc.text.random()` will never return null, this is a safer way to import the data! 
Nice @roschaefer

mattwr18: one thing I didn't notice though @roschaefer is that you are using `A-Z`, shouldn't we avoid using slug with capital letters?
I already merged it, but we could refactor, or you could explain why it's fine...

roschaefer: @mattwr18 oh, yes you're right. We could also prefix the slug with `unknown-`... so these users show up at the end of the list.

PR #1277, opened 2019-08-15
Fix `Cannot read property countUser of null`

roschaefer:
This commit message is a great example of why you should explain
**the reason** of your commit.

When I came across this bug I had a quick look into the code where it
came from. I could see that the bug only happened after a timer and
apparently reverting the biggest part of d84892930295dcfd3f6687fc33c7234446127099
would fix the bug. However I have no idea what the following commit
message means:

```
commit d84892930295dcfd3f6687fc33c7234446127099
Author: Grzegorz Leoniec <greg@app-interactive.de>
Date:   Wed Mar 6 18:45:57 2019 +0100

    Improved countTo component
```

I just don't know why the code is there - like what is it's purpose
@appinteractive? I can only guess: I believe that it's supposed to
update the counters in-place (without starting from 0 everytime).
Because apollo was set to poll the data every second.

Taking that into account I would rather remove this polling feature
completely and have less code and less complexity. Admins can still
refresh the page.

### Issues

- None

### Todo

- [X] None

### DIFF
--- a/webapp/components/CountTo.vue
+++ b/webapp/components/CountTo.vue
@@ -2,7 +2,7 @@
   <span>
     <no-ssr placeholder="0" tag="span">
       <count-to
-        :start-val="lastEndVal || startVal"
+        :start-val="startVal"
         :end-val="endVal"
         :duration="duration"
         :autoplay="autoplay"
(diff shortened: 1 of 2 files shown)


roschaefer:
@ogerly we came across this bug in a pair-programming

Issue #1276, opened 2019-08-15
🐛 [Bug] Hashtag links are not URL safe


roschaefer:
## :bug: Bugreport
Some of the imported hashtags contain whitespaces. If you click on one of them, you get to page `/search/hashtag/#%20Appeninnen` and redirected to the index page, thus not filtering for any hashtags.

### Steps to reproduce the behavior
1. Write a post with a hashtag, choose an existing one with a whitespace
2. Save the post
3. Click on the hashtag
4. Get redirected to the index page with the list of all posts unfiltered

![2019-08-15-223129_1920x1080_scrot](https://user-images.githubusercontent.com/2110676/63124735-42f2cf80-bfac-11e9-8b84-238d29e7ca47.png)

### Expected behavior
You should see a list of posts that all have been tagged with the hashtag.

### Additional context

There are two things to be fixed here:
1. The database import should make use of [apoc.text.snakeCase()](https://neo4j-contrib.github.io/neo4j-apoc-procedures/3.5/utilities/text-functions/#_convert_a_string_to_snake_case_with_literal_snakecase_literal) (this will be solved here #1399 )
2. The unnecessary redirect should be removed. Instead, the link to the hashtag search should have *safely encode* the filter object as url params. This is an example of the string `# Appeninen` safely encoded: `http://localhost:3000/?hashtag=#%20Appenninen`

mattwr18:
I disagree with this @roschaefer `The database import should make use of apoc.text.snakeCase() (this will be solved here #1399 )`, we did not attempt to solve this in that `PR`... we don't want to transform people's tags to the case we think they should use.

PR #1278, opened 2019-08-15
Fix hashtag db import

roschaefer:
## 🍰 Pullrequest
This would fix the database import part of #1276 

### Issues

- None

### Todo

- [X] None


deployment/legacy-migration/maintenance-worker/migration/mongo/export.sh
```diff
@@ -38,14 +38,12 @@ export_collection "emotions"
...
 export_collection_query "follows" "{'foreignService': 'users'}" "users"
 export_collection "invites"
-export_collection "notifications"
 export_collection "organizations"
 export_collection "pages"
 export_collection "projects"
 export_collection "settings"
 export_collection "shouts"
 export_collection "status"
-export_collection "systemnotifications"
```
mattwr18: oh no, code creep!!

deployment/legacy-migration/maintenance-worker/migration/neo4j/users/delete.cql
```diff
@@ -1 +1,2 @@
-MATCH (n:User) DETACH DELETE n;
\ No newline at end of file
+MATCH (n:User) DETACH DELETE n;
+MATCH (e:EmailAddress) DETACH DELETE e;
```
mattwr18: code creep?

roschaefer: No, you get a unique constrain violation error if you run the database import multiple times. @ulfgebhardt deletes each label separately instead of removing all at once (including in this case `EmailAddress`)

PR #1280, opened 2019-08-16
Refactor graphql queries

mattwr18:
- Remove duplicate queries
- Use smart query in pages/post/_id/_slug/index.vue to avoid multiple db requests for a post and its comments.
We cannot update the apollo cache with asyncData and smart queries have a prefetch policy set to true by default, which
means that they will resolve in a similar timeframe. https://stackoverflow.com/questions/55885337/in-nuxt-should-i-use-asyncdata-or-default-apollo-queries
https://vue-apollo.netlify.com/api/smart-query.html#options
https://vue-apollo.netlify.com/guide/ssr.html#vue-cli-plugin

### Issues

- None

### Todo

- [X] None


webapp/components/comments/CommentList/index.vue
```diff
@@ -41,29 +40,5 @@ export default {
...
-        return PostCommentsQuery(this.$i18n)
-      },
-      variables() {
-        return {
-          slug: this.post.slug,
-        }
-      },
-      fetchPolicy: 'cache-and-network',
-    },
-  },
```
roschaefer: Yesss, less code, less state, less bugs!

***State is the root of all evil***
root of all evil

webapp/pages/post/_id/_slug/index.vue
```diff
@@ -122,29 +122,6 @@ export default {
...
-    const client = apolloProvider.defaultClient
-    const query = PostQuery($i18n)
-    const variables = { slug: params.slug }
-    const {
-      data: { Post },
-    } = await client.query({ query, variables })
-    if (Post.length <= 0) {
-      // TODO: custom 404 error page with translations
-      const message = 'This post could not be found'
-      return error({ statusCode: 404, message })
```
roschaefer: Removing `asyncData` will also remove the feature of sending a 404 if there is no post!

mattwr18: I don't think that is true
Screenshot at 2019-08-16 13:04:24

PR #1281, opened 2019-08-16
272 add error reporting

### LINKED ISSUE #272: Add error tracking
Add error tracking like sentry to know if there are errors that we need to tackle.

roschaefer:
### Issues

- fix #272

### Todo

- [ ] Deploy a self-hosted instance of Sentry
- [ ] Connect Sentry with Github's releases


roschaefer:
@mattwr18 I used a hosted instance. For data privacy, we shouldn't do that on `production`. I think for https://nitro-staging.human-connection.org/ we don't need to be so careful. I'll add the DSNs once this is deployed.

roschaefer:
@mattwr18 I asked @datenbrei to give us a self-hosted sentry instance.

PR #1283, opened 2019-08-18
fix(header-alignment): Fix logo and locale position.

### LINKED ISSUE #1246: 🐛 [Bug] Login design lack, have the Header items left and right aligned
## :bug: Bugreport

On the login page the items in the Header should be left and right aligned with the login box.

1. go to the login page (may logout)

Bildschirmfoto 2019-08-09 um 17 49 34

gha-gozu:
## 🍰 Pullrequest
Change the alignment of the header items on Login page. Header items are now aligned on top of the login box.
image

### Issues
- fixes #1246

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -48,7 +48,7 @@ services:
     networks:
       - hc-network
     ports:
-      - 80:80
+      - 8001:8001
 
 networks:
   hc-network:
--- a/webapp/layouts/default.vue
+++ b/webapp/layouts/default.vue
@@ -4,7 +4,7 @@
       <ds-container class="main-navigation-container" style="padding: 10px 10px;">
         <div>
           <ds-flex class="main-navigation-flex">
-            <ds-flex-item :width="{ lg: '3.5%' }" />
+            <ds-flex-item :width="{ lg: '30%' }" />
             <ds-flex-item :width="{ base: '80%', sm: '80%', md: '80%', lg: '15%' }">
               <a @click="redirectToRoot">
                 <ds-logo />
@@ -17,7 +17,7 @@
               <ds-button icon="bars" @click="toggleMobileMenuView" right />
             </ds-flex-item>
             <ds-flex-item
-              :width="{ base: '85%', sm: '85%', md: '50%', lg: '50%' }"
+              :width="{ base: '85%', sm: '85%', md: '50%', lg: '20%' }"
               :class="{ 'hide-mobile-menu': !toggleMobileMenu }"
             >
               <div id="nav-search-box" v-if="isLoggedIn">


docker-compose.yml
```diff
@@ -48,7 +48,7 @@ services:
     networks:
       - hc-network
     ports:
-      - 80:80
+      - 8001:8001
```
roschaefer:
```suggestion
      - 8001:80
```
Good catch!

I think the maintenance screen is just a nginx running on port 80 displaying static html. So you would have to forward port 8001 on host to port 80 in the container.

webapp/layouts/default.vue
```diff
@@ -4,7 +4,7 @@
       <ds-container class="main-navigation-container" style="padding: 10px 10px;">
         <div>
           <ds-flex class="main-navigation-flex">
-            <ds-flex-item :width="{ lg: '3.5%' }" />
+            <ds-flex-item :width="{ lg: '30%' }" />
```
roschaefer: This looks strange on my machine:
Screenshot - 2019-08-19T092707 051

docker-compose.yml
gha-gozu: Yeah, I was wondering why I can't run anything before.

webapp/layouts/default.vue
gha-gozu: I forgot to check the view on other devices. Will change it later today.

gha-gozu:
Updated the PR message

roschaefer:
@gha-gozu thank you for updating the PR description! Now I understand what you are trying to improve. It's certainly an improvement that the login page looks better aligned but unfortunately a change to the `default` layout affects all other pages, too!

Have you seen the other pages locally? You can [seed your database](https://docs.human-connection.org/human-connection/backend#seed-database) to get the same data as on https://nitro-staging.human-connection.org/.

For the other pages, the Human Connection logo looks misaligned. If we really wanted, we could probably use a dedicated *layout* for the login page. See [NuxtJS docs](https://nuxtjs.org/guide/views#layouts) to learn more about the layouts.

Do you need help on this? Feel free to ping @mattwr18 @alina-beck @Tirokk or me!

roschaefer:
@gha-gozu I had a look at #1246 and I think: Why not doing it the opposite way? If you can't change the top bar so easily, maybe you can change the login window?

It might look terrible, but we could give it a try, no?

gha-gozu:
@roschaefer Yesterday I tried to look more into the code and I realized, as you said that it is a default layout.
Maybe we could use something like conditional rendering, no? I mean the login window is already good as it is.

alina-beck:
I haven't looked into this in detail – but my idea would be to put the header in a separate component and then pass it a `content-width` prop that can be used in the `ds-container`.

This would mean we have to check if we are on the `login` route in the default layout... not ideal but maybe doable? What do you think @gha-gozu? Or @roschaefer do you know of a way to pass props to layouts?

gha-gozu:
As there is `isLoggedIn` flag, how about using it to use ternary to pass the props value? @alina-beck @roschaefer

roschaefer:
It is not easy at all to pass props to the layout. https://github.com/nuxt/nuxt.js/issues/647
If you have the store involved then you have state and as you know:
> State is the root of all evil

So I would rather not change the layout at all than adding logic/behaviour to it for the sake of the width of the top bar. We have some behaviour here:
https://github.com/Human-Connection/Human-Connection/blob/master/webapp/layouts/default.vue#L225
and it's already quite buggy sometimes.

No one likes my idea of changing the width of the login screen? :crying_cat_face: 

Ok, @gha-gozu if you want to change the `default` layout conditionally: Create a computed property in the layout and call it a telling name (difficult!). Return the `firstRoute` similar to `showFilterPostsDropdown`. You could check if that route === `login` and there you have your condition for the width of the layout.

appinteractive:
Why not just remove the topbar on the login page by using a blank layout? This was the way it was intended in the first place as the logo and language menu are already inside the login card.

roschaefer:
@gha-gozu I like @appinteractive suggestion the most. Just refactor the footer here:
https://github.com/Human-Connection/Human-Connection/blob/master/webapp/layouts/default.vue#L129
into it's own component. Use this component on the `blank` and the `default` layout. Easy!

mattwr18:
ahhh, yeah, duh! I just noticed that you can change the language from the login card hahaha
I really like @appinteractive suggestion!!

roschaefer:
My brother just said he would prefer to have a wider login window.

appinteractive:
One rule in UI/UX is to not make forms to wide. Just as a hint.

appinteractive:
I can also recommend some posts from Adam: https://refactoringui.com/

mattwr18:
the first thing we did today in our pairing session was to use the blank layout to remove the navbar as we discussed :point_up: 
then we realized the `passwor-reset.vue` page had kind of the same issue
and actually there are now 8 pages where a user is not logged in, so only needs the logo to return to the login page and the locale switch to translate the text.
we didn't think it made much sense to have the logo and locale switch dropdown at different widths based on where the user was, and since having it too narrow looked poorly on the pages with more content, like the terms and conditions, imprint, etc... left the navbar with the logo and locale switch in more or less the same position as when a user is logged in, and in the same place as all the other pages when a user is logged out.
Screenshot from 2019-09-04 00-13-16


this therefore does not fix the originally issue @Tirokk you opened, but maybe there are other considerations to discuss, as stated :point_up:

appinteractive:
Hey guys, don’t reinvent the wheel, the questions that you have are already solved in the alpha, or what is the exact issue here? 

Feels like that topic grows out of proportions. 🧐

mattwr18:
haha I agree! I'm not sure whether it was an issue or not, but there was an issue opened because the navbar header and locale dropdown didn't line up with the login card.
On the alpha, there is no navbar, neither on the log in card, nor on the reset password card, but then there is a link to translate, and a link to go back to the login page.
Also, on the alpha the footer links don't go to separate pages, rather they are popups, so the user can change the language in the log in menu, then when they click on the terms and conditions link, a modal opens translated to their chosen language.

For nitro, we have all of these footer links linked to other pages with a navbar with the logo to get back to login, and a way to translate the text to the chosen language.

I guess the question is do we want to have no navbar on the login and reset password pages and a basic header on the other 6 pages, or leave it like it is in this `PR` where all 8 pages have a basic header and not worry that it's lined up with the cards on the login/reset password pages. I guess a third option could be to have different alignment for the navbar links based on what goes best with the page the user is on.

webapp/components/PageFooter/PageFooter.vue
```diff
@@ -0,0 +1,15 @@
+<template>
+  <div id="footer" class="ds-footer">
+    <a href="https://human-connection.org" target="_blank" v-html="$t('site.made')"></a>
+    &nbsp;-&nbsp;
```
Tirokk: This `-` (\&ndash;) is an other than … 👇🏼

webapp/components/PageFooter/PageFooter.vue
```diff
@@ -0,0 +1,15 @@
...
+    <nuxt-link to="/imprint">{{ $t('site.imprint') }}</nuxt-link>
+    &nbsp;‑&nbsp;
```
Tirokk: … this `‑` (dash) and see the others below …

webapp/pages/login.vue
```diff
@@ -10,9 +10,6 @@
       <ds-card class="login-card">
         <ds-flex gutter="small">
           <ds-flex-item :width="{ base: '100%', sm: '50%' }" centered>
-            <client-only>
-              <locale-switch class="login-locale-switch" offset="5" />
-            </client-only>
```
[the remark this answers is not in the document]
vardyb: @Tirokk Thanks for the feedback - are you suggesting that we remove the header from all the pages that are now using the new basic layout?  If so we would need to consider where we place the language selector on some of the other pages such as the Imprint, Terms and Conditions, Data Privacy... pages?  Or should we keep the basic layout for these pages and create a new blank layout for the Login and possibly password reset pages?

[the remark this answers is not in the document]
roschaefer: I think the idea in https://github.com/Human-Connection/Human-Connection/pull/1283#issuecomment-523478461 was to remove the top bar *only* from the login page, but keep the top bar on the pages linked in the footer (code of conduct etc). Maybe you could use the `blank` layout and add the footer manually?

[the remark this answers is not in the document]
roschaefer: Oh and http://localhost:3000/registration/create-user-account and http://localhost:3000/password-reset/request would have the same layout as login. The language chooser is missing there in the component, I guess (you could add that probably :wink: )

vardyb:
@roschaefer  I'm not sure why you are seeing the top banner on the Login and Password pages - I made changes with @mattwr18  & @aonomike earlier this morning to remove the header on these two pages?

vardyb:
Login Page

Reset Page

T&C Page

@roschaefer it should look like the above now? ☝️

webapp/pages/login.vue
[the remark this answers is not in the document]
vardyb: @roschaefer & @Tirokk I've now updated the Login, Password Reset and Registration pages so that the header has been removed.  The other pages such as Imprint, terms and conditions etc continue to use the `basic` layout.  On my development machine the logo is in line with the content on these pages.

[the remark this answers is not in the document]
vardyb: Registration Page

This is how the registration pages displays locally for me now with the latest changes 👆

vardyb:
@roschaefer Thanks for the feedback. I was sure I had passing local tests but guess not. I will take a look at the existing tests as per your suggestion and update as appropriate.  It will probably be tomorrow (Thursday) now before I get the chance to take a look.

mattwr18:
hey @vardyb, I already fixed the tests... I understood we were gonna work on something else tomorrow.
sorry if I misunderstood

Tirokk:
Sorry, I have to approve this tomorrow.
Time lack …

mattwr18:
@appinteractive, @vardyb is not available today, and this `PR` is getting a bit out of hand, in my opinion
we have 40 conversations and making multiple changes that have nothing to do with the original ticket.
Is there a reason you want to have it right aligned? the login page is not right aligned.
this is what it looks like
Screenshot from 2019-09-11 16-35-43

mattwr18:
I think you mean left aligned, which would match the login page???
Screenshot from 2019-09-11 16-38-10

PR #1295, opened 2019-08-19
Configure dbms.security.procedures

roschaefer:
Bug:
> apoc.cluster.graph is unavailable because it is sandboxed and has dependencies
> outside of the sandbox. Sandboxing is controlled by the
> dbms.security.procedures.unrestricted setting

See: https://stackoverflow.com/a/48773575

### Issues

- None

### Todo

- [X] None


roschaefer:
It's already deployed by the way..

PR #1313, opened 2019-08-20
[WIP] Does a .first() break the cypress test silently?

### LINKED ISSUE #1312: 🐛 [Bug] One cypress test fails but it does not fail the build
I have no idea how to reproduce this error. As a matter of fact [this build](https://travis-ci.com/Human-Connection/Human-Connection/builds/123931292) is green when it should be red:

Screenshot - 2019-08-20T232340 509

However:
Screenshot - 2019-08-20T232412 906

### Expected behavior
The build should fail.

### Version & Environment
Travis CI! :fearful:

roschaefer:
I don't know but I'll push that commit to Github to trigger Travis CI. I
hope it will fail now.

### Issues

- close #1312 

### Todo

- [X] None

### DIFF
--- a/cypress/integration/common/steps.js
+++ b/cypress/integration/common/steps.js
@@ -354,6 +354,7 @@ When("mention {string} in the text", mention => {
 });
 
 Then("the notification gets marked as read", () => {
+  cy.get(".notifications-menu-popover").contains('.notification')
   cy.get(".notification")
     .first()
     .should("have.class", "read");


mattwr18:
@roschaefer I fixed this in https://github.com/Human-Connection/Human-Connection/pull/1268/files
see what you think about it

roschaefer:
@mattwr18 it's great but isn't it scary that the cypress test *does not fail the build*? :fearful: that's what I call a false negative.

mattwr18:
totally agree, any idea why that is, in the end?

mattwr18:
I can open an issue on cypress about this @roschaefer, I'm not sure if that is the right place for it or not, but travis is no longer accepting issues on their github repo, only on the community forum
Do you think it would be a good idea to sign up and create a forum post about it there, or just create an issue on cypress, both?

roschaefer:
@mattwr18 I suspect a bug in cypress or maybe in cypress-cucumber-preprocessor. I don't believe it has to do with Travis CI because it's *always this particular* cypress test.

mattwr18:
closing as this step was rewritten and is now passing

PR #1334, opened 2019-08-21
Check if user has agreed to the current terms and conditions

### LINKED ISSUE #1081: 🚀 [Feature] Confirmation of the terms  DB
## :rocket: Feature
the first confirmation of the terms and conditions comes with the registration of a user. 

- if the general terms and conditions change then there must be a date for it in the DB. 

- if the date of the current general terms and conditions from the DB is more current than the last confirmation of the user, a confirmation window must open for the registered user in which he must confirm the new changed general terms and conditions. 

this means:
- the user must have an additional data record (DATETIME) in the db. 
the date of the last confirmation of the general terms and conditions

## last_TaC_confirmation [DATETIME]

with the registration of a user this date will be set for the first time

ogerly:
## 🍰 Pullrequest
All data necessary for server-side rendering are taken over.
when a user registers he automatically agrees to the terms of use. 
If the version of the terms of use changes then every user will be asked after logging in to agree to the current terms of use. 

# !!
If you're going to branch out, you're going to have to
1. reset your DB
2. reseed your DB
because new user data will be added
 
FireShot Capture 030 - New Terms and Conditions - Human Connection - localhost

FireShot Capture 029 - Neue Nutzungsbedingungen - Human Connection - localhost

### Issues
- fixes #1081


appinteractive:
Please NEVER overlay complimentary colors as it not only hurts eyes but also destroys accessibility. Also be careful with colored backgrounds with white text on it for the same reasons.

backend/src/schema/types/type/EmailAddress.gql
```diff
@@ -19,5 +18,6 @@ type Mutation {
     avatarUpload: Upload
     locationName: String
     about: String
+    termsAndConditionsAgreedVersion: String!
```
roschaefer:
```suggestion
    termsAndConditionsAgreedVersion: String
```
@ogerly I think you can allow this to be `null`. If we import the users, they will have `termsAndConditionsAgreedVersion == null`. So that's a valid situation where this is null. The check in the nuxtjs middleware should still work as expected.

webapp/components/Registration/CreateUserAccount.vue
```diff
@@ -84,9 +87,24 @@ import gql from 'graphql-tag'
...
+/* TODO: hier muss die version rein */
 export const SignupVerificationMutation = gql`
-  mutation($nonce: String!, $name: String!, $email: String!, $password: String!) {
-    SignupVerification(nonce: $nonce, email: $email, name: $name, password: $password) {
+  mutation(
+    $nonce: String!
+    $name: String!
+    $email: String!
+    $password: String!
+    $termsAndConditionsAgreedVersion: String!
```
Tirokk: Has the exclamation mark to be removed here as well?!
… after @roschaefer suggestion?

Tirokk:
For testing visit login page …

And visit `create-user-account`:
http://127.0.0.1:3000/registration/create-user-account

backend/src/schema/resolvers/users/termsAndConditions.spec.js
```diff
@@ -0,0 +1,24 @@
+describe('SignupVerification', () => {
+  describe('given a valid version', () => {
+    // const version = '1.2.3'
+
+    it.todo('saves the version with the new created user account')
+    it.todo('saves the current datetime in `termsAndConditionsAgreedAt`')
```
roschaefer: Please implement those test cases! :pray:

webapp/components/Registration/CreateUserAccount.vue
roschaefer: This is an **input** param! So if the user creates the account, he/she has to tell us which version of the terms and conditions he/she agreed to. This is *different* from the scenario, that we import the database (version is `undefined`) and the server should respond with an object of type `User` and unfortunately that is a user who has not agreed to any terms and conditions. If the server would respond `undefined` for `User.termsAndConditions` the server would crash.

Does that make sense? @Tirokk So the exclamation mark here is correct.

roschaefer: Actually, the server will still crash. Ie. if you request `termsAndConditionsAgreedVersion` and we haven't used `neo4jgraphqljs` in the resolver. It is a rare situation though.

backend/src/schema/types/type/User.gql
```diff
@@ -152,6 +154,7 @@ type Query {
   ): [User]
 
   blockedUsers: [User]
+  currentUser: User
```
roschaefer: What is this? :slightly_frowning_face:

webapp/locales/de.json
```diff
@@ -21,6 +21,7 @@
     }
   },
   "site": {
+    "thx": "Danke!",
```
roschaefer: No abbreviations please and be a little more verbose on the confirmation message.

webapp/locales/en.json
```diff
@@ -21,6 +21,7 @@
     }
   },
   "site": {
+    "thx": "Thanks!",
```
roschaefer: See above

webapp/store/search.js
```diff
@@ -69,7 +69,7 @@ export const actions = {
       .then(res => {
         commit('SET_QUICK_RESULTS', res.data.findPosts || [])
       })
-      .catch(() => {
+      .catch(_err => {
```
roschaefer: Why this change?

mattwr18: I think I added that to debug why the search feature wasn't working on @ogerly machine, but it was because the indices hadn't been set up... then maybe I forgot to remove and the linting added an underscore(?) 
anyways, it can be removed

webapp/pages/terms-and-conditions.vue
```diff
@@ -22,6 +22,8 @@
 </template>
 
 <script>
+export const VERSION = '0.0.2'
```
Tirokk: Should go into @alina-beck const folder, see ☝🏼

webapp/store/search.js
Tirokk: @ogerly used the `err` for logs and as he removed the logs I told him that it is nice to have possible parameters visible …

PR #1338, opened 2019-08-21
Implement visual feedback if filter is active

### LINKED ISSUE #1335: 🚀 [Feature] Filter bubble indicator
## :rocket: Feature

Sometimes I don't realize that I have a filter currently active. It would be better user experience to have some visual feedback in that case.

As @ulfgebhardt said on discord:
> Es gibt keinen Indikator, dass der Filter gesetzt ist - dieser Feed ist nur nach Just for Fun gefiltert

### Design & Layout

Screenshot - 2019-08-21T114457 722

roschaefer:
### Issues

- close #1335 

### Todo

- [X] None

### DIFF
--- a/webapp/components/FilterPosts/FilterPosts.spec.js
+++ b/webapp/components/FilterPosts/FilterPosts.spec.js
@@ -54,6 +54,7 @@ describe('FilterPosts.vue', () => {
       'postsFilter/TOGGLE_CATEGORY': jest.fn(),
     }
     getters = {
+      'postsFilter/isActive': () => false,
       'auth/user': () => {
         return { id: 'u34' }
       },
@@ -63,7 +64,7 @@ describe('FilterPosts.vue', () => {
     const openFilterPosts = () => {
       const store = new Vuex.Store({ mutations, getters })
       const wrapper = mount(FilterPosts, { mocks, localVue, propsData, store })
-      menuToggle = wrapper.findAll('a').at(0)
+      menuToggle = wrapper.findAll('button').at(0)
       menuToggle.trigger('click')
       return wrapper
     }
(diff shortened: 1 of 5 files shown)


roschaefer:
@mattwr18 no problem to make only the icon green. You would have a `<ds-button icon="filter" ghost primary>` then. Look here: https://styleguide.human-connection.org/navigation/dsbutton

I removed the `ghost` attribute on purpose when it's primary because I could hardly see the difference.

alina-beck:
@mattwr18 @roschaefer sorry I didn't respond earlier! I think what we could do – after the release – is display the active filters in the same place we display the active hashtag. Ideally with the possibility to disable all filters (or even single filters) right there. Would you like that?

appinteractive:
@alina-beck do you mean in the search bar?

alina-beck:
@appinteractive no, I mean in the one below – which only displays at the moment when filtering by hashtag:

Screenshot 2019-09-02 at 10 11 04

That is, _if_ we want to keep this at all. I definitely see some room for design improvements there. :smile:

appinteractive:
I don’t know if duplicating that stuff there makes sense here as it should be in the searchbar (at least the hashtag and the x button) at least the shadow has to die before it starts to look a bit better. The filter also may apply to other areas like the recommended section so it would anyway only make sense to show it in the topbar where it belongs. 

That’s just my humble opinion.

alina-beck:
I would prefer to have it in the topbar as well. The question then is how exactly this would look like and what to do when too many filters are active to display them side by side. Do you have a solution in mind already?

appinteractive:
Something like that?

grafik

appinteractive:
I think that is already integrated to the input field that we use

alina-beck:
For manual filters maybe, but not for the ones selected in the filter menu.

appinteractive:
Okay then I don’t know that you are talking about.

alina-beck:
I just checked – this is neither implemented in the current alpha nor nitro version of the app. The search just gives you a dropdown with corresponding posts (i.e. you can't just enter a search term and hit return) and the selected filters are only displayed in the filter menu. This is what I mean:

Screenshot 2019-09-03 at 10 03 00

As soon as you close the menu there is no way to tell which filters are currently selected. This whole PR was about showing the user that the content is filtered (which we now did by coloring the filter button green) but it would be even better to see _which_ filters are active. And I would very much like a solution similar to the one you proposed above to do this. 🙂

appinteractive:
I will push a repo with some stuff I experimenting with in the next days, there we could just try out different approaches before integrating them, if you are interested. 

I currently doing that for the comments (how to display reply’s) and the topbar (best ways for responsiveness)

PR #1347, opened 2019-08-22
Remove custom entrypoint, and db_setup.sh

mattwr18:


mattwr18:
we like unique constraints... and it helps the build pass...closing

PR #1349, opened 2019-08-22
Fix expiration dates of JWT and cookie

roschaefer:
This won't fix the bug that can happen in `nuxtServerInit`. However,
according to the docs, we accepted JWT which was valid for 1000 days and
our cookie was valid for 3 days - completely weird.

See:
https://github.com/auth0/node-jsonwebtoken
https://github.com/nuxt-community/apollo-module

### Issues

- None

### Todo

- [X] None

### DIFF
--- a/backend/src/jwt/encode.js
+++ b/backend/src/jwt/encode.js
@@ -4,7 +4,7 @@ import CONFIG from './../config'
 // Generate an Access Token for the given User ID
 export default function encode(user) {
   const token = jwt.sign(user, CONFIG.JWT_SECRET, {
-    expiresIn: 24 * 60 * 60 * 1000, // one day
+    expiresIn: '1d',
     issuer: CONFIG.GRAPHQL_URI,
     audience: CONFIG.CLIENT_URI,
     subject: user.id.toString(),
--- a/webapp/nuxt.config.js
+++ b/webapp/nuxt.config.js
@@ -271,7 +271,7 @@ module.exports = {
   apollo: {
     tokenName: 'human-connection-token', // optional, default: apollo-token
     cookieAttributes: {
-      expires: 3, // optional, default: 7 (days)
+      expires: 1, // optional, default: 7 (days)
     },
     // includeNodeModules: true, // optional, default: false (this includes graphql-tag for node_modules folder)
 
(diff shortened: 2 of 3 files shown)


backend/src/jwt/encode.js
```diff
@@ -4,7 +4,7 @@ import CONFIG from './../config'
 // Generate an Access Token for the given User ID
 export default function encode(user) {
   const token = jwt.sign(user, CONFIG.JWT_SECRET, {
-    expiresIn: 24 * 60 * 60 * 1000, // one day
+    expiresIn: '1d',
```
mattwr18: I think this is more human readable and doesn't need a comment to explain it, but how is this fixing the expiration date? both are valid, correct?

webapp/nuxt.config.js
```diff
@@ -271,7 +271,7 @@ module.exports = {
   apollo: {
     tokenName: 'human-connection-token', // optional, default: apollo-token
     cookieAttributes: {
-      expires: 3, // optional, default: 7 (days)
+      expires: 1, // optional, default: 7 (days)
```
mattwr18: this is fixing it since it's setting the cookie to expire after the same period the jwt expires?

webapp/store/auth.js
```diff
@@ -144,10 +144,4 @@ export const actions = {
     commit('SET_TOKEN', null)
     return this.app.$apolloHelpers.onLogout()
   },
-
-  register({ dispatch, commit }, { email, password, inviteCode, invitedByUserId }) {},
-  async patch({ state, commit, dispatch }, data) {},
-  resendVerifySignup({ state, dispatch }) {},
-  resetPassword({ state }, data) {},
-  setNewPassword({ state }, data) {},
```
mattwr18: unrelated changes, but a clean up, correct?

backend/src/jwt/encode.js
roschaefer: The was a misconception about the meaning of this number. It's **seconds** not milliseconds. So the previous expiration was after 1000 days.

webapp/store/auth.js
roschaefer: Yes.

webapp/nuxt.config.js
roschaefer: This is not fixing the `nuxtServerInit` bug at all. Just an alignment that cookie expiration and JWT expiration should align.

PR #1351, opened 2019-08-22
Fix database import, pagination and counters on user profile page

### LINKED ISSUE #1333: 🐛 [Bug] User profile page: Result set sizes !== pagination counts
## :bug: Bugreport

@ulfgebhardt wrote us in Discord:

> Es werden 4 Kommentiert angezeigt, aber die Liste enthält 6 Einträge https://nitro-production.human-connection.org/profile/5acfe2083f62d4000b5b423c/dornhoeschen

The displayed counts of authored posts, commented posts and shouted posts are not the same as the actual number of results for the respective queries

1. Login at https://nitro-production.human-connection.org/
2. https://nitro-production.human-connection.org/profile/5acfe2083f62d4000b5b423c/dornhoeschen
3. Click on "commented"
4. Displayed number of results and actual number of results differ

Screenshot - 2019-08-21T113249 025

Screenshot - 2019-08-21T113404 055

### Additional context
The root of the problem is that we use the counts on the `User` type like:
https://github.com/Human-Connection/Human-Connection/blob/master/backend/src/schema/types/type/User.gql#L58
as pagination limit. See here:
https://github.com/Human-Connection/Human-Connection/blob/master/webapp/pages/profile/_id/_slug.vue#L306
But the filtered list of results might differ from that. Especially after we implemented #1209 the lists will vary **per user** who is looking at the profile page.

Therefore it is less buggy to use a pagination limit like "fetch and see if result.length < pageSize" instead of a total limit. Of course a correct total for any query would be the perfect solution but `neo4j-graphql-js` does not support this. Example code:
https://github.com/Human-Connection/Human-Connection/blob/master/webapp/pages/index.vue#L173

roschaefer:
Ok, so here are multiple issues:
1. In cypher, `NOT NULL` will return `NULL` not `FALSE`. If we want
`FALSE` to be set in the database import, we should use `COAELESCE`
to find the first not-null value.
See:
https://neo4j.com/docs/cypher-manual/current/syntax/working-with-null/
https://markhneedham.com/blog/2017/02/22/neo4j-null-values-even-work/

2. I removed the `disabled` and `deleted` checks on the commented
counter. With `neo4j-graphql-js` it is not possible to filter on the
join models (at least not without a lot of complexity) for disabled or
deleted items. Let's live with the fact that the list of commented posts
will include those posts, where the user has deleted his comment or where
the user's comment was disabled. It's being displayed as "not available"
so I think this is OK for now.

3. De-couple the pagination counters from the "commented", "shouted"
etc. counters. It might be that the list of posts is different for
different users. E.g. if the user has blocked you, the "posts" list
will be empty. The "shouted" or "commented" list will not have the
posts of the author. If you are a moderator, the list will include
disabled posts. So the counters are not in sync with the actual list
coming from the backend. Therefore I implemented "fetch and check if
resultSet < pageSize" instead of a global counter.

### Issues

- fix #1333 

### Todo

- [X] None

### DIFF
--- a/deployment/legacy-migration/maintenance-worker/migration/neo4j/contributions/contributions.cql
+++ b/deployment/legacy-migration/maintenance-worker/migration/neo4j/contributions/contributions.cql
@@ -137,8 +137,8 @@ p.contentExcerpt = post.contentExcerpt,
 p.visibility     = toLower(post.visibility),
 p.createdAt      = post.createdAt.`$date`,
 p.updatedAt      = post.updatedAt.`$date`,
-p.deleted        = COALESCE(post.deleted,false),
-p.disabled       = NOT post.isEnabled
+p.deleted        = COALESCE(post.deleted, false),
+p.disabled       = COALESCE(NOT post.isEnabled, false)
 WITH p, post
 MATCH (u:User {id: post.userId})
 MERGE (u)-[:WROTE]->(p)
--- a/backend/src/schema/resolvers/helpers/Resolver.js
+++ b/backend/src/schema/resolvers/helpers/Resolver.js
@@ -61,7 +61,6 @@ export default function Resolver(type, options = {}) {
         const id = parent[idAttribute]
         const statement = `
           MATCH(u:${type} {${idAttribute}: {id}})${connection}
-          WHERE NOT related.deleted = true AND NOT related.disabled = true
           RETURN COUNT(DISTINCT(related)) as count
         `
         const result = await instance.cypher(statement, { id })
--- a/backend/src/schema/resolvers/users.js
+++ b/backend/src/schema/resolvers/users.js
@@ -147,12 +147,15 @@ export default {
           'MATCH (this)<-[:BLOCKED]-(u:User {id: $cypherParams.currentUserId}) RETURN COUNT(u) >= 1',
       },
       count: {
-        contributionsCount: '-[:WROTE]->(related:Post)',
+        contributionsCount:
+          '-[:WROTE]->(related:Post) WHERE NOT related.disabled = true AND NOT related.deleted = true',
         friendsCount: '<-[:FRIENDS]->(related:User)',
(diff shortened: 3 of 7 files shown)


roschaefer:
Hang on people, I would like to double-checked if I might be wrong and deleted/disabled posts indeed get filtered. Please don't merge

roschaefer:
This is again ready for review. I'm still not absolutely sure about `FALSE` and `NULL` values in the database and how to treat them the same. Maybe the easiest fix is to change the database migration and don't set the property `disabled` or `deleted` at all. So we could only check for `EXISTS(p.disabled)`.

The improvement of this PR is that the pagination *total* is based on the actual result set instead of a counter. It might be that users still see inconsistencies between the counters and the results set. That will be the case if e.g. the user blacklisted you.

deployment/legacy-migration/maintenance-worker/migration/neo4j/contributions/contributions.cql
```diff
@@ -137,8 +137,8 @@ p.contentExcerpt = post.contentExcerpt,
 p.visibility     = toLower(post.visibility),
 p.createdAt      = post.createdAt.`$date`,
 p.updatedAt      = post.updatedAt.`$date`,
-p.deleted        = COALESCE(post.deleted,false),
-p.disabled       = NOT post.isEnabled
+p.deleted        = COALESCE(post.deleted, false),
+p.disabled       = COALESCE(NOT post.isEnabled, false)
```
roschaefer: this would return `NULL` otherwise

roschaefer:
@mattwr18 update the title and the description

deployment/legacy-migration/maintenance-worker/migration/neo4j/contributions/contributions.cql
mattwr18: thanks for the explanation

PR #1371, opened 2019-08-23
Fix "cannot read property id of null" in staging

roschaefer:
## 🍰 Pullrequest

I was wondering: Why don't we see AuthenticationErrors on Sentry? Why don't we have waay more spam? Await the incoming spam after this has been merged :wink: 

@datenbrei sb. deleted the user account with `admin` privileges on https://nitro-staging.human-connection.org/. On https://nitro-production.human-connection.org/ you can still log in with an admin account. The "cannot read property id of null" happens if there is an authentication error and sentry middleware is active.

@mattwr18 can you restore staging data please?

### Issues

- None

### Todo

- [X] None

### DIFF
--- a/backend/src/middleware/sentryMiddleware.js
+++ b/backend/src/middleware/sentryMiddleware.js
@@ -14,13 +14,16 @@ if (sentryConfigs.SENTRY_DSN_BACKEND) {
     },
     withScope: (scope, error, context) => {
       scope.setUser({
-        id: context.user.id,
+        id: context.user && context.user.id,
       })
       scope.setExtra('body', context.req.body)
       scope.setExtra('origin', context.req.headers.origin)
       scope.setExtra('user-agent', context.req.headers['user-agent'])
     },
   })
+} else {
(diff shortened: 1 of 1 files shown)


backend/src/middleware/sentryMiddleware.js
```diff
@@ -14,7 +14,7 @@ if (sentryConfigs.SENTRY_DSN_BACKEND) {
     },
     withScope: (scope, error, context) => {
       scope.setUser({
-        id: context.user.id,
+        id: context.user && context.user.id,
```
roschaefer: That's the actual fix.

mattwr18:
unfortunately, this is failing the build.. it kinda exploded the backend @roschaefer 
```
Test Suites: 4 failed, 25 passed, 29 total
Tests:       41 failed, 2 todo, 247 passed, 290 total
````
this is a good example of why I don't like to approve `PR`s until they pass the build @Kachulio1

PR #1376, opened 2019-08-23
Fix infinite-loading infinitely loading

roschaefer:
## 🍰 Pullrequest
The infinite loading component was indeed loading infinitely.

### Issues

- None

### Todo

- [X] None

### DIFF
--- a/webapp/pages/profile/_id/_slug.vue
+++ b/webapp/pages/profile/_id/_slug.vue
@@ -247,6 +247,7 @@
           :infinite-scroll-disabled="$apollo.loading"
           :infinite-scroll-distance="10"
           :infinite-scroll-throttle-delay="800"
+          :infinite-scroll-immediate-check="true"
         >
           <hc-load-more :loading="$apollo.loading" @click="showMoreContributions" />
         </div>
--- a/webapp/pages/profile/_id/_slug.vue
+++ b/webapp/pages/profile/_id/_slug.vue
@@ -244,8 +244,10 @@
         <div
           v-if="hasMore"
           v-infinite-scroll="showMoreContributions"
-          infinite-scroll-disabled="$apollo.loading"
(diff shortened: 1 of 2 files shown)


webapp/pages/profile/_id/_slug.vue
```diff
@@ -238,8 +238,10 @@
         <div
           v-if="hasMore"
           v-infinite-scroll="showMoreContributions"
-          infinite-scroll-disabled="$apollo.loading"
-          infinite-scroll-distance="10"
+          :infinite-scroll-immediate-check="true"
```
mattwr18: what is the purpose of adding this?
doesn't this default to true anyways?
https://sroze.github.io/ngInfiniteScroll/documentation.html
or am I looking at a different library??

webapp/pages/profile/_id/_slug.vue
```diff
@@ -238,8 +238,10 @@
...
+          :infinite-scroll-disabled="$apollo.loading"
+          :infinite-scroll-distance="10"
+          :infinite-scroll-throttle-delay="800"
```
mattwr18: default is 200ms... what went into the thinking to make it `800`? just your feeling? or was there something specific?

roschaefer: It was try/error with @alina-beck. We tested it with seed data. Maybe we can increase it even further after we learned about the low performance of our neo4j pod.

webapp/pages/profile/_id/_slug.vue
roschaefer: Right! I double-checked and it seems it is the default per documentation. I wonder why this fixed some our cypress tests earlier. Maybe another change or the documentation is just wrong.

roschaefer: Done, ping @mattwr18

PR #1377, opened 2019-08-25
Simplify docker entrypoint of neo4j image

roschaefer:
## 🍰 Pullrequest

So we have a permanent downtime of https://nitro-staging.human-connection.org/ and https://nitro-production.human-connection.org/. It's because the database is not running. If you e.g. `kubectl -n human-connection exec -it nitro-neo4j-<ID> bash` and you run `neo4j status` then you see:
`Neo4j is not running`.

I couldn't figure out what's the problem. However, when I install `htop` and run it in the pod, then I see a `cypher-shell` spinning, it is trying to connect and being terminated all the time:

2019-08-25-125303_1920x1080_scrot
2019-08-25-125347_1920x1080_scrot

These two screens :point_up: are always alternating. I don't think this is good, because it costs us CPU cycles and money to have a script that's continuously failing and restarting.

The docker image's entrypoint is wrapping the original entrypoint with this script that is trying to setup the indices. So, if the script is restarting in a loop then also this custom entrypoint never exits - even in case of an error. If the entrypoint would fail, then we would also see the neo4j pod being terminated. So that's why I believe it's better to remove the whole thing completely and instead tell our developers to setup the indices manually.

### Issues

- None

### Todo

- [X] None


roschaefer:
@mattwr18 and @Tirokk I've pushed the neo4j docker image from my machine (instead on Travis CI on `master`) to test it out on kubernetes.

1. `docker build -t humanconnection/neo4j:latest  neo4j` (you could pass in the BUILD_COMMIT if you wanted, have a look into [deploy.sh](https://github.com/Human-Connection/Human-Connection/blob/master/scripts/deploy.sh))
2. `docker push humanconnection/neo4j:latest`

And now both `production` and `staging` are back. I have no idea what went wrong. Maybe the script was stressing out the neo4j database too early when the kubernetes cluster was starting the pod? :woman_shrugging:

Issue #1390, opened 2019-08-26
🐛 [Bug] Index page on production data is too slow!


roschaefer:
## :bug: Bugreport
When I access the index page at https://nitro-production.human-connection.org/ and scroll down, I see too many `pending` requests being sent. This was no issue a while ago.

![2019-08-26-094415_1920x1080_scrot](https://user-images.githubusercontent.com/2110676/63674349-a23fb380-c7e6-11e9-9911-50cf10a96b9f.png)

### Steps to reproduce the behavior
1. Login at https://nitro-production.human-connection.org/
2. Scroll down
3. Observe the network panel of your developer tools
4. See to many slow requests

### Expected behavior

The backend should be much more responsive.

### Version & Environment
```sh
robert@e480 ~> chromium --version
Chromium 76.0.3809.100 Arch Linux
robert@e480 ~> cat /proc/version
Linux version 5.2.8-arch1-1-ARCH (builduser@heftig-63215) (gcc version 9.1.0 (GCC)) #1 SMP PREEMPT Fri Aug 9 21:36:07 UTC 2019
```

### Additional Context

![2019-08-26-095220_1920x1080_scrot](https://user-images.githubusercontent.com/2110676/63674553-0febdf80-c7e7-11e9-973a-25681d77c8cb.png)

The main memory is already full when the database is almost idle!

mattwr18:
I'll post my findings here so that we have a copy of them... 
I set the `Post` query debug to true
```
Query: {
    Post: async (object, params, context, resolveInfo) => {
      params = await filterForBlockedUsers(params, context)
      return neo4jgraphql(object, params, context, resolveInfo, true)
    }
}
```
in the posts resolver, and copy and pasted the `neo4j` generated query, cleaned it up and pasted it into localhost:7474/ neo4j browser
here is that query:
```
MATCH (post:Post {deleted:false, disabled:false}) 
WHERE (EXISTS((post)<-[:WROTE]-(:User)) 
AND NONE(user IN [(post)<-[:WROTE]-(_user:User) | _user] WHERE (user.id IN []))) 
WITH post ORDER BY post.createdAt DESC 
[...]

PR #1397, opened 2019-08-26
Check there are ids in the badIds array

mattwr18:
- if there are no ids, we shouldn't add an empty array since it adds unnecessarily to our auto-generated post query and affects greatly performance. see issue #1390

### Issues

- relates #1390... too early to say if it fixes it or not

### Todo

- [ ] Test it further @roschaefer


mattwr18:
Nice improvement @roschaefer... I'll merge when it passes the build

mattwr18:
I tried fixing this warning
```
router.scrollBehavior property is deprecated in favor of using ~/app/router.scrollBehavior.js file, learn more: https://nuxtjs.org/api/configuration-router#scrollbehavior  
```
but couldn't get it working... maybe I'm missing something?

webapp/store/search.js
```diff
@@ -46,8 +46,8 @@ export const actions = {
     await this.app.apolloProvider.defaultClient
       .query({
         query: gql`
-          query findPosts($query: String!) {
-            findPosts(query: $query, limit: 10) {
```
roschaefer: I find `query` a better naming for a string that is used for full-text search.

backend/src/schema/types/schema.gql
```diff
@@ -4,16 +4,15 @@ type Query {
...
     @cypher(
       statement: """
-      CALL db.index.fulltext.queryNodes('full_text_search', $query)
+      CALL db.index.fulltext.queryNodes('full_text_search', $filter)
       YIELD node as post, score
       MATCH (post)<-[:WROTE]-(user:User)
       WHERE score >= 0.2
       AND NOT user.deleted = true AND NOT user.disabled = true
       AND NOT post.deleted = true AND NOT post.disabled = true
-      AND NOT user.id in COALESCE($filter.author_not.id_in, [])
```
roschaefer: If you remove this hack (it is a hack) what is there as a replacement?

Is it possible that the cypress test checking the search results is a false negative? The build server should not pass...

mattwr18: no @roschaefer, I think what there is as a replacement is your function that gets called here https://github.com/Human-Connection/Human-Connection/blob/master/backend/src/schema/resolvers/posts.js#L35

webapp/store/search.js
mattwr18: I could agree with that, but if it's named query it breaks the functionality. not sure why exactly cause I'm not sure I completely understand the flow here, but it seems it passes through the `schema.gql` where we use the magic here https://github.com/Human-Connection/Human-Connection/blob/master/backend/src/schema/types/schema.gql#L10
then it gets passed to the `posts` resolver, which calls the `filterForBlockedUsers`
then cause there are no `badIds` it returns the `params`, otherwise it adds them to the `filter`, but the posts which match the query have already been returned from the database :point_up:

webapp/store/search.js
```diff
@@ -46,8 +46,8 @@ export const actions = {
...
+          query findPosts($query: String!, $filter: _PostFilter) {
+            findPosts(query: $query, limit: 10, filter: $filter) {
```
roschaefer: This is the only required change? Hell yeah.

PR #1398, opened 2019-08-26
Update to use <client-only> instead of <no-ssr>

mattwr18:


mattwr18:
I tried fixing this warning
```
router.scrollBehavior property is deprecated in favor of using ~/app/router.scrollBehavior.js file, learn more: https://nuxtjs.org/api/configuration-router#scrollbehavior  
```
but couldn't get it working... maybe I'm missing something?

mattwr18:
not sure why this is erroring out the build... tried restarting twice with the same results :thinking:

PR #1399, opened 2019-08-26
🍰 Correct the import of tags from the Alpha and refactor editor hashtags

### LINKED ISSUE #1395: 🐛 [Bug] Hashtags are imported with not allowed characters
## :bug: Bugreport

In the Alpha are hashtags which have chars that are not part of usual hashtags in the Nitro. (See rules for hashtags on Facebook etc. EN: https://en.wikipedia.org/wiki/Hashtag#Style , GER: https://de.wikipedia.org/wiki/Hashtag#Eigenschaften )

Bildschirmfoto 2019-08-26 um 11 29 23
Bildschirmfoto 2019-08-26 um 11 33 32
Bildschirmfoto 2019-08-26 um 11 35 25
Bildschirmfoto 2019-08-26 um 11 33 02

1. Create a new post.
2. Use a hashtag with typing of `#`.
3. Look at the menu.

### Expected behavior

Have hashtags after common hashtag rules.

Tirokk:
#809  🍰 Pullrequest

Correct the import of tags from the Alpha.
Remove the tags which can not be corrected.

I included the refactoring of the editor hashtag functionality as well !!!

### Issues
- fixes #1395
- relates #1276

### Todo

- [x] test if it’s working
- [x] editor hashtags
- [x] store the hashtags to the database
- [x] shorten hashtags in suggestion menu


Tirokk:
The functionality of the import has to be tested. @roschaefer @mattwr18

mattwr18:
that doesn't say anything about matching the `hashtag` and removing ones that only have numbers @roschaefer, but maybe we could refactor to use a combination of them both... I'll test it out

Tirokk:
No @roschaefer .
In a hashtags only `0-9a-bA-B` are allowed. If it starts with a number a letter has to follow somewhere. See the fixed issue #1395.

- There are several hashtags existing at the moment on production which are not allowed.
- And why converting all upper case chars to lower case? We are losing information, because it is a legitim and often user style element.

Tirokk:
Thanks a lot @mattwr18 that you have made this working. 🙏🏼😍
Cool 😎

mattwr18:
yeah, true... I was thinking of trying out `apoc.text.clean`, but @Tirokk makes a really good point, there is no reason we need to change the case of the hashtag, actually "experts" suggest using `Pascal` case for readability, and anyways they are legit hashtags. the purpose of this is to remove illegitimate ones ie ones with spaces, only numbers, etc...

@Tirokk this wouldn't remove the really long one from you screen shot, but I wasn't able to find any material, from what you provided, or checking myself that indicates there is a minimum or maximum, except for with twitter cause there is an overall limit of 140 characters.

One thing to keep in mind is that this would remove any accents from hashtags as they would not match `a-zA-Z`, and from what I have read, they are allowed in hashtags, so don't know what we can do about that....

Tirokk:
I haven’t red about accents in hashtags till now. But you brought up an issue @mattwr18 .
Here is info what Facebook and Twitter do:
https://soporte.easypromosapp.com/hc/en-us/articles/207886346-Tips-and-recommendations-to-define-your-contest-hashtag
It states:

About accents in the hashtags: 
- On Twitter: you can decide and communicate a hashtag with accents, but the platform will retrieve all the versions of this hashtag (with and without accent). You should discard the non-valid entries after they have been retrieved to the application.
- On Instagram: the platform will retrieve the identical hashtag you have written in the Instagram integration section. If the hashtag is written with accent, only those posts on Instagram with the accent on the hashtag will be retrieved to the application.

So, what are your suggestions @roschaefer , @ogerly , @alina-beck ?
Allow them? Or some of them? And therefore don’t filter them out of the Alphas.

### An other point is, what about other languages?

We rarely have Cyrillic letters in out Alpha hashtag database as an example. They seem to be allowed in general:
https://stackoverflow.com/questions/14823376/what-characters-are-allowed-in-twitter-hashtags

But for now I would suggest to get rid of them and later find a more general solution, if we establish other languages.

alina-beck:
I don't have a very strong opinion on this. Long-term I'd say it would be good to allow accents in hashtags as it feels more inclusive. But for now let's just go with the easiest solution? Or is this something we need to decide now because we already have hashtags with accents?

Tirokk:
I think in the Alpha are a lot of accents in the hashtags. They are not visible in the Nitro-Production, because the function `apoc.text.clean` used at the Moment converts them to multi letter ascii code.
I searched some in the Alpha and found them like `#Frühlingsnacht`, `#Beliebteste Beiträge`, or `#Nationalität` and so on.
We should convert the German ones at least …

Tirokk:
Now all unicode letters are working like on Twitter! @roschaefer @alina-beck @ogerly 
And the import preserves all the unicode letters as well … 😎
Yuhuuuu !!!!
Thanks a lot @mattwr18 😍🚀🚀💫💫💫

deployment/legacy-migration/maintenance-worker/migration/neo4j/contributions/contributions.cql
```diff
@@ -148,6 +148,10 @@ MATCH (c:Category {id: categoryId})
 MERGE (p)-[:CATEGORIZED]->(c)
 WITH p, post.tags AS tags
 UNWIND tags AS tag
-MERGE (t:Tag {id: apoc.text.clean(tag), disabled: false, deleted: false})
+WITH apoc.text.replace(tag, '[^\\p{L}0-9]', '') as tagNoSpacesAllowed
+CALL apoc.when(tagNoSpacesAllowed =~ '^((\\p{L}+[\\p{L}0-9]*)|([0-9]+\\p{L}+[\\p{L}0-9]*))$', 'RETURN tagNoSpacesAllowed', '', {tagNoSpacesAllowed: tagNoSpacesAllowed})
```
roschaefer: Are you trying to remove hashtags that don't match the regex?

I believe that the empty string `''` is not null, so we import the hashtag in any case.

mattwr18: when I was doing the import without this line of code it was failing on line `155` saying that it could not create a `Tag` with `null` as id, so it was coming from somewhere

Tirokk: The hashtag should be imported only if:
- after cleaning the hashtag from all not allowed chars:
```
apoc.text.replace(tag, '[^\\p{L}0-9]', ‚‘)
```
- it matches the pattern:
```
'^((\\p{L}+[\\p{L}0-9]*)|([0-9]+\\p{L}+[\\p{L}0-9]*))$'
```

In this new formulation it seems to me that the in front cleaning is not necessary anymore …

Tirokk: Ah @mattwr18 replied.
He can answer you question better and tryout, because I can not import my self.

mattwr18: that might actually be true @Tirokk... certainly I can test it out, if we don't mind waiting longer for it to get merged in

roschaefer: If this solution saves us complexity in the frontend and backend, please remove the obsolete source code! :relieved:

Tirokk: Would be great if you could care for it @mattwr18 !!! 😘
I am not in a big hurry. Would be just nice to have it merged before the next production gets deployed. @roschaefer

mattwr18: sure! I'll take care of it today

roschaefer: I debugged this with the following:
```
...
WITH p, post, post.tags AS tags
UNWIND tags AS tag
WITH post, apoc.text.replace(tag, '[^a-z^A-Z^0-9]', '') as tagNoSpacesAllowed
CALL apoc.when(tagNoSpacesAllowed =~ '^(([a-zA-Z]+[a-zA-Z0-9]*)|([0-9]+[a-zA-Z]+[a-zA-Z0-9]*))$', 'RETURN tagNoSpacesAllowed', '', {tagNoSpacesAllowed: tagNoSpacesAllowed})
YIELD value as validated
WITH validated, post
WHERE validated.tagNoSpacesAllowed IS NULL
CALL apoc.log.info('Post %s', post)
...
```
then the posts with tags that would become NULL are logged to `/var/log/neo4j/debug.log` (on my machine).

I used MongoDB compass to check out those contributions with tags converted to NULL:

2019-08-30-151035_1920x1080_scrot
2019-08-30-151144_1920x1080_scrot
2019-08-30-151236_1920x1080_scrot

mattwr18: ok, I don't see why they would be converted to null @roschaefer... is it supposed to be obvious?
I just retried the import without the `IS NOT NULL` and it failed as yesterday.

well, I guess the ones that return null would be the ones that after stripped contain only numbers?
do we care about this? I'm personally ok to remove those three tags, what about others?

roschaefer: Yes, it's because of the numbers. I'm OK to remove these tags, it's just that you could probably use a `COALESCE(somelist)` here.

mattwr18: but where is `somelist` coming from? these `tags` only become null as they are passed one-by-one through the regex and then the tags are one-by-one created with their id. we need to skip the `NULL` ones so as not to attempt to create them.

mattwr18:
hey @roschaefer this example is not exhaustive or clear enough
```
it('ignores Hashtag links with not allowed character combinations', () => {
  const content =
    '<p>Something inspirational about <a href="/search/hashtag/AbcDefXyz0123456789!*(),2" class="hashtag" target="_blank">#AbcDefXyz0123456789!*(),2</a>, <a href="/search/hashtag/0123456789" class="hashtag" target="_blank">#0123456789</a>, <a href="/search/hashtag/0123456789a" class="hashtag" target="_blank">#0123456789a</a> and <a href="/search/hashtag/AbcDefXyz0123456789" target="_blank">#AbcDefXyz0123456789</a>.</p>'
  expect(extractHashtags(content)).toEqual(['0123456789a', 'AbcDefXyz0123456789'])
})
```

deployment/legacy-migration/maintenance-worker/migration/neo4j/contributions/contributions.cql
```diff
@@ -148,6 +148,10 @@ MATCH (c:Category {id: categoryId})
...
```
alina-beck: Could you put these expressions in a variable and give them a human readable name? `stringConsistingOfNumbersOnly` might not be short and sweet but gives me an idea of what's happening here. :wink:

webapp/components/Editor/Editor.vue
```diff
@@ -212,8 +213,9 @@ export default {
     },
     sanitizeQuery(query) {
       if (this.suggestionType === HASHTAG) {
-        // remove all not allowed chars
-        query = query.replace(/[^a-zA-Z0-9]/gm, '')
+        // remove all unallowed chars
+        const regX = build('[^\\pL0-9]')
```
alina-beck: same as above – I could go check what this regex does exactly but would prefer to see it directly in the code, either as variable name or, since there's a comment already, as explanation in the comment

webapp/components/Editor/SuggestionList.vue
```diff
@@ -7,15 +7,15 @@
       :class="{ 'is-selected': navigatedItemIndex === index }"
       @click="selectItem(item)"
     >
-      {{ createItemLabel(item) }}
+      {{ createItemLabel(item) | truncate(50) }}
```
alina-beck: :+1:

webapp/components/Editor/SuggestionList.vue
```diff
@@ -7,15 +7,15 @@
...
     <template v-if="isHashtag">
-      <li v-if="!query" class="suggestion-list__item hint">
-        {{ $t('editor.hashtag.addLetter') }}
-      </li>
+      <li v-if="!query" class="suggestion-list__item hint">{{ $t('editor.hashtag.addLetter') }}</li>
       <template v-else-if="!filteredItems.find(el => el.id === query)">
         <li class="suggestion-list__item hint">{{ $t('editor.hashtag.addHashtag') }}</li>
-        <li class="suggestion-list__item" @click="selectItem({ id: query })">#{{ query }}</li>
+        <li class="suggestion-list__item" @click="selectItem({ id: query })">
+          #{{ query | truncate(50) }}
```
alina-beck: :+1:

I just had to google what this does – very confusing if you don't know about Vue.js filters... Great opportunity to learn about them, though! So thanks for this! :smile:

backend/src/middleware/hashtags/extractHashtags.js
```diff
@@ -20,7 +21,7 @@ export default function(content) {
   const hashtags = []
   urls.forEach(url => {
     let match
-    while ((match = ID_REGEX.exec(url)) != null) {
+    if ((match = exec(url, regX)) != null) {
```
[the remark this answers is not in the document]
Tirokk: Thanks you @alina-beck !!!
Nice to hear.

Yes, an improvement! 🤗😍

backend/src/middleware/hashtags/extractHashtags.js
```diff
@@ -20,7 +21,7 @@ export default function(content) {
   const hashtags = []
   urls.forEach(url => {
     let match
```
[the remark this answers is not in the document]
Tirokk: Same same …

deployment/legacy-migration/maintenance-worker/migration/neo4j/contributions/contributions.cql
Tirokk: Yes, a good improvement! 🤗😍
@mattwr18 did already … 😍

… he gave an other name  …
… and I changed the name again for clarification.

webapp/components/Editor/Editor.vue
Tirokk: Wonderful 😍

@mattwr18 did already … 😍
… and I changed the name again for clarification.

webapp/components/Editor/SuggestionList.vue
Tirokk: @mattwr18 did already … 😍
Thanks dear. 😘

webapp/components/Editor/SuggestionList.vue
Tirokk: @mattwr18 and I learned this by @roschaefer ! 😍💫

mattwr18:
hey, I misunderstood the diff between the `master` branch and this feature branch and thought that it had been removed with @roschaefer new epic `PR`, but I was wrong

PR #1400, opened 2019-08-26
Update the apollo cache

### LINKED ISSUE #1391: 🐛 [Bug] Deleted Comment reappears after editing another
## :bug: Bugreport

A once deleted comment reappears, when another comment on the same post ist edited.

1. Write 4 Comments
Bildschirmfoto_2019-08-26_11-03-18

2. Delete number 2
Bildschirmfoto_2019-08-26_11-03-52

3. Edit number 3 and save
4. Number 2 appears again
Bildschirmfoto_2019-08-26_11-04-19

### Expected behavior

A deleted comment should be gone for one and ever.

### Version & Environment
 Type: [Desktop]      
 - OS: [Linux]      
 - Browser: [Firefox] 
 - Version [68.0.1]  
 - Device: [Notebook]  

### Additional context

This does not happen, when the whole page is reloaded after deleting post number 2.

mattwr18:
### Issues

- fixes #1391


webapp/components/Comment.vue
```diff
@@ -140,24 +141,33 @@ export default {
...
+            }
+            store.writeQuery({ query: PostQuery(this.$i18n), data })
+          },
+        })
+        .then(res => {
+          this.$toast.success(this.$t(`delete.comment.success`))
+          this.$emit('deleteComment')
+        })
+        .catch(err => {
+          this.$toast.error(err.message)
```
mattwr18: @Tirokk, I originally tried using `update` with `async/await`, but I couldn't the syntax correct.
I just changed it to using a promise cause I have examples in the code base of it working... we could refactor it, but I'm not sure if there is a need, just to let you know why I refactored it so.

appinteractive: Just add `async` to the outer method, put a `await` before the request and wrap that inside the try/catch

webapp/components/Comment.vue
```diff
@@ -140,24 +141,33 @@ export default {
...
-          }
-        `
-        await this.$apollo.mutate({
-          mutation: gqlMutation,
+    deleteCommentCallback() {
+      this.$apollo
+        .mutate({
+          mutation: CommentMutations(this.$i18n).DeleteComment,
           variables: { id: this.comment.id },
+          update: store => {
```
appinteractive: `update: async store => {`

webapp/components/Comment.vue
```diff
@@ -140,24 +141,33 @@ export default {
...
+              variables: { id: this.post.id },
+            })
+
+            const index = data.Post[0].comments.findIndex(
+              deletedComment => deletedComment.id === this.comment.id,
+            )
+            if (index !== -1) {
+              data.Post[0].comments.splice(index, 1)
+            }
+            store.writeQuery({ query: PostQuery(this.$i18n), data })
```
appinteractive: `await store.writeQuery(`

webapp/components/Comment.vue
```diff
@@ -140,24 +141,33 @@ export default {
...
-        `
-        await this.$apollo.mutate({
-          mutation: gqlMutation,
+    deleteCommentCallback() {
+      this.$apollo
+        .mutate({
+          mutation: CommentMutations(this.$i18n).DeleteComment,
           variables: { id: this.comment.id },
+          update: store => {
+            const data = store.readQuery({
```
appinteractive: `const data = await store.readQuery(`

mattwr18:
hey thanks @appinteractive... the issue was that I was trying to put the update outside of the closing `}`, but inside the closing `)`... when I moved it inside both, it worked fine. I added the `async/awaits` to the update, readQuery, and writeQuery, as you suggested. I have a couple questions, if/when you have time to answer them.
1) is there any benefit to using `try/catch(async/await)` over `promises/then` syntax? I feel like it's just a style thing, but would love to know if there are any benefits/downsides to each.
2) the docs don't use `async/await` for updates or `readQuery`/`writeQuery`, is that cause the examples are in smart queries, or some other reason you know of?

PR #1407, opened 2019-08-27
Increase nginx client max body size, split size legacy migration

### LINKED ISSUE #1405: 🐛 [Bug] 413 Request Entity too large
## :bug: Bugreport

Bigger photos give an error

1. Write a new post
2. Add an bigger image (here 1,7 MB)
3. See the Error:
bild-zu-gross

### Expected behavior

It should be clear and statet what the biggest photo is, that can be uploaded, a clear limit.

### Version & Environment
 Type: [Desktop]      
 - OS: [Linux]      
 - Browser: [Firefox] 
 - Version [68.0.1]  
 - Device: [Notebook]

mattwr18:
- allows the client to send files at least the size we agreed in the vue-dropzone
- we don't need such a small split size as we don't run it on the remote server

### Issues

- fixes #1405

### DIFF
--- a/deployment/legacy-migration/maintenance-worker/migration/mongo/.env
+++ b/deployment/legacy-migration/maintenance-worker/migration/mongo/.env
@@ -12,6 +12,6 @@
 # On Windows this resolves to C:\Users\dornhoeschen\AppData\Local\Temp\mongo-export (MinGW)
 EXPORT_PATH='/tmp/mongo-export/'
 EXPORT_MONGOEXPORT_BIN='mongoexport'
-MONGO_EXPORT_SPLIT_SIZE=100
+MONGO_EXPORT_SPLIT_SIZE=4000
 # On Windows use something like this
 # EXPORT_MONGOEXPORT_BIN='C:\Program Files\MongoDB\Server\3.6\bin\mongoexport.exe'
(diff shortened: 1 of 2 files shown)


mattwr18:
learned about kubernetes and annotations here
https://kubernetes.github.io/ingress-nginx/user-guide/nginx-configuration/annotations/#custom-max-body-size

Tirokk:
Hey @mattwr18 ,

that's a good point to increase the kubernetes body proxy size. 👍🏼🚀

That works well so far, but I found a lack in functionality which you could solve in this PR or (somebody else) in an other one:

If you place an image bigger than 5m as the teaser image you have an error message. But it appears on the post page. Then afterwards you will not have it in your post.
So I would suggest to not have it in your post on creation if there’s an error. Or?

Bildschirmfoto 2019-08-28 um 09 34 44

Bildschirmfoto 2019-08-28 um 09 35 04

deployment/legacy-migration/maintenance-worker/migration/mongo/.env
```diff
@@ -12,6 +12,6 @@
 # On Windows this resolves to C:\Users\dornhoeschen\AppData\Local\Temp\mongo-export (MinGW)
 EXPORT_PATH='/tmp/mongo-export/'
 EXPORT_MONGOEXPORT_BIN='mongoexport'
-MONGO_EXPORT_SPLIT_SIZE=100
+MONGO_EXPORT_SPLIT_SIZE=4000
```
roschaefer: code creep - but actually I think that's a better default

PR #1411, opened 2019-08-28
Simplify cypress testing on docker

roschaefer:
This should fix our endless debugging

@ogerly all you have to do now is to `docker-compose up`. On your host
machine, you can run `yarn run cypress:open` to run the tests.

### Issues

- None

### Todo

- [X] None

### DIFF
--- a/cypress/README.md
+++ b/cypress/README.md
@@ -1,6 +1,16 @@
 # End-to-End Testing
 
-## Configure cypress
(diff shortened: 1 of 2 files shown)


roschaefer:
@ogerly if you `git checkout master` and `git pull` you should be able to run the cypress tests without any further configuration.

PR #1439, opened 2019-08-30
Move Neo4j configuration in configmap

roschaefer:
I think it's beneficial to have these settings in a central location.
That way, we're not running into the danger of overwriting the
`nitro-deployment` by accident with a `kubectl apply -f ...`.

### Issues

- None

### Todo

- [X] None


mattwr18:
@roschaefer these tests below are intermittently failing the build
Screenshot at 2019-08-31 09:35:11

PR #1451, opened 2019-09-02
Limit suggestions list to 15, add component tests

### LINKED ISSUE #1396: Feature: Less hashtag suggestions in the editor menu
## :bug: Bugreport

Hashtags menu is has to many items.

Bildschirmfoto 2019-08-26 um 11 33 02

1. Create a post.
2. Use hashtag by typing `#`.
3. Look at menu length …

### Expected behavior

Menu with not more then 20-30 items!?

mattwr18:
### Issues

- fixes #1396

### DIFF
--- a/webapp/components/Editor/Editor.spec.js
+++ b/webapp/components/Editor/Editor.spec.js
@@ -83,22 +83,35 @@ describe('Editor.vue', () => {
         expect(wrapper.vm.editor.extensions.options.mention.items()).toEqual(propsData.users)
       })
 
-      it('limits the suggestion list to 15 users', () => {
-        let manyUsersList = []
-        for (let i = 0; i < 25; i++) {
-          manyUsersList.push({ id: `user${i}` })
-        }
-        propsData.users = manyUsersList
-        wrapper = Wrapper()
-        expect(wrapper.vm.editor.extensions.options.mention.items()).toHaveLength(15)
-      })
-
       it('mentions is not an option when there are no users', () => {
         expect(wrapper.vm.editor.extensions.options).toEqual(
           expect.not.objectContaining({
             mention: expect.anything(),
           }),
         )
+      })
+
+      describe('limists suggestion list to 15 users', () => {
+        beforeEach(() => {
+          let manyUsersList = []
+          for (let i = 0; i < 25; i++) {
+            manyUsersList.push({ id: `user${i}` })
+          }
+          propsData.users = manyUsersList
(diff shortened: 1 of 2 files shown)


webapp/components/Editor/Editor.vue
```diff
@@ -205,10 +205,13 @@ export default {
       if (!query) {
         return items
       }
-      return items.filter(item => {
+
+      items = this.suggestionType === HASHTAG ? this.hashtags : this.users
```
alina-beck: I think I would rather pass all items (hashtags or users) to the new `Hashtag` or `Mention` instead of checking for the suggestion type again during the filtering process and then resetting items to the corresponding type.

We could still limit the number of displayed suggestions by returning `items.slice(0, 15)` when there is no query (in line `206`).

Seems a bit cleaner to me. What do you think?

webapp/components/Editor/Editor.vue
```diff
@@ -205,10 +205,13 @@ export default {
...
+      const filteredList = items.filter(item => {
         const itemString = item.slug || item.id
         return itemString.toLowerCase().includes(query.toLowerCase())
       })
+      return filteredList.slice(0, 14)
```
alina-beck: Is there a reason why you used `14` here and `15` above? I'd say let's maybe use the same number in both places? :smile:

webapp/components/Editor/Editor.spec.js
```diff
@@ -98,6 +111,16 @@ describe('Editor.vue', () => {
         expect(wrapper.vm.editor.extensions.options.hashtag.items()).toEqual(propsData.hashtags)
       })
 
+      it('limits the suggestion list to 15 users', () => {
```
alina-beck: copy pasta! 🍝 😆 should be `hashtags` instead of `users` here...

webapp/components/Editor/Editor.vue
mattwr18: yes, it does! I got it working, but then didn't take a step back to think about it critically... good catch!

webapp/components/Editor/Editor.vue
mattwr18: haha, just cause I originally had all as 14, but then forgot to change this one

webapp/components/Editor/Editor.spec.js
mattwr18: you caught me :open_mouth:

PR #1457, opened 2019-09-03
Fix #1394

### LINKED ISSUE #1394: 🐛 [Bug] A new post is added on the news feed page to the end of the loaded ones
## :bug: Bugreport

New posts is added on the news feed page to the end of the loaded ones.

Bildschirmfoto 2019-08-26 um 14 01 37

1. Create a new post.
2. Go to the news feed page.
3. See that your post is not in front of all other posts.
4. Scroll down and your post is at the end of the already loaded ones.
5. On automatic load it stay at this position.

### Expected behavior

Post should be in front of all other posts, because it is the newest one.

### LINKED ISSUE #1438: 🐛 [Bug] Mouse cursor over Logo seems to be non-clickable
## :bug: Bugreport

If the mouse cursor is over the top left logo it does not change to a "hand". This looks like, as if the logo would not be clickable, but it is.

1. Move the mouse over the top left logo

### Expected behavior

As on every other clickable part of the page the logo should also be appear as clickable.

### Version & Environment
 Type: [Desktop]      
 - OS: [Linux]      
 - Browser: [Firefox] 
 - Version [68.0.1]  
 - Device: [Notebook]  

### Additional context

Minor bug, not important, just not nice.

roschaefer:
This bug is actually an unpleasant combination of bugs. We first ran into:
https://github.com/Akryum/vue-apollo/issues/492

The people there suggest to use `no-cache`. But that would let us face
yet another bug:
https://github.com/Akryum/vue-apollo/issues/263

So I fixed this with an obscure workaround and assumed that whenever the
two results intersect then this is a cache update.

### Issues

- fix #1394 
- fix #1438

### Todo

- [X] None


roschaefer:
Hm, seems this does not play nice with our pagination. Probably we cannot rely on the fact that there have been duplicates.

roschaefer:
closed in favour of #1479

PR #1476, opened 2019-09-04
1393 fix show less bugginess

### LINKED ISSUE #1393: 🐛 [Bug] Show less functionality broken
## :bug: Bugreport

Long comments are truncated and there is a link to `show more`, which still works fine, but recently `show less` is broken

1. visit any post with comments longer than 180 characters
2. click on the `show more` link
3. click on the `show less` link
4. observe the comment doesn't get truncated 
5. Profit

### Expected behavior

The comment should be collapsible again

### Version & Environment
 Type: [Desktop]      
 - OS: [Linux/Ubuntu]      
 - Browser: [Firefox/Chromium] 
 - Version []  
 - Device: []  

### Additional context

 
2019-08-26-show-less-broken

mattwr18:
### Issues

- fixes #1393 

this fixes the bug that the `show more` link is shown when a @mention occurs for comments that are shorter than 180 characters... 
at the moment, we were just checking if there is any difference between content and contentExcerpt, then showing the show more, but for some reason, there is always a difference between comments content longer or shorter than 180 characters

on my local development, the show less is still buggy, because it used the `content-viewer`, which used the editor, but in production it seems to not be an issue.
### Todo

- [X] None

### DIFF
--- a/webapp/components/Comment.spec.js
+++ b/webapp/components/Comment.spec.js
@@ -63,6 +63,7 @@ describe('Comment.vue', () => {
         propsData.comment = {
           id: '2',
           contentExcerpt: 'Hello I am a comment content',
+          content: 'Hello I am comment content'
         }
       })
 
--- a/webapp/components/Comment.spec.js
+++ b/webapp/components/Comment.spec.js
@@ -63,7 +63,7 @@ describe('Comment.vue', () => {
         propsData.comment = {
           id: '2',
           contentExcerpt: 'Hello I am a comment content',
-          content: 'Hello I am comment content'
+          content: 'Hello I am comment content',
         }
       })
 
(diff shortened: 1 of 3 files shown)


mattwr18:
I could, but that is not a bug fix, so I think it shouldn't block a fix for an embarrassing bug get merged in
merge that :hankey:

mattwr18:
thanks @Tirokk, don't know if you saw my comments :point_up: in the description of the `PR`, but I also get that bug in a development environment, but since it doesn't seem to be reproducible in a production environment, I'm just ignoring it for now.
I think it's to due to the `Editor`

mattwr18:
the white space hasn't changed as I can tell... but I can understand

PR #1512, opened 2019-09-06
Refactor email middleware

roschaefer:
## 🍰 Pullrequest

Apart from the refactoring, this fixes the missing `from` attribute when sending signup mails.

### Issues

- None

### Todo

- [X] None


backend/src/schema/resolvers/passwordReset.spec.js
```diff
@@ -16,158 +21,203 @@ const getAllPasswordResets = async () => {
   return resets
```
mattwr18: where is this coming from?

backend/src/schema/resolvers/passwordReset.spec.js
```diff
@@ -16,158 +21,203 @@ const getAllPasswordResets = async () => {
...
 
-      const session = driver.session()
-      await createPasswordReset({ driver, email, issuedAt, code })
-      session.close()
+  const mutation = gql`
+    mutation($nonce: String!, $email: String!, $newPassword: String!) {
+      resetPassword(nonce: $nonce, email: $email, newPassword: $newPassword)
     }
+  `
+  const nonce = 'abcdef'
```
mattwr18: why do you have a default and then set the same value to a variable?

backend/src/schema/resolvers/passwordReset.spec.js
```diff
@@ -16,158 +21,203 @@ const getAllPasswordResets = async () => {
...
+  beforeEach(() => {
+    variables = { ...variables, newPassword: 'supersecret' }
```
mattwr18: weak password :stuck_out_tongue_closed_eyes:

backend/src/schema/resolvers/passwordReset.spec.js
mattwr18: so this calls `createPasswordReset` with the nonce `abcdef` as a default, then uses that same nonce to find the `PasswordReset` node, but if the user passes in a different nonce like `slkdjf`, then it will create a `PasswordReset` node with that `nonce`, but will not match the `resetPassword` nonce?

backend/src/schema/resolvers/passwordReset.js
```diff
@@ -2,39 +2,47 @@ import uuid from 'uuid/v4'
...
+    const records = transactionRes.records.map(record => {
+      const { email } = record.get('e').properties
+      const { nonce } = record.get('pr').properties
+      const { name } = record.get('u').properties
+      return { email, nonce, name }
+    })
+    response = records[0] || {}
+  } finally {
+    session.close()
+  }
```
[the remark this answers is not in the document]
roschaefer: I guess so, yes

backend/src/schema/resolvers/passwordReset.spec.js
roschaefer: two lines above?

backend/src/schema/resolvers/passwordReset.spec.js
```diff
@@ -137,7 +136,7 @@ describe('resetPassword', () => {
     describe('invalid email', () => {
       it('resolves to false', async () => {
         await setup()
-        variables = { ...variables, email: 'non-existent@example.org', nonce }
+        variables = { ...variables, email: 'non-existent@example.org', nonce: 'abcdef' }
```
mattwr18: I like this change!

PR #1513, opened 2019-09-06
Bugfix: `about me` not saved on signup

roschaefer:
## 🍰 Pullrequest

Our dear @Hardy-Peaceglobal has reported this issue. This should update the user's `about` attribute on signup.

### Issues

- there is no issue for that yet

### Todo

- [X] None

### DIFF
--- a/backend/src/seed/factories/users.js
+++ b/backend/src/seed/factories/users.js
@@ -5,7 +5,7 @@ import slugify from 'slug'
 
 export default function create() {
   return {
-    factory: async ({ args, neodeInstance }) => {
+    factory: async ({ args, neodeInstance, factoryInstance }) => {
       const defaults = {
         id: uuid(),
         name: faker.name.findName(),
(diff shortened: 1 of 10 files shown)


backend/src/seed/factories/users.js
```diff
@@ -24,7 +24,7 @@ export default function create() {
       }
       args = await encryptPassword(args)
       const user = await neodeInstance.create('User', args)
-      const email = await neodeInstance.create('EmailAddress', { email: args.email })
+      const email = await factoryInstance.create('EmailAddress', { email: args.email })
```
mattwr18: what's the difference between `factoryInstance` and `neodeInstance`?

roschaefer: We might want to refactor this with https://github.com/aexmachina/factory-girl (although the repo looks rather old).

`factoryInstance` is a reference to the other factories. If you would do `factoryInstance.create(..)` that would give you all the defaults and also setup associations. If you use `neodeInstance` you would end up with a `User` without `EmailAddress` etc.

mattwr18: ahhh, ok! great... thanks

mattwr18:
great job @roschaefer 
and just at the threshold of merging without multiple `PR` reviews!!
:stuck_out_tongue_closed_eyes:

PR #1516, opened 2019-09-08
Remove nodemon, use `nuxt` cli tools

roschaefer:
@appinteractive can you tell me why we had the custom code in
`server/index.js`?

### Issues

- None

### Todo

- [X] None


backend/Dockerfile
```diff
@@ -15,10 +15,10 @@ COPY .env.template .env
 CMD ["yarn", "run", "start"]
 
 FROM base as builder
-RUN yarn install --frozen-lockfile --non-interactive
+RUN yarn install --production=false --frozen-lockfile --non-interactive
 COPY . .
 RUN cp .env.template .env
-RUN yarn run build
+RUN NODE_ENV=production yarn run build
```
mattwr18: this is a bit confusing for me.. I saw your commit message "webapp was built with `NODE_ENV=production` but started with
`NODE_ENV=development`"
but it seems that now you are doing the opposite, oder? Line 18 says don't install production dependencies... but build it in a production environment??

roschaefer: In `nuxt.config.js` our webapp is configured to build for "modern" browsers if `NODE_ENV == production`. If you later `yarn run start` it will complain if the modern build is missing. Yet, you must not to set `ENV NODE_ENV=production` in the docker image, because that will confuse `yarn run test`.

It took me some failing builds to figure that out.

roschaefer: I think for `backend` it's indeed unnecessary to build in production. However it also does not hurt..

Issue #1367, opened 2019-08-23
🐛 [Bug] GraphQL error locationName is invalid


datenbrei:
## :bug: Bugreport

When saving the personal information in my data an error arises when the locationName is changed.
![location-name-error](https://user-images.githubusercontent.com/13457963/63573323-4f1af600-c585-11e9-80ab-f260091f9efa.png)

### Steps to reproduce the behavior
1. Log in
2. Go to Settings
3. Change the city or region
4. save

### Expected behavior

No error should arise

### Version & Environment
 Type: [Desktop]      
 - OS: [Linux]      
 - Browser: [Firefox] 
 - Version [68.0.1]  
 - Device: [Notebook]  

### Additional context

This bug may be related to bug:
[GraphQl error when saving the residence in the profile settings #850](https://github.com/Human-Connection/Human-Connection/pull/905)

ogerly:
there is a problem. and that is with hamburg. city state and city .. no federal state .. a problem with the location.ids
I've explained this a lot before. 
it requires a decision which mistakes are bearable ;) and how we must approach them by hand. 
hamburg == such a special extra case

ogerly:
I've worked on that bug before. two ways why a bug come like yours comes. 
1. the external api node was not available when saving. 
2. the external api node could not be called. 

i tested in the nitro production. 
the problem occurs with all locations... this is not the error ... 
in the current local master this error does not occur.

mattwr18:
This is being closed because it has been fixed... @datenbrei please test it out on `nitro-production` and `nitro-staging`, but it turns out that our `MAPBOX_TOKEN` base64 encoded access token was off by one character.

I've use `kubectl` to update them both on `nitro-staging`, and `nitro-production`
also, we've added debugging to `locations.js` to make it easier to see what the issue is in the future if we run into any other problems

PR #1534, opened 2019-09-09
Add console log for erros geolocation

mattwr18:
@roschaefer, I was trying to get this without needing to merge into master, but I'm not sure how I can edit a file in place and restart the server... 
I was able to narrow down issue #1367 to our backend/src/middleware/nodes/locations.js
there are two places in `createOrUpdateLocations` where the error message `locationName is invalid` can possibly be called.

### DIFF
--- a/webapp/pages/settings/index.vue
+++ b/webapp/pages/settings/index.vue
@@ -101,7 +101,7 @@ export default {
     async submit() {
       this.loadingData = true
       const { name, about } = this.formData
-      let { locationName } = this.formData
+      let { locationName } = this.formData || this.currentUser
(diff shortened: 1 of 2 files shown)


webapp/pages/settings/index.vue
```diff
@@ -101,7 +101,7 @@ export default {
     async submit() {
       this.loadingData = true
       const { name, about } = this.formData
-      let { locationName } = this.formData
+      let { locationName } = this.formData || this.currentUser
```
roschaefer: Although this is already changing the behaviour that you want to debug.

PR #1537, opened 2019-09-09
[WIP] 🍰 Fixes a create and update comment problem in the comments list

### LINKED ISSUE #1455: 🐛 [Bug] Creating multiple comments on a post does not update the CommentList
## :bug: Bugreport

Writing a comment should not require reloading the page to update the comment list

1. go to https://nitro-production.human-connection.org/
2. select a post at random, or one that you'd like to comment on
3. make a comment
4. write another comment, even more if you'd like
5. notice they don't appear
6. reload the page
7. confirm you comments were created
### Expected behavior

A user should see their comment without the need to reload the page
Screenshot from 2019-09-03 23-24-12
Screenshot from 2019-09-03 23-24-26

Tirokk:
## 🍰 Pullrequest

Fixes an issue with writing multiple comments in a row and can sometimes not see them in the browser before you reload.

### Issues

- fixes #1455

### Todo

- [ ] writing frontend tests for the edited events


webapp/components/CommentForm/CommentForm.vue
```diff
@@ -2,18 +2,18 @@
   <ds-form v-model="form" @submit="handleSubmit">
     <template slot-scope="{ errors }">
       <ds-card>
-        <hc-editor
-          ref="editor"
-          :users="users"
-          :hashtags="null"
```
mattwr18: nice, you tested and this is not needed after all, right?

webapp/components/Comment.vue
```diff
@@ -132,6 +135,9 @@ export default {
     editCommentMenu(showMenu) {
       this.openEditCommentMenu = showMenu
     },
+    async updateComment(comment) {
```
mattwr18: why is this `async` if there is no `await`?
Is this really the best way to handle this situation @Tirokk @roschaefer 
we are emitting an event from a grandchild to a parent component, which just emits an event to its parent(granparent of CommentForm) so that it can update the list??
Seems strange this would be the best way to do this....

Tirokk: The `async` slipped in by copy and paste I suppose and I have overseen it. @mattwr18 
I removed it.

This @roschaefer suggested and I followed him, because I don’t know a better way yet. And we need something what works for sure for the moment.

May the `vue-states` + `vue-history` packages from our last Vue MeetUp would help here in the future?

webapp/components/CommentForm/CommentForm.vue
Tirokk: Yeah !!!

PR #1540, opened 2019-09-09
Remove obsolete code

roschaefer:
I believe this is obsolete code. Could someone double-check?

@mattwr18 I tried to see any difference to `master` and I couldn't. Did
I miss sth.?

For context: During my work on https://github.com/Human-Connection/Human-Connection/pull/1517
the webapp was complaining about missing mutations when I changed the
dropdown.

### Issues

- Refactoring

### Todo

- [X] None


Tirokk:
This code is obsolet. I removed it yesterday already in my PR for the comment list update, except the placeholder. But the placeholder translation is now managed in the editor it self when I think about. So it should be obsolete as well.

Tirokk:
My PR is #1537

PS: I can not edit my comment on iPhone.

webapp/components/Editor/Editor.vue
```diff
@@ -70,7 +69,9 @@ export default {
     }
   },
   computed: {
-    ...mapGetters({ placeholder: 'editor/placeholder' }),
+    placeholder() {
+      return this.$t('editor.placeholder')
+    },
```
mattwr18: much better solution!!
state is the root of all evil!

PR #1542, opened 2019-09-10
Test filter by followed by

### LINKED ISSUE #1538: [Unit tests] Add unit tests for filtering by Followed By
## :rocket: Feature

Add unit tests to test filtering posts by user `followed by`

aonomike:
## 🍰 Pullrequest

- The filterBubble.js file no longer exists hence tests to check that posts can be filtered by user followed have been moved to posts.spec.js
- Refactor variables used in the tests to have better names
### Issues

- Fixes #1538 

### Todo

- [X] None


backend/src/schema/resolvers/posts.spec.js
```diff
@@ -91,44 +91,63 @@ afterEach(async () => {
...
+            Post(filter: $filter) {
+              id
+            }
+          }
+        `
+        const expected = [{ id: 'happy-post' }, { id: 'cry-post' }, { id: 'post-by-followed-user' }]
+        variables = { filter: {} }
+        await expect(query({ query: postQueryNoFilters, variables })).resolves.toMatchObject({
+          data: {
+            Post: expect.arrayContaining(expected),
```
roschaefer: If you use `toMatchObject` you don't need the `expect.arrayContaining` unless you expect even more items than the three.

backend/src/schema/resolvers/posts.spec.js
```diff
@@ -91,44 +91,63 @@ afterEach(async () => {
...
 
+    describe('no filter', () => {
+      it('returns all posts', async () => {
+        const postQueryNoFilters = gql`
+          query Post($filter: _PostFilter) {
+            Post(filter: $filter) {
+              id
+            }
+          }
+        `
```
roschaefer: I believe it helps to avoid false negatives (test does not fail although there is a bug) if you use the very same graphql query throughout the scope. `await exect(...).resolves.toMatchObject(...)` should be lenient enough about any additional data in the `actual` response.

backend/src/schema/resolvers/posts.spec.js
```diff
@@ -140,45 +159,88 @@ describe('Post', () => {
...
+                emotions: [{ emotion: 'happy' }],
+              },
+            ],
+          },
+        }
+        await user.relateTo(happyPost, 'emoted', { emotion: 'happy' })
+        variables = { ...variables, filter: { emotions_some: { emotion_in: ['happy'] } } }
+        await expect(
+          query({ query: postQueryFilteredByEmotions, variables }),
+        ).resolves.toMatchObject(expected)
```
roschaefer: Here you omit the `expect.arrayContaining` :+1:

backend/src/schema/resolvers/posts.spec.js
mattwr18: Actually, we started off by not using `arrayContaining` and the test failed because it returned the posts in a different order than our `expected` array, so we refactored it to say the order is not important.

backend/src/schema/resolvers/posts.spec.js
mattwr18: yes, we wanted to use this for the `multiple emotions` test as well, but it failed as I stated :point_up:

backend/src/schema/resolvers/posts.spec.js
roschaefer: You could `orderBy` in the query? It does not really matter. I like your approach, too! :+1:

PR #1549, opened 2019-09-11
Speed up builds, `docker-compose build --parallel`

roschaefer:
Found this: https://devops.stackexchange.com/questions/1343/is-it-possible-to-build-docker-images-using-docker-compose-concurrently

Tought: Let's try out immediately

### Issues

- None

### Todo

- [X] None


roschaefer:
hm, not really faster..

PR #1550, opened 2019-09-11
Refactor createdAt to use DateTime and be adding at insertion in the database 

### LINKED ISSUE #1494: 🐛 [Bug] Sometimes `createdAt` not set in `CreatePost`
## :bug: Bugreport

This is probably a hard to reproduce bug but we have post without a `createdAt` attribute in the dabase:
2019-09-05-195250_1920x1080_scrot

https://nitro-production.human-connection.org/post/a8ab8deb-a3ba-42bb-9764-c6a4d209272c/smiling-face-1

### Expected behavior
The `createdAt` attribute should always be set.

mattwr18:
- not happy with this solution, but it adds a test case, which can be
monitored on every test run to see if we ever have a post created
without createdAt, also it sets createdAt in the CreatePost resolver if
for some reason it didn't get set in the dateTimeMiddleware

- Co-authored-by: Mike Aono <aonomike@gmail.com>

### Issues

- fixes #1494

### DIFF
--- a/backend/src/schema/resolvers/posts.js
+++ b/backend/src/schema/resolvers/posts.js
@@ -89,15 +89,7 @@ export default {
         UNWIND $categoryIds AS categoryId
         MATCH (category:Category {id: categoryId})
         MERGE (post)-[:CATEGORIZED]->(category)
-        RETURN post {.id,
-          .title,
-          .content, 
-          .slug, 
-          .disabled, 
-          .deleted, 
-          .language, 
-          created_at: { formatted: toString(post.created_at) }} 
-          AS post`
+        RETURN post, toString(post.created_at) as postCreatedAt`
 
       const createPostVariables = { userId: context.user.id, categoryIds, params }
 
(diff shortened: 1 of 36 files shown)


backend/src/schema/resolvers/posts.js
```diff
@@ -79,6 +79,7 @@ export default {
       delete params.categoryIds
       params = await fileUpload(params, { file: 'imageUpload', url: 'image' })
       params.id = params.id || uuid()
+      params.createdAt = params.createdAt || new Date().toISOString()
```
appinteractive: We should really always working with adding utc dates to avoid issues down the road. 

Also: why the heck is it possible to overwrite the createdAt date, this should never, ever happen.

mattwr18: good question... I pushed up some code that would set the `created_at` in a cypher statement, and then parse the `post.created_at` to return a String as `createdAt`... it's not done yet, but I am for the day... 
would you and/or @roschaefer have a look and tell me if you don't like it/want to go in a different direction?

mattwr18:
Closing this as it's currently blocked by https://github.com/neo4j-graphql/neo4j-graphql-js/issues/272
We can re-open and finish when it's resolved

PR #1558, opened 2019-09-12
Push all docker images at VERSION to dockerhub

roschaefer:
Inspired by the tagging of e.g. `node` on dockerhub:
https://hub.docker.com/_/node/?tab=description

I would like to push the current version of all our images to dockerhub.
This is a first step to push to production later on.

If sb. increments the VERSION file, this is considered a release.

### Issues

- None

### Todo

- [X] None


roschaefer:
This will now push the tag `0`, `0.1` and `0.1.0` of all our apps.

PR #1571, opened 2019-09-13
🍰 Make hashtag links URL safe

### LINKED ISSUE #1276: 🐛 [Bug] Hashtag links are not URL safe

Tirokk:
## 🍰 Pullrequest

Make hashtag links URL safe.
Have the routing of a hashtag link direct.

### Issues

- fixes #1276

### Todo

- [x] direct routing of hashtag link


webapp/middleware/searchHashtag.js
```diff
@@ -0,0 +1,16 @@
+import { exec, build } from 'xregexp/xregexp-all.js'
+
+export default async ({ store, env, route, redirect }) => {
+  let publicPages = env.publicPages
+  // only affect non public pages
```
mattwr18: sorry, what does this mean? 
by `publicPages` do you mean pages that a user can visit without being signed in?
or is there some meaning for this I don't know about?

backend/src/middleware/hashtags/hashtagsMiddleware.spec.js
```diff
@@ -129,8 +129,8 @@ describe('hashtags', () => {
         )
       })
 
-      describe('afterwards update the Post by removing a Hashtag, leaving a Hashtag and add a Hashtag', () => {
-        // The already existing Hashtag has no class at this point.
+      describe('afterwards update the Post by removing a hashtag, leaving a hashtag and add a hashtag', () => {
```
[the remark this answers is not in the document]
Tirokk: I my self had to read it twice, but than understand.
I should have corrected this at that moment, but somehow got over it. @mattwr18 

There is already a post generated with #Democracy and #Liberty hashtags.
By updating the post with #Elections and #Liberty only this both should be related to the post. So the statement was one (#Democracy) is removed, one is left (#Liberty) and one is new (#Elections). Much to complicated …
But a statement like `only the both hashtags in the updated content are related` will be clear and much more understandable.

webapp/middleware/searchHashtag.js
Tirokk: This I have taken over from `webapp/middleware/authenticated.js`. @mattwr18 
In my understanding `publicPages ` do mean pages that a user can visit without being signed in. Yes.

backend/src/middleware/hashtags/hashtagsMiddleware.spec.js
roschaefer:
```suggestion
      describe('updates the Post by removing, keeping and adding one hashtag respectively', () => {
```

webapp/components/Editor/nodes/Hashtag.js
```diff
@@ -25,7 +25,7 @@ export default class Hashtag extends TipTapMention {
           'a',
           {
             class: this.options.mentionClass,
-            href: `/search/hashtag/${node.attrs.id}`,
+            href: `/search/hashtag/${encodeURI(node.attrs.id)}`,
```
roschaefer: Can't we use query parameters? :cry: 

I found out how to use [URL](https://nodejs.org/api/url.html) to encode/decode url params for relative URLs: Just use a dummy domain!

Example:

```js
const url = new URL('http://example.org')
url.searchParams.append('hashtag', 'ä  #')
url.search // returns ?hashtag=%C3%A4++%23
```

webapp/middleware/searchHashtag.js
roschaefer: You don't have to check public pages again. The `authenticated` middleware is one of two default middlewares checked for every route, as specified in [nuxt.config.js](https://github.com/Human-Connection/Human-Connection/blob/master/webapp/nuxt.config.js#L126).

I would not use a middleware for sth. that affects only one route, though!

webapp/middleware/searchHashtag.js
```diff
@@ -0,0 +1,16 @@
...
+  if (publicPages.indexOf(route.name) >= 0) {
+    return true
+  }
+
+  const regX = build('^/search/hashtag/((\\pL+[\\pL0-9]*)|([0-9]+\\pL+[\\pL0-9]*))$')
+  const matchHashtag = route.fullPath ? exec(decodeURI(route.fullPath), regX) : null
+
+  if (!matchHashtag) return true
+
+  return redirect(`/?hashtag=${encodeURI(matchHashtag[1])}`)
```
roschaefer: Something like this URL could be in the `href` above! :point_up:

webapp/nuxt.config.js
```diff
@@ -123,7 +123,7 @@ export default {
   ],
 
   router: {
-    middleware: ['authenticated', 'termsAndConditions'],
+    middleware: ['authenticated', 'termsAndConditions', 'searchHashtag'],
```
roschaefer:
```suggestion
    middleware: ['authenticated', 'termsAndConditions'],
```

You are adding a middleware for every route here. I would say the feature does not justify the added complexity :neutral_face:

webapp/pages/index.vue
```diff
@@ -75,7 +75,10 @@ export default {
     MasonryGridItem,
   },
   data() {
-    const { hashtag = null } = this.$route.query
+    let { hashtag = null } = this.$route.query
+    if (hashtag) {
+      hashtag = decodeURI(hashtag)
```
roschaefer: Weird, the hashtag is not automatically decoded? :confused:

Tirokk: May I add to much `decodeURI` …

backend/src/middleware/hashtags/extractHashtags.js
```diff
@@ -6,21 +6,21 @@ import { exec, build } from 'xregexp/xregexp-all.js'
...
-const regX = build('^/search/hashtag/((\\pL+[\\pL0-9]*)|([0-9]+\\pL+[\\pL0-9]*))$')
+const regX = build('^((\\pL+[\\pL0-9]*)|([0-9]+\\pL+[\\pL0-9]*))$')
 
 export default function(content) {
   if (!content) return []
   const $ = cheerio.load(content)
   // We can not search for class '.hashtag', because the classes are removed at the 'xss' middleware.
   //   But we have to know, which Hashtags are removed from the content as well, so we search for the 'a' html-tag.
-  const urls = $('a')
+  const ids = $('a[data-hashtag-id]')
```
mattwr18: I like this change!

backend/src/middleware/hashtags/extractHashtags.spec.js
```diff
@@ -8,9 +8,24 @@ describe('extractHashtags', () => {
...
-    it('finds links with and without ".hashtag" class and extracts Hashtag names', () => {
-      const content =
-        '<p><a class="hashtag" href="/search/hashtag/Elections">#Elections</a><a href="/search/hashtag/Democracy">#Democracy</a></p>'
+    it('not `class="hashtag"` but `data-hashtag-id="something"` makes a link a hashtag link', () => {
+      const content = `
+        <p>
+          <a
+            class="hashtag"
+            data-hashtag-id="Elections"
+            href="/?hashtag=Elections"
```
Tirokk: What I don’t like at this link query in general: It doesn’t say anything about this query is to be searched for.

webapp/components/Editor/nodes/Hashtag.js
```diff
@@ -21,11 +21,15 @@ export default class Hashtag extends TipTapMention {
     return {
       ...super.schema,
       toDOM: node => {
+        // use a dummy domain because URL cannot handle relative urls
+        const url = new URL('/', 'http://example.org')
```
mattwr18: shouldn't we just use something else that can handle it then?

mattwr18: actually, looking at the docs, and what you are actually doing here, if the only thing you are using is `url.search` then you could just use our domain instead of a dummy one

backend/src/middleware/hashtags/extractHashtags.spec.js
```diff
@@ -8,9 +8,24 @@ describe('extractHashtags', () => {
   })
 
   describe('searches through links', () => {
-    it('finds links with and without ".hashtag" class and extracts Hashtag names', () => {
-      const content =
-        '<p><a class="hashtag" href="/search/hashtag/Elections">#Elections</a><a href="/search/hashtag/Democracy">#Democracy</a></p>'
+    it('not `class="hashtag"` but `data-hashtag-id="something"` makes a link a hashtag link', () => {
```
Tirokk: This sounds nice, but is not true in real.
Because the `data-hashtag-id` is deleted by saving the content in the database.
On reedit it is no link anymore in this sense.

backend/src/middleware/hashtags/hashtagsMiddleware.spec.js
```diff
@@ -129,10 +148,29 @@ describe('hashtags', () => {
         )
       })
 
-      describe('afterwards update the Post by removing a Hashtag, leaving a Hashtag and add a Hashtag', () => {
-        // The already existing Hashtag has no class at this point.
-        const postContent =
-          '<p>Hey Dude, <a class="hashtag" href="/search/hashtag/Elections">#Elections</a> should work equal for everybody!? That seems to be the only way to have equal <a href="/search/hashtag/Liberty">#Liberty</a> for everyone.</p>'
+      describe('updates the Post by removing, keeping and adding one hashtag respectively', () => {
```
Tirokk: This is a false successful test !!! 🤔
Because the `data-hashtag-id` is deleted by saving the content in the database.
On reedit it is no link anymore.
Therefore here you create a test that’s not existent in reality.
That’s why we used the `href` to search for a hashtag.

webapp/components/Editor/nodes/Hashtag.js
Tirokk: Yes @mattwr18 , I would like to see our domain here as well! 😘

backend/src/middleware/hashtags/extractHashtags.js
Tirokk: @mattwr18 This change breaks the functionality. 😂

webapp/components/Editor/nodes/Hashtag.js
Tirokk: What will be the network domain at the end?
`https://human-connection.org` is used for the donation and information page at the moment.
But I decided to change it to this. Hope it is okay for you @roschaefer !?

Tirokk:
I correct myself: The a-tag property `data-hashtag-id` gets stored in the database:

Bildschirmfoto 2019-09-17 um 10 28 52

… but does not make it into the editor again:

Bildschirmfoto 2019-09-17 um 10 31 45

roschaefer:
@Tirokk thank you for the review! This is a nasty bug with tiptap and re-rendering/parsing existing content.

2019-09-17-110850_1920x1080_scrot

The `data-hashtag-id` should be there now.

Tirokk:
This works now for me @roschaefer .
So I formal approve our PR !!! And you can merge.
PS: … if you get the build passing. Which seems unrelated …

PR #1573, opened 2019-09-13
Fix import to add createdAt and updatedAt for comments

### LINKED ISSUE #1555: 🐛 [Bug][Import] Comment dates are missing
## :bug: Bugreport

Comment dates are missing on all imported comments.

1. login into nitro-production
2. go on a imported post
3. see the comments list
4. there should be a date 

Bildschirmfoto 2019-09-12 um 12 00 03

### Expected behavior

Correct import of the comment dates.

roschaefer:
### Issues

- fix #1555 

Screenshot - 2019-09-13T140344 806

### Todo

- [X] None

### DIFF
--- a/deployment/legacy-migration/maintenance-worker/migration/neo4j/comments/comments.cql
+++ b/deployment/legacy-migration/maintenance-worker/migration/neo4j/comments/comments.cql
@@ -55,6 +55,8 @@ ON CREATE SET
 c.content        = comment.content,
 c.contentExcerpt = comment.contentExcerpt,
 c.deleted        = comment.deleted,
+c.createdAt      = comment.createdAt.`$date`,
+c.updatedAt      = comment.updatedAt.`$date`,
 c.disabled       = false
 WITH c, comment, comment.contributionId as postId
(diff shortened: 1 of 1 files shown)


deployment/legacy-migration/maintenance-worker/migration/neo4j/comments/comments.cql
```diff
@@ -55,6 +55,8 @@ ON CREATE SET
 c.content        = comment.content,
 c.contentExcerpt = comment.contentExcerpt,
 c.deleted        = comment.deleted,
+c.createdAt      = comment.createdAt.`$date`,
+c.updatedAt      = comment.updatedAt.`$date`,
```
Tirokk: I suppose you have tested it.
Therefore I approve. 💫

But where is this `$date` coming from?

mattwr18: the date would be coming from what is in the database in the `alpha`, I would guess... this is the same way we are setting the date for `Post`s

Issue #1583, opened 2019-09-14
🚀 [Feature] Update follow/unfollow mutation to return user


vbelolapotkov:
## :rocket: Feature
The feature relates to #293 . To make it right a server should return an updated user object with proper `followedByCount`

### Additional context
Use case: 

1. User A clicks 'Follow/unfollow' button in popup of the other user (User B)
2. Follow button state should change (as it is now) and followers count for User B should be updated.

Current solution drawbacks:

1. If the button is used on a user B page, then followedByCount number has glitches.

2. The current solution only makes an increase/decrease of the followedByCount by 1 which is fine for the optimistic update but might be incorrect (other users can follow/unfollow the user at the same time => this info will not appear until next page reload).

PR #1596, opened 2019-09-16
Improve follow/ufollow mutation

vbelolapotkov:
## 🍰 Pullrequest
Update follow/unfollow mutation to return followed user

### Issues
- relates #1583 

### Todo
- [x] Update schema
- [x] Update spec
- [x] Update mutation usage in a client

### DIFF
--- a/backend/src/schema/resolvers/follow.js
+++ b/backend/src/schema/resolvers/follow.js
@@ -1,51 +1,44 @@
+import { neode as getNeode } from '../../bootstrap/neo4j'
+
+const neode = getNeode()
+
 export default {
   Mutation: {
     follow: async (_object, params, context, _resolveInfo) => {
-      const { id, type } = params
+      const { id: followedId, type } = params
+      const { user: currentUser } = context
 
-      const session = context.driver.session()
-      const transactionRes = await session.run(
-        `MATCH (node {id: $id}), (user:User {id: $userId})
-          WHERE $type IN labels(node) AND NOT $id = $userId
-          MERGE (user)-[relation:FOLLOWS]->(node)
-          RETURN COUNT(relation) > 0 as isFollowed`,
-        {
-          id,
-          type,
-          userId: context.user.id,
-        },
-      )
+      if (type === 'User' && currentUser.id === followedId) {
+        return null
+      }
 
-      const [isFollowed] = transactionRes.records.map(record => {
-        return record.get('isFollowed')
-      })
-
-      session.close()
-
-      return isFollowed
+      const [user, followedNode] = await Promise.all([
+        neode.find('User', currentUser.id),
+        neode.find(type, followedId),
+      ])
+      await user.relateTo(followedNode, 'following')
+      return followedNode.toJson()
     },
 
     unfollow: async (_object, params, context, _resolveInfo) => {
-      const { id, type } = params
-      const session = context.driver.session()
+      const { id: followedId, type } = params
+      const { user: currentUser } = context
 
-      const transactionRes = await session.run(
-        `MATCH (user:User {id: $userId})-[relation:FOLLOWS]->(node {id: $id})
+      /*
+       * Note: Neode doesn't provide an easy method for retrieving or removing relationships.
+       * It's suggested to use query builder feature (https://github.com/adam-cowley/neode/issues/67)
+       * However, pure cypher query looks cleaner IMO
+       */
+      await neode.cypher(
+        `MATCH (user:User {id: $currentUser.id})-[relation:FOLLOWS]->(node {id: $followedId})
           WHERE $type IN labels(node)
           DELETE relation
           RETURN COUNT(relation) > 0 as isFollowed`,
-        {
-          id,
-          type,
-          userId: context.user.id,
-        },
+        { followedId, type, currentUser },
       )
-      const [isFollowed] = transactionRes.records.map(record => {
-        return record.get('isFollowed')
-      })
-      session.close()
 
-      return isFollowed
+      const followedNode = await neode.find(type, followedId)
+      return followedNode.toJson()
     },
   },
 }
--- a/backend/src/schema/resolvers/follow.spec.js
+++ b/backend/src/schema/resolvers/follow.spec.js
@@ -1,10 +1,17 @@
-import { GraphQLClient } from 'graphql-request'
+import { createTestClient } from 'apollo-server-testing'
 import Factory from '../../seed/factories'
-import { host, login } from '../../jest/helpers'
+import { getDriver } from '../../bootstrap/neo4j'
+import createServer from '../../server'
 
-const factory = Factory()
-let clientUser1
-let headersUser1
+let factory
+let driver
+
+let query
+let mutate
+let authenticatedUser
+
+let user1
+let user2
 
 const mutationFollowUser = id => `
   mutation {
@@ -17,20 +24,42 @@ const mutationUnfollowUser = id => `
   }
 `
 
-beforeEach(async () => {
-  await factory.create('User', {
-    id: 'u1',
-    email: 'test@example.org',
-    password: '1234',
-  })
-  await factory.create('User', {
-    id: 'u2',
-    email: 'test2@example.org',
-    password: '1234',
+beforeAll(() => {
+  factory = Factory()
+  driver = getDriver()
+
+  const { server } = createServer({
+    context: () => ({
+      driver,
+      user: authenticatedUser,
+      cypherParams: {
+        currentUserId: authenticatedUser ? authenticatedUser.id : null,
+      },
+    }),
   })
 
-  headersUser1 = await login({ email: 'test@example.org', password: '1234' })
-  clientUser1 = new GraphQLClient(host, { headers: headersUser1 })
+  const testClient = createTestClient(server)
+  query = testClient.query
+  mutate = testClient.mutate
+})
+
+beforeEach(async () => {
+  user1 = await factory
+    .create('User', {
+      id: 'u1',
+      email: 'test@example.org',
+      password: '1234',
+    })
+    .then(user => user.toJson())
+  user2 = await factory
+    .create('User', {
+      id: 'u2',
+      email: 'test2@example.org',
+      password: '1234',
+    })
+    .then(user => user.toJson())
+
+  authenticatedUser = user1
 })
 
 afterEach(async () => {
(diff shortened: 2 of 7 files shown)


vbelolapotkov:
FYI, the PR is going to be quite big to review at once. At the moment only the spec was refactored to use `apollo-server-testing`

backend/src/schema/resolvers/follow.spec.js
```diff
@@ -1,10 +1,17 @@
-import { GraphQLClient } from 'graphql-request'
+import { createTestClient } from 'apollo-server-testing'
```
roschaefer: :heart_eyes_cat: you refactor our legacy backend tests!

@aonomike :point_up:

backend/src/schema/resolvers/follow.spec.js
```diff
@@ -17,20 +24,42 @@ const mutationUnfollowUser = id => `
...
-  await factory.create('User', {
-    id: 'u1',
-    email: 'test@example.org',
-    password: '1234',
-  })
-  await factory.create('User', {
-    id: 'u2',
-    email: 'test2@example.org',
-    password: '1234',
+beforeAll(() => {
```
roschaefer: FYI @vbelolapotkov the `beforeAll` in most of our backend tests is *not only* a performance thing. There is also a bug in [graphql middlewares](https://github.com/prisma/graphql-middleware/issues/63#issuecomment-520786385) which will wrap another layer of middlewares around each resolver for every test case if you use `beforeEach`.

backend/src/schema/resolvers/follow.spec.js
```diff
@@ -41,44 +70,57 @@ describe('follow', () => {
   describe('follow user', () => {
     describe('unauthenticated follow', () => {
       it('throws authorization error', async () => {
-        const client = new GraphQLClient(host)
-        await expect(client.request(mutationFollowUser('u2'))).rejects.toThrow('Not Authorised')
+        authenticatedUser = null
+        const { errors } = await mutate({
+          mutation: mutationFollowUser('u2'),
```
roschaefer:
```suggestion
          mutation: mutationFollowUser, variables
```
Make the `mutationFollowUser` a `const` instead of a method. The string interpolation originates from a time when we weren't familiar enough with GraphQL and we did not know how to use `variables`.

The pattern which I happen to love is to mutate a global `variables` object in a `beforeEach` of a scope. Then pass these `variables` into the call of `mutate` or `query`. In my opinion that's a good way to avoid false negatives.

backend/src/schema/resolvers/follow.spec.js
```diff
@@ -41,44 +70,57 @@ describe('follow', () => {
...
-      const res = await clientUser1.request(mutationFollowUser('u2'))
-      const expected = {
-        follow: true,
-      }
-      expect(res).toMatchObject(expected)
+      const { data: result } = await mutate({
+        mutation: mutationFollowUser(user2.id),
+      })
+
+      const expected = { follow: true }
```
roschaefer:
```suggestion
      const expected = { name: 'Name of the followed user', followedBy: [ { name: 'Me' } ], followedByCurrentUser: true }
```

backend/src/schema/resolvers/follow.spec.js
```diff
@@ -41,44 +70,57 @@ describe('follow', () => {
...
+      expect(result).toMatchObject(expected)
+
+      const {
+        data: { User },
+      } = await query({
```
roschaefer: Remove this whole workaround.

backend/src/schema/resolvers/follow.spec.js
```diff
@@ -41,44 +70,57 @@ describe('follow', () => {
...
     it('I can`t follow myself', async () => {
-      const res = await clientUser1.request(mutationFollowUser('u1'))
-      const expected = {
-        follow: false,
-      }
-      expect(res).toMatchObject(expected)
+      const { data: result } = await mutate({
+        mutation: mutationFollowUser(user1.id),
+      })
+      const expected = { follow: false }
```
roschaefer:
```suggestion
      const expected = { follow: null }
```

I think it's good practice to respond with `null` when nothing has been mutated in any way (e.g. nothing has been deleted).

backend/src/schema/resolvers/follow.spec.js
```diff
@@ -41,44 +70,57 @@ describe('follow', () => {
...
+        data: { User },
+      } = await query({
+        query: `{
+        User(id: "${user1.id}") {
           followedBy { id }
           followedByCurrentUser
         }
-      }`)
+      }`,
+      })
```
roschaefer: Same, remove, as described :point_up:

backend/src/schema/resolvers/follow.spec.js
```diff
@@ -87,32 +129,39 @@ describe('follow', () => {
     })
   })
   describe('unfollow user', () => {
+    beforeEach(async () => {
+      await mutate({ mutation: mutationFollowUser(user2.id) })
```
roschaefer: You can either setup the data through a mutation or you can do:

```js
await myUser.relateTo(anotherUser, 'following')
```

If `myUser` and `anotherUser` is a neode node (that's the return type of the factory method).

vbelolapotkov:
@roschaefer thanks for review and suggestions. Will continue working on improvements.

backend/src/schema/resolvers/follow.spec.js
vbelolapotkov: @roschaefer I wouldn't agree on this. Suggested improvement manipulates DB "directly" rather than through known server API, but we test API and we shouldn't reason about the DB or any other tools used by the server.

In the future relation name, `neode` model or some other properties can change which will break the test setup. So we'll have to get back and fix the test.

If we keep the current approach the test setup will break only if we change mutation definition. However, if this happens we'll have to rewrite all the test anyways.

backend/src/schema/resolvers/follow.spec.js
vbelolapotkov: Good point, thanks for the info

backend/src/schema/resolvers/follow.spec.js
vbelolapotkov: @roschaefer good suggestion. Definitely will use it with further steps :)

backend/src/schema/resolvers/follow.spec.js
vbelolapotkov: @roschaefer probably idea is good however this change requires schema update because of the mutation result is defined as `Boolean!` value. As we're about to change the schema significantly with further steps this change doesn't bring a lot of value IMO at the moment.

backend/src/schema/resolvers/follow.spec.js
vbelolapotkov: done

backend/src/schema/resolvers/follow.js
```diff
@@ -1,51 +1,44 @@
...
+      const { id: followedId, type } = params
+      const { user: currentUser } = context
 
-      const transactionRes = await session.run(
-        `MATCH (user:User {id: $userId})-[relation:FOLLOWS]->(node {id: $id})
+      /*
+       * Note: Neode doesn't provide an easy method for retrieving or removing relationships.
+       * It's suggested to use query builder feature (https://github.com/adam-cowley/neode/issues/67)
+       * However, pure cypher query looks cleaner IMO
+       */
```
roschaefer: Agree!

roschaefer: I haven't made a final decision about it, ie. if we should use it for production code or if we should only use it for our factories. :thinking: :woman_shrugging:

backend/src/schema/types/schema.gql
```diff
@@ -32,9 +32,9 @@ type Mutation {
   # Unshout the given Type and ID
   unshout(id: ID!, type: ShoutTypeEnum): Boolean!
   # Follow the given Type and ID
-  follow(id: ID!, type: FollowTypeEnum): Boolean!
+  follow(id: ID!, type: FollowTypeEnum): User
```
roschaefer: :+1: nice, looks great!

If you need any help or another review, feel free to ask for it @vbelolapotkov.

backend/src/schema/resolvers/follow.spec.js
```diff
@@ -40,84 +97,73 @@ afterEach(async () => {
...
     describe('unauthenticated follow', () => {
-      it('throws authorization error', async () => {
-        // follow
-        await clientUser1.request(mutationFollowUser('u2'))
-        // unfollow
-        const client = new GraphQLClient(host)
-        await expect(client.request(mutationUnfollowUser('u2'))).rejects.toThrow('Not Authorised')
+      test('throws authorization error', async () => {
+        authenticatedUser = null
+        const { errors, data } = await mutate({ mutation: mutationUnfollowUser, variables })
```
roschaefer: You could also do this:
```js
await expect(mutate({ mutation: mutationUnfollowUser, variables })).resolves.toMatchObject({
  errors: [{'message': 'Not Authorised!'}],
  data: { unfollow: null },
})

Comment (vbelolapotkov) [reply backend/src/schema/resolvers/follow.spec.js]: Thanks, I'm aware of it but I personally prefer separating a call for the method under test and assertions for the result. However, I'm fine to change it if there is a pattern in the codebase, 
@roschaefer please, let me know if this is a case.

File: backend/src/schema/types/schema.gql (hunk 10 above)
Comment (vbelolapotkov) [reply backend/src/schema/types/schema.gql]: @roschaefer actually this looks a little be weird, especially `FollowTypeEnum` argument. I suppose this was done to reuse the mutation in the future to allow following other entities like posts or something else.

In that case, I'd expect some new `Followable` type (e.g. union type) or interface is returned. What do you think?

**Comment by vbelolapotkov** [comment]
@roschaefer this one is ready for review

File: backend/src/schema/resolvers/follow.spec.js (hunk 11 above)
Comment (roschaefer) [reply backend/src/schema/resolvers/follow.spec.js]: It's the pattern I stick to right now, but totally up to you! I thought you would be interested to see this.

Comment (roschaefer) [reply backend/src/schema/resolvers/follow.spec.js]: I just fond `toMatchObject` quite robust for testing a graphql response with very helpful error messages.

File: webapp/pages/profile/_id/_slug.vue [hunk 12]
```diff
@@ -402,6 +403,27 @@ export default {
...
+        this.user.followedBy = this.user.followedBy.filter(user => user.id !== currentUser.id)
+      }
+      this.user.followedByCurrentUser = followedByCurrentUser
+    },
+    updateFollow({ followedByCurrentUser, followedBy, followedByCount }) {
+      this.followedByCountStartValue = this.user.followedByCount
+      this.user.followedByCount = followedByCount
+      this.user.followedByCurrentUser = followedByCurrentUser
+      this.user.followedBy = followedBy
+    },
```
roschaefer: Wow, this is really sophisticated. I would have been fine with updating the user the pessimistic way™. This is really advanced and snappy.

backend/src/schema/types/schema.gql
roschaefer: This makes no sense at all. I guess the `FollowTypeEnum` was supposed to include `Organization`s as well. Something that we decided to not implement any soon. @vbelolapotkov do you want to land another quick contribution and remove `FollowTypeEnum`?

vbelolapotkov: @roschaefer will be happy to help here as well. I'd rename the mutation as well to `(un)followUser`

vbelolapotkov: @roschaefer here it is #1660 , please review when you have time

PR #1599, opened 2019-09-16
Configure emails

### LINKED ISSUE #1186: 🚀 [Feature] Make Invite an Registration E-Mails translatable and pretty
## :rocket: Feature

At the moment the e-mails are not very nice and only in english.
Therefore we like to have the ability to make them translatable and pretty.

alina-beck:
## 🍰 Pullrequest
Make emails translatable and pretty. :grin:

Behold the email magic:
ezgif com-optimize(2)

Here are the actual email files – ready to download and open in the email program of your choice:
https://send.firefox.com/download/80902f5af59aa4f7/#qp3JcGkmEu2pR-omPwqzhw

### Issues
- fixes #1186

### ToDo
- [x] once approved: apply design to `password reset` and `wrong account` emails
- [ ] if anybody could open the emails with `Outlook` on `Windows` it would help me sleep better (it's the only email client I'm worrying about... 😝)


backend/src/middleware/email/templates/passwordReset.js
```diff
@@ -53,6 +54,7 @@ export const wrongAccountMail = options => {
   } = options
   const actionUrl = new URL('/password-reset/request', CONFIG.CLIENT_URI)
   return {
+    from: '"Human Connection" <info@human-connection.org>',
```
roschaefer: We already forgot to set this. Can we store this somewhere at a central location, so we never forget to add it? Seeing it twice in the same file makes me nervous.

roschaefer: Well, it feels more natural to have it there. Maybe it should be the task of QR to not have it slip through. :thinking:

alina-beck:
@roschaefer so many good questions... :smile:

1. Martin and Matt were in favor of including both English and German for now. Later I would love if the admin could choose a language when inviting people – using the browser language wouldn't make much sense because an English-speaking admin could also invite German-speaking users... It's different when users sign up themselves, though! Anyway, the toggle is just a temporary solution to avoid displaying the email twice.

2. I was wondering where the `nonce` should go! I'll remove the sentence from the email along with the obsolete code.

3. The hard work is done already – using the design for the other two email templates is a matter of an hour, maybe. So please wait with merging! I'll finish it tomorrow.

backend/src/middleware/email/templates/passwordReset.js
alina-beck: I also felt it made more sense to place the `from` alongside the other attributes. But I'm hoping to refactor the passwordReset in such a way that I can get rid of the duplicate!

roschaefer:
@alina-beck I think it's perfectly fine to choose the current browser language of the admin who invites the user as a first language and English as a fallback. The admin probably knows the language of the invited user, because invited user has donated from a French bank account or contacted us with an E-Mail in Swahili. So the admin could choose the desired language from the language chooser when sending invite mails. This is probably quite counter-intuitive but since it's an admin, we can expect a higher level of familiarity with the application.

For now, I'm absolutely OK *not* to implement this feature, because of [KISS](https://en.wikipedia.org/wiki/KISS_principle). Also we would have to worry about the combination English - English (fallback).

mattwr18:
hey @Tirokk, I like some of your suggestions, but some of them I think are not necessary, not sure how @alina-beck feels about them. 

Pointing out the no space before an ellipsis I think is an unnecessary change, both ways are valid. 
I can understand that you prefer the space before, and have added several places in the code base this way, so maybe it's better to keep it uniform all around. 
I personally don't like the space before, but I don't feel strong enough about it to request that `PR` creators change it.

Also, with email, both email and e-mail are legit, and the `PR` creator can simply choose this, and not have it pointed out to them that I, or you, prefer it the other way. 

I do really like that you pointed out the link to copy and paste is not obvious and this text is missing in German, so maybe these are the only changes you see blocking this `PR`?

The others certainly shouldn't slow us down from inviting new users. I'm sure you know this already, but Thursdays and Fridays are @alina-beck days away from `human-connection`

Tirokk:
@alina-beck wrote on discord she will care for things before noon today.

I pointed out the space before ellipse, because we found out it’s the way it should be.
This is for sure no reason to hold back the PR and I make aware of it and hold the PR back, because there are other things more important, like you said. @mattwr18 

If you are in a hurry to merge this in then go ahead …

datenbrei:
Hi Alina. Great work. One small change. Everywhere else we was Writing "Du" instead of "du", so in uppercase. I very much apprecheate your work.

German and English in one was a decision we (all) made a few days ago for now, later we can change this. And also extend to further languages.

mattwr18:
The thing is that it is not the way "it should be", both are grammatically correct, it's a preference.

I think you know that I am kind of never in a hurry to merge in `PR`s, and as I stated, I really like your suggestion to add the link to copy and paste, or even remove that text.

I saw that she said she can make changes today before 12, not sure what time zone she meant, and also thinking that it's nice if we can take care of it, if she wants.
I know you can do things on Tuesday, and often do, but prefer not to.

Tirokk:
> The thing is that it is not the way "it should be", both are grammatically correct, it's a preference.

@mattwr18 I don’t force anybody to do something. It is just a suggestion and make someone aware of a questionable thing. Nothing else.
You looked it up at that time and stated to me that you found it should be the way we have changed it at that time. If I have misunderstood than I apologise. Than it’s a preference … 🤗

roschaefer:
Hey everyone, if @alina-beck has her time-off, why don't we implement the changes ourselves (as long as @alina-beck ) does not mind?

@Tirokk I think we do not disagree. I would like to have the mother tongue as the first language and English as a fallback (2nd).

@datenbrei a teacher for journalism has told me that it is best practice to use "du", not "Du". The upcase version is old-fashioned.

alina-beck:
Thank you all for your great input! 😍 

You're right @mattwr18, many of the things pointed out were about personal preference (and I followed mine when creating this PR) – but it's also about being consistent. So if we are using a space before … and `Du` instead of `du` in other places at Human Connection, especially user facing communication, it's probably a good idea to stick to it. That's why I did follow your suggestions @datenbrei @Tirokk. Thanks for pointing these things out to me – I'll try to remember ... :wink:

alina-beck:
Also @Tirokk, good catch with the code! That should indeed have been removed. And it is now. :smile:

Tirokk:
> @Tirokk I think we do not disagree. I would like to have the mother tongue as the first language and English as a fallback (2nd).

@roschaefer nice that you like this idea as well.

mattwr18:
Great @alina-beck thanks for being patient with us and making the changes, sorry if I let my personal preferences come out, and make it more complicated...

PR #1613, opened 2019-09-17
feat(deployment): Add helm charts for deploy

### LINKED ISSUE #1629: 🚀 [Feature] Refactor our kubernetes configuration into a helm chart
## :rocket: Feature

[Helm](https://helm.sh/) is a package manager for kubernetes. Refactoring our [current instructions](https://docs.human-connection.org/human-connection/deployment) would make it a lot easier to deploy our application:

* One click install
* Conditional deployment of mailserver
* Easy switch to maintenance mode
* Reliable rollouts by incrementing one docker image tag

roschaefer:
### Issues

- fixes #1629 

### Todo

Taking from notes in our Daily Standup 2020-01-26, @roschaefer reports, we need to consider these things.
- [ ] helm chart volumes

- ~I spent a big part of today trying to get this worked out and it didn't in the end. I was able to create `PersistentVolumeClaims`, but when they created `PersistentVolumes` in DigitalOcean, it messed up the `neo4j` deployment, so that it could not create the needed `data/databases` directory, etc.~

- ~The issue is known and explained here https://www.digitalocean.com/docs/kubernetes/how-to/add-volumes/#setting-permissions-on-volumes ... I tried to get this working, and in the end it did get passed the originally error, but then I couldn't restore the neo4j data! it would load successfully, but then when the neo4j pod restarted, the data would dissapear :frowning_face:~

- ~In the end, it was much easier to set up the `PersistenVolumeClaims` with `kubectl` ...~
- If you have read the above, it turns out that the issue was most likely due to existing PV and PVC, I was able to get it to work with `Helm`, and this is preferable cause not only does it mean it's easier to set up, but Helm manages making sure the pods that have a PVC get assigned correctly.
I dealt with issues when setting up the PVC with `kubectl` and everything else with `Helm`
- [ ]  admin-user
- please give a use case for this @roschaefer ... in Kubernetes >= 1.16 RBAC(role-based access control) is enabled by default. See, https://helm.sh/docs/topics/rbac/
- @roschaefer says this is more about database admin, oder?
- [ ]  categories

- I need clarification on what this means.
- @roschaefer clarifies that we would like a way for other people who want to host the software to be able to adapt it to their needs, so one of those would be the ability to define Categories to suit their needs.
We also agreed this could be done in a/several separate `PR`s.
We should open an `EPIC` for this, with sub-issues moving us closer to this goal.
 
- [x]  services, ie cert-manager, ingress, maintenance
- [ ]  upgrade
- what do you mean by this @roschaefer ? clarify please, we can use the `helm upgrade` command to update our version, for example, or scale. It's not something we need to set up, we need to document it, for sure.
- [x]  hooks, ie migration

- done, but a doubt whether we really want to run `migrate init` every time we deploy? if not, should we just run it manually the first time, then again anytime there are new indexes/constraints??

after meeting with @roschaefer, we talked about the reasons to favor a `Job` that handles the database migrations, for example, when/if we scale the backend, we do not want every pod to run the migrations.


mattwr18:
hey @drone-cloud, just saw this now... we would really like to use node affinity for this purpose, yes
I attempted to do this following the docs and it didn't really work the way I thought it would. I added this:
```
nodeSelector:
  node: neo4j
```
to our `nitro-neo4j` deployment and added a label `node=neo4j` to our node `pool-gxch7n9pm-r1i0`
which seems like it worked in the sense that `ntiro-neo4j` is on that node, but so is every other pod
```
maintenance-5dd465cb6b-wkt55     1/1     Running   0          22d   10.244.3.124   pool-gxch7n9pm-r1i0   <none>           <none>
nitro-backend-5d687d5f85-8sxmv   1/1     Running   0          5d    10.244.3.203   pool-gxch7n9pm-r1i0   <none>           <none>
nitro-neo4j-76d8654997-l84t9     1/1     Running   0          5d    10.244.3.102   pool-gxch7n9pm-r1i0   <none>           <none>
nitro-web-856545b495-qf2zs       1/1     Running   0          5d    10.244.3.71    pool-gxch7n9pm-r1i0   <none>           <none>
nitro-web-856545b495-z54cx       1/1     Running   0          5d    10.244.3.131   pool-gxch7n9pm-r1i0   <none>           <none>
```

I couldn't find anything about restricting pods from assigning themselves to certain nodes, or if that is even a good practice.

I don't know about using local values for every pod, maybe @roschaefer has a comment about that...

KapilJ22:
> hey @drone-cloud, just saw this now... we would really like to use node affinity for this purpose, yes
> I attempted to do this following the docs and it didn't really work the way I thought it would. I added this:
> 
> ```
> nodeSelector:
>   node: neo4j
> ```
> 
> to our `nitro-neo4j` deployment and added a label `node=neo4j` to our node `pool-gxch7n9pm-r1i0`
> which seems like it worked in the sense that `ntiro-neo4j` is on that node, but so is every other pod
> 
> ```
> maintenance-5dd465cb6b-wkt55     1/1     Running   0          22d   10.244.3.124   pool-gxch7n9pm-r1i0   <none>           <none>
> nitro-backend-5d687d5f85-8sxmv   1/1     Running   0          5d    10.244.3.203   pool-gxch7n9pm-r1i0   <none>           <none>
> nitro-neo4j-76d8654997-l84t9     1/1     Running   0          5d    10.244.3.102   pool-gxch7n9pm-r1i0   <none>           <none>
> nitro-web-856545b495-qf2zs       1/1     Running   0          5d    10.244.3.71    pool-gxch7n9pm-r1i0   <none>           <none>
> nitro-web-856545b495-z54cx       1/1     Running   0          5d    10.244.3.131   pool-gxch7n9pm-r1i0   <none>           <none>
> ```
> 
> I couldn't find anything about restricting pods from assigning themselves to certain nodes, or if that is even a good practice.
> 


@mattwr18 - Sorry for the late reply. I saw your reply today only. Why do you need to restrict pods from assigning themselves to certain nodes? If there is a specific reason for this, then maybe we can assign a new label to other nodes e.g. "non-neo4j" and have all other pods use this as `nodeSelector` and use `preferredDuringSchedulingIgnoredDuringExecution` as `nodeAffinity`.

Also, did you try restarting those pods a few times to see if they are in fact assigned to different nodes?

mattwr18:
I think we very much still want this stale bot... just need some time to finish it

deployment/helm/human-connection/templates/service-backend.yaml
```diff
@@ -2,6 +2,7 @@ apiVersion: v1
 kind: Service
 metadata:
   name:  {{ .Release.Name }}-backend
+  namespace: human-connection
```
roschaefer: Are you sure? Consider that you could run two different instances of human connection in the same kubernetes cluster in two different namespaces. As far as I'm concerned, this is the reason why setting the namespace is considered bad practice in helm charts. You can set it when you install the chart.

deployment/helm/human-connection/values.yaml
```diff
@@ -1,31 +1,27 @@
...
 pullPolicy: Always
 
 # configs
-developmentMailserverDomain: "helm-nitro-mailserver.human-connection.org"
-humanConnectionDomain: "helm-staging.human-connection.org"
-supportEmail: "developer@human-connection.org"
-letsencryptIssuer: "letsencrypt-staging"
+developmentMailserverDomain: "mailserver.human-connection.org"
+humanConnectionDomain: "develop.human-connection.org"
+supportEmail: "devcom@human-connection.org"
```
roschaefer: Do we have that email? :thinking: Or just to avoid spam?

deployment/helm/human-connection/templates/service-backend.yaml
mattwr18: You are right! I still have some cleanup to do, thanks for pointing this out!

deployment/helm/human-connection/values.yaml
mattwr18: this is the email that @datenbrei asked to add, but it goes back to the point I made in `Discord` that if we are going to publish our `helm` chart publicly, it would be nice to get set up with `helm-secrets` and encrypt the files.

deployment/helm/human-connection/Chart.yaml
```diff
@@ -1,5 +1,5 @@
 apiVersion: v1
-appVersion: "0.1.0"
+appVersion: "0.2.2"
```
roschaefer:
```suggestion
appVersion: "0.3.0"
```
Now. We should keep that in sync somehow. Maybe with `yarn run release`?

deployment/README.md
```diff
@@ -1,11 +1,10 @@
 # Human-Connection Nitro \| Deployment Configuration
 
-We deploy with [kubernetes](https://kubernetes.io/). In order to deploy your own
-network you have to [install kubectl](https://kubernetes.io/docs/tasks/tools/install-kubectl/)
-and get a kubernetes cluster.
+There are a couple different ways we have tested to deploy an instance of Human Connection, with [kubernetes](https://kubernetes.io/) and via [Helm](https://helm.sh/docs/). In order to manage your own
```
roschaefer: I would even argue that we do not support plain kubernetes configurations once we're familiar enough with helm.

deployment/helm/human-connection/README.md
```diff
@@ -0,0 +1,72 @@
...
+$ echo -n 'admin' | base64
+YWRtaW4=
+```
+Those secrets get `base64` decoded and are available as environment variables in
+your deployed kubernetes pods.
+
+# https
+If you start with setting up the `https`, when you install the app, it will automatically take care of the certificates for you.
+
+First check that you are using `Helm v3`, this is important since it removes the need for `Tiller`. See, [FAQ](https://helm.sh/docs/faq/#removal-of-tiller)
```
roschaefer: Good to know! :+1:

deployment/helm/human-connection/README.md
```diff
@@ -0,0 +1,72 @@
...
+$ helm install cert-manager --namespace cert-manager --version v0.13.0 jetstack/cert-manager
+```
+
+# Deploy
+
+Once you are satisfied with the configuration, you can install the app.
+
+```bash
+# in folder /deployment/helm
+$ helm install develop ./ --namespace human-connection
```
roschaefer: Excellent!

deployment/helm/human-connection/templates/deployments/deployment-neo4j.yaml
```diff
@@ -33,15 +38,6 @@ spec:
             memory: {{ .Values.neo4jResourceRequestsMemory | default "1G" | quote }}
           limits:
             memory: {{ .Values.neo4jResourceLimitsMemory | default "1G" | quote }}
-        env:
-        - name: NEO4J_dbms_security_procedures_unrestricted
-          value: "algo.*,apoc.*"
```
roschaefer: Obsolete? We don't need that anymore?

mattwr18: we do need it actually! I ran into an issue with running the `yarn prod:migrate init` yesterday on the `develop` server because this was removed... I'll add it back.

I did have a question about whether we should have `production` config stored locally and `.gitignored`, or how we should handle that? I would like to have a copy of what is being deployed in production in case there is a need to recreate the cluster for some reason.

I have looked into `helm-secrets` and got it installed on my computer, but when I tried to use it, it threw errors. Before debugging, I though it would be best to come to some consensus on it.

deployment/README.md
mattwr18: yeah, I hear ya! One more dependency that needs to be updated/tested

deployment/helm/human-connection/Chart.yaml
mattwr18: I could try that out... actually, we had some complaints that the version number is not correct in production ... I think I didn't run it properly

deployment/helm/human-connection/templates/deployments/deployment-neo4j.yaml
mattwr18: Actually, I hadn't realized that you are `.gitignore`ing the `configmap.yaml` and `secrets.yaml` ... I have extracted all the `env` variables there.
My suggestion would be for myself and you to have different files like `develop.configmap.yaml`/`production.configmap.yaml` or different `values.yaml` and `.gitignore` them.

deployment/volumes/volume-snapshots/snapshot.yaml
```diff
@@ -2,9 +2,9 @@
   apiVersion: snapshot.storage.k8s.io/v1alpha1
   kind: VolumeSnapshot
   metadata:
-    name: neo4j-data-snapshot
+    name: uploads-snapshot
     namespace: human-connection
   spec:
     source:
-      name: neo4j-data-claim
```
roschaefer: What happened to the neo4j snapshot? Anyways, as far as I remember this was supposed to be edited by the developer.

roschaefer:
@mattwr18 I cannot I approve my own PR but consider this as an approval.

PR #1614, opened 2019-09-17
Refactor Embeds to use custom components

mattwr18:
### Todo

- [X] None


backend/src/schema/resolvers/embeds.js
```diff
@@ -11,16 +11,17 @@ export default {
...
       'publisher',
-      'date',
-      'description',
       'url',
+      'date',
       'image',
       'audio',
       'video',
       'lang',
-      'html',
```
roschaefer: Hm, I think for the trusted oembed providers in `providers.json` we can return plain html to the client.

webapp/components/Embeds/VideoEmbed.vue
```diff
@@ -0,0 +1,110 @@
...
+    },
+  },
+  data() {
+    return {
+      renderEmbed: false,
+    }
+  },
+  computed: {
+    embedUrl() {
+      if (this.embedData.publisher === 'YouTube') {
```
roschaefer: No, I don't want to see this kind of complexity. I would prefer to stick to the current solution for more complex embeds. 

 For security, have a look if we can trust the oembed provider. For data privacy, wrap the embed with an additional click (are you OK to load stuff from 3rd parties?). For mobile responsiveness, try out if the iframe or whatever breaks the layout.

webapp/components/Editor/nodes/Embed.spec.js
```diff
@@ -9,23 +9,17 @@ describe('Embed.vue', () => {
   beforeEach(() => {
```
roschaefer: I love that you created test cases for the embeds :heart:

mattwr18:
closing this in favor of https://github.com/Human-Connection/Human-Connection/pull/1686
we could maybe still refactor in the future to have separate components based on what is returned from `oEmbed`...

PR #1617, opened 2019-09-17
Resend registration mail if user has not yet registered

### LINKED ISSUE #1616: 🐛 [Bug] Re-send registration mail if user has not yet signed up
## :bug: Bugreport

If an admin has invited a user but this user has not registered yet or the email was lost - it won't be possible to resend the registration mail. The error shown looks like this:

Screenshot - 2019-09-17T163729 658

1. Log in as `admin@example.org` at nitro-staging.human-connection.org
2. Visit https://nitro-staging.human-connection.org/admin/invite
3. Invite `test@example.org`
4. Refresh the page
5. Try to invite the same email again
6. See the error page above

### Expected behavior
For not-registered users it would be great to resend the registration email.

roschaefer:
### Issues

- fix #1616

### Todo

- [X] None

### DIFF
--- a/backend/src/seed/factories/users.js
+++ b/backend/src/seed/factories/users.js
@@ -24,7 +24,15 @@ export default function create() {
       }
       args = await encryptPassword(args)
       const user = await neodeInstance.create('User', args)
-      const email = await factoryInstance.create('EmailAddress', { email: args.email })
+
+      let email
+      if (typeof args.email === 'object') {
+        // probably a neode node
+        email = args.email
+      } else {
+        email = await factoryInstance.create('EmailAddress', { email: args.email })
+      }
+
       await user.relateTo(email, 'primaryEmail')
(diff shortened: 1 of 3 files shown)


backend/src/seed/factories/users.js
```diff
@@ -24,7 +24,15 @@ export default function create() {
       }
       args = await encryptPassword(args)
       const user = await neodeInstance.create('User', args)
-      const email = await factoryInstance.create('EmailAddress', { email: args.email })
+
+      let email
+      if (typeof args.email === 'object') {
+        // probably a neode node
```
mattwr18: hahah probably??

roschaefer:
@mattwr18 yes, one has to test manually that the email gets send. We *could* test the email middleware automatically maybe. By bubbling up a fake email response and mocking out nodemailer.

PR #1640, opened 2019-09-18
1486 make slug more usable

### LINKED ISSUE #1486: 🚀 [Feature] Make the slug more visible and usable
## :rocket: Feature

Make the slug visible for to be referenced in comments. This could be done in a way that the slug is following the visible username or by a hover and a popup-menu, so that it is better looking.

### Design & Layout

See here, you just see the username but not the slug behind it: 
username

But when commenting just the slug is shown:
slug

### Additional context

The usernames (John Doe) are not unique, but the slugs (john-doe) are. So for referencing a user in a comment the user id must be choosen right. This can just be done, if one can see the slug .behind a username.

roschaefer:
## 🍰 Pullrequest

Screenshot - 2019-09-18T171336 328

Design Team, I need your help: Can you tell me if the datetime next to the user's slug is a problem?

### Issues

- fix #1486 

### Todo

- [X] None

### DIFF
--- a/webapp/components/User/User.vue
+++ b/webapp/components/User/User.vue
@@ -11,17 +11,19 @@
         <div @mouseover="openMenu(true)" @mouseleave="closeMenu(true)">
           <hc-avatar class="avatar" :user="user" />
           <div>
-            <b class="username">{{ userName | truncate(18) }}</b>
+            <ds-text align="left">
+              <b class="username">{{ userName | truncate(18) }}</b>
+              <ds-text v-if="dateTime" size="small" color="soft">
+                <ds-icon name="clock" />
+                <client-only>
+                  <hc-relative-date-time :date-time="dateTime" />
+                </client-only>
+              </ds-text>
+            </ds-text>
           </div>
           <!-- Time -->
-          <ds-text align="right" size="small" color="soft">
-            <span class="slug">{{ userSlug }}</span>
-            <div v-if="dateTime" style="display: inline;">
-              <ds-icon name="clock" />
-              <client-only>
-                <hc-relative-date-time :date-time="dateTime" />
-              </client-only>
-            </div>
+          <ds-text align="left" size="small" color="soft">
+            {{ userSlug }}
           </ds-text>
         </div>
       </nuxt-link>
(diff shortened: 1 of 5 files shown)


alina-beck:
Hmm... Have you tried putting the slug right behind the user name? I don't _love_ the current solution on the posts but I don't really see a quick and easy way to make it prettier, so maybe just go with it for now.

I would make the font a bit smaller on the user profile (top left in your screenshot), though.

webapp/components/User/User.vue
```diff
@@ -14,14 +14,15 @@
             <b class="username">{{ userName | truncate(18) }}</b>
           </div>
           <!-- Time -->
-          <div v-if="dateTime" style="display: inline;">
-            <ds-text align="left" size="small" color="soft">
+          <ds-text align="right" size="small" color="soft">
+            <span class="slug">{{ userSlug }}</span>
+            <div v-if="dateTime" style="display: inline;">
```
alina-beck: If you use a `<span>` here instead of the `<div>` there's no need for the `inline` style. It also makes more sense semantically. ;)

webapp/pages/profile/_id/_slug.vue
```diff
@@ -27,7 +27,12 @@
             />
           </client-only>
           <ds-space margin="small">
-            <ds-heading tag="h3" align="center" no-margin>{{ userName }}</ds-heading>
+            <ds-heading tag="h3" align="center" no-margin>
+              {{ userName }}
+            </ds-heading>
+            <ds-text size="large" align="center" color="soft">
```
alina-beck: Already mentioned it – I would prefer a smaller text size here

appinteractive:
Hey You May want to talke a look at my take on it: https://dist.appinteractive.now.sh/ I do it more or less like Twitter.

roschaefer:
Screenshot - 2019-09-18T232900 878


This is how it looks like. @appinteractive I like your styling but I think you don't cover the case that the date is empty, do you? So that's why I went with @alina-beck's suggestion now and moved the date next to the user name. It might happen that the date overflows the post card width on long user names, though.

alina-beck:
I actually meant it the way @appinteractive suggested it in the design :laughing: – but you're right, it could look weird with empty dates. How often do we have this case?

Anyway, I can live with both. You decide!

roschaefer:
Oh, @alina-beck you're right! Sorry for not reading your suggestions carefullly.

The empty date happens on the user profile in the list of followed/unfollowed users. See picture 1 on the left side.

PR #1649, opened 2019-09-19
Fix bug where short comments scrub links

mattwr18:
## 🍰 Pullrequest

the comments are still strange cause they have a height of like `200px` 
maybe @alina-beck can help :pray: 
or I can keep playing...


mattwr18:
thanks @Tirokk 
your keen eye has inspired me to improve it more in https://github.com/Human-Connection/Human-Connection/pull/1667
please have a look, I think you'll like it

PR #1651, opened 2019-09-19
Change your own slug

### LINKED ISSUE #1650: 🚀 [Feature] Change slug
## :rocket: Feature

A user has reported this on the alpha:
> For privacy reasons ... (she once said) reduced her last name in the nick - but the user ID ignores that. 

So people changed their unique slug for privacy reasons, when they originally signed up with their full name. We should fix that.

### Design & Layout

Screenshot - 2019-09-19T202740 758

roschaefer:
## 🍰 Pullrequest

This adds a feature to change your own slug on users settings.

### Issues

- fix #1650 

### Todo

- [ ] make sure the provided slug is html safe and lowercased

### DIFF
--- a/backend/src/models/User.js
+++ b/backend/src/models/User.js
@@ -4,7 +4,7 @@ module.exports = {
   id: { type: 'string', primary: true, default: uuid }, // TODO: should be type: 'uuid' but simplified for our tests
   actorId: { type: 'string', allow: [null] },
   name: { type: 'string', disallow: [null], min: 3 },
-  slug: 'string',
+  slug: { type: 'string', regex: /^[a-z0-9_-]+$/, lowercase: true },
   encryptedPassword: 'string',
   avatar: { type: 'string', allow: [null] },
   coverImg: { type: 'string', allow: [null] },
--- a/webapp/components/utils/UniqueSlugForm.spec.js
+++ b/webapp/components/utils/UniqueSlugForm.spec.js
@@ -1,0 +1,80 @@
+import UniqueSlugForm from './UniqueSlugForm'
+import Schema from 'async-validator'
+
+let translate
+let apollo
+let currentUser
+
+beforeEach(() => {
+  translate = jest.fn(() => 'Validation error')
+  apollo = {
+    query: jest.fn().mockResolvedValue({ data: { User: [] } }),
+  }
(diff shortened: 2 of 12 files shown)


roschaefer:
@mattwr18 this time you were a little too quick :wink: I would like to test this and I want to make this available on the create user account page.

mattwr18:
hehe... yeah, too fast, but I saw that it's passing the build and works :stuck_out_tongue_closed_eyes:

roschaefer:
Please don't merge. I saw that the backend does not ensure that the `slug` is html safe and lowercased when passed via `args`.

webapp/components/utils/UniqueSlugForm.spec.js
```diff
@@ -0,0 +1,80 @@
...
+  describe('regex', () => {
+    describe('non URL-safe characters, e.g. whitespaces', () => {
+      it('rejects', async () => {
+        await expect(validate({ slug: 'uh oh' })).rejects.toThrow('Validation error')
+      })
+    })
+
+    describe('alphanumeric, hyphens or underscores', () => {
+      it('validates', async () => {
+        await expect(validate({ slug: '_all-right_' })).resolves.toBeUndefined()
```
mattwr18: this is confusing for me, it resolves, but to undefined?
can't it return a boolean? or it works like if it fails the validations there is a message to display and if not, there is no message?

roschaefer: Yes, it's the specific implementation of [async-validator](https://github.com/yiminghe/async-validator) where you call a callback on validation error and do nothing otherwise. Thus, no return value.

backend/src/models/User.js
```diff
@@ -4,7 +4,7 @@ module.exports = {
   id: { type: 'string', primary: true, default: uuid }, // TODO: should be type: 'uuid' but simplified for our tests
   actorId: { type: 'string', allow: [null] },
   name: { type: 'string', disallow: [null], min: 3 },
-  slug: 'string',
+  slug: { type: 'string', regex: /^[a-z0-9_-]+$/, lowercase: true },
```
mattwr18: we could maybe add `unique: true` here as well, but it can be a refactor, for sure

PR #1657, opened 2019-09-20
[BACKEND ONLY] save user confirmation for permanent embedding of iframe

ogerly:
## 🍰 Pullrequest
save user confirmation for permanent embedding of iframe

## allowEmbedIframes 

created in the database

### DIFF
--- a/backend/src/models/User.js
+++ b/backend/src/models/User.js
@@ -104,4 +104,8 @@ module.exports = {
     target: 'Location',
     direction: 'out',
   },
+  allowEmbedIframes: {
+    type: 'boolean',
+    default: false,
+  },
 }
--- a/backend/src/schema/resolvers/users.js
+++ b/backend/src/schema/resolvers/users.js
@@ -175,6 +175,7 @@ export default {
         'about',
         'termsAndConditionsAgreedVersion',
         'termsAndConditionsAgreedAt',
+        'allowEmbedIframes',
       ],
       boolean: {
         followedByCurrentUser:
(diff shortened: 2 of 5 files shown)


appinteractive:
What if I want to decide based on provider?

ogerly:
> What if I want to decide based on provider?

I've been thinking about that, too. 
basically we have a list that is manageable. 
if it should be nice. must be with the confirmations a list to select.

roschaefer:
@appinteractive @ogerly [KISS](https://en.wikipedia.org/wiki/KISS_principle) please! One setting for all or no provider is enough, in my opinion.

Every piece of functionality or programming logic is adding more bugs: If we have a list, we must make sure that it works as expected. If we don't have a list, then it's a lot easier.

appinteractive:
I agree with @roschaefer, my comment was related with saving the setting as it was not clear to me if it gets forcebly saved when I show it once.

backend/src/models/User.js
```diff
@@ -104,4 +104,8 @@ module.exports = {
     target: 'Location',
     direction: 'out',
   },
+  allowEmbedIframes: {
+    type: 'boolean',
```
roschaefer: You could even think about using a `date` here. The `false` would be if it's null.

mattwr18: I like that suggestion... maybe it's useful to know at some point when the user opted in to this "feature"

mattwr18: Although, it would have to be the `String` representation of a `datetime`, no @roschaefer?

ogerly:
> @ogerly looks good but I would prefer to have a user-facing feature before we merge changes to the code. It would be confusing to see these changes if they have no effect right now.

 

this PR and the PR https://github.com/Human-Connection/Human-Connection/pull/1682 I would like to have together already in it around the issue to convert https://github.com/Human-Connection/Human-Connection/issues/1684

ogerly:
the PR is integrated here https://github.com/Human-Connection/Human-Connection/pull/1685

PR #1658, opened 2019-09-20
🍰 Notifications self update and refactoring

### LINKED ISSUE #1637: 🐛 [Bug] Notifications do just update with page reload
## :bug: Bugreport

The notifiactions are not counting down when viewing the related content. Just after a page reload.

1. get a notification
2. click on the icon and choose some content
3. The notifications do not change or count down or disappear.
4. Reload the page and everythinig is fine
5. Profit

### Expected behavior

The notification counter should decrease when viewing the related content.

### Version & Environment
 Type: [Desktop]      
 - OS: [Linux]      
 - Browser: [Firefox] 
 - Version [69.0]  
 - Device: [Notebook]  

### Additional context

Maybe #1332 and #1456

Tirokk:
## 🍰 Pullrequest

Notifications self update and refactoring.

### Issues

- fixes #1637

### Todo

- [ ] Change notification query to not older than 1 day if already red.


roschaefer:
@Tirokk note that the actual bug is not the list but the counter on the menu button. The *counter* does not get updated when you click on a notification. We should fix that first before we do this kind of stuff.

webapp/components/notifications/NotificationMenu/NotificationMenu.vue
```diff
@@ -18,6 +18,7 @@
 
 <script>
 import Dropdown from '~/components/Dropdown'
+import { REFRECH_MILISEC } from '~/constants/notifications'
```
mattwr18:
```suggestion
import { REFRESH_MILLISEC } from '~/constants/notifications'
```

webapp/components/notifications/NotificationMenu/NotificationMenu.vue
```diff
@@ -35,7 +36,39 @@ export default {
   props: {
     placement: { type: String },
   },
+  created() {
+    setInterval(this.updateNotifications, REFRECH_MILISEC)
```
mattwr18:
```suggestion
    setInterval(this.updateNotifications, REFRESH_MILLISEC)
```

webapp/constants/notifications.js
```diff
@@ -0,0 +1 @@
+export const REFRECH_MILISEC = '60000'
```
mattwr18:
```suggestion
export const REFRESH_MILLISEC = '60000'
```

mattwr18:
> @Tirokk note that the actual bug is not the list but the counter on the menu button. The _counter_ does not get updated when you click on a notification. We should fix that first before we do this kind of stuff.

I don't think I agree with this @roschaefer... if we were talking about it before and planning, maybe I would, but the work is already done and a `PR` is already put in... What's the harm in merging in a fix that doesn't close this issue, but deals with another one?
Also, it's `[WIP]`, so maybe @Tirokk planned to deal with the counter in a separate commit, maybe he can if you think it's important in this `PR`, but from my perspective, there would be no imperative to merge the counter fix before this one.

Could you explain what you mean by "we should fix..."?

roschaefer:
@mattwr18 ok ok! I don't mean *don't* merge this PR I just wanted to make @Tirokk aware that the bug  that @datenbrei reported is not about the list but about the counter. That's something I got quite lately. But maybe it's just me having a slow uptake :laughing:

webapp/constants/notifications.js
Tirokk: Thanks for your hint @mattwr18 👍🏼

Tirokk:
Thanks for the `count the unread` hint @roschaefer 👍🏼
I had that as well in mind.

Here our support has to explain:

Bildschirmfoto 2019-09-21 um 07 49 27

Translation:
```
Notifications come only on page reload - press F5
```

webapp/components/notifications/NotificationMenu/NotificationMenu.vue
```diff
@@ -35,7 +49,39 @@ export default {
   props: {
     placement: { type: String },
   },
+  created() {
+    setInterval(this.updateNotifications, REFRESH_MILLISEC)
+  },
+  destroyed() {
+    clearInterval(this.updateNotifications)
+  },
```
roschaefer: scared

Please use `pollInterval` by `vue-apollo` and don't re-invent the wheel! See: https://vue-apollo.netlify.com/api/smart-query.html#options

webapp/components/notifications/NotificationMenu/NotificationMenu.vue
```diff
@@ -35,7 +49,39 @@ export default {
...
+    async updateNotifications() {
+      try {
+        const {
+          data: { notifications },
+        } = await this.$apollo.mutate({
+          mutation: notificationQuery(this.$i18n),
+        })
+        // add all the new notifications to the old notifications at top of the list
+        if (notifications) {
+          notifications.forEach(udatedListNotification => {
```
roschaefer:
```suggestion
          notifications.forEach(updatedListNotification => {
```

webapp/components/notifications/NotificationMenu/NotificationMenu.vue
```diff
@@ -35,7 +49,39 @@ export default {
...
+                oldListNotification.from.id === udatedListNotification.from.id &&
+                oldListNotification.createdAt === udatedListNotification.createdAt &&
+                oldListNotification.reason === udatedListNotification.reason
+              )
+            })
+            if (sameNotification === undefined) {
+              this.notifications.unshift(udatedListNotification)
+            }
+          })
+        }
```
roschaefer: I find this code way to complex. Can't you just append two arrays and the remove duplicates based on a criteria? I googled and found https://medium.com/dailyjs/how-to-remove-array-duplicates-in-es6-5daa8789641c there is also one by Mozilla using `reduce` https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/Reduce#Remove_duplicate_items_in_array

webapp/components/notifications/NotificationMenu/NotificationMenu.vue
```diff
@@ -35,7 +49,39 @@ export default {
...
+  destroyed() {
+    clearInterval(this.updateNotifications)
+  },
   methods: {
+    async updateNotifications() {
+      try {
+        const {
+          data: { notifications },
+        } = await this.$apollo.mutate({
+          mutation: notificationQuery(this.$i18n),
```
roschaefer: Why is it called `notificationQuery` if it's a mutation? :thinking:

webapp/components/notifications/NotificationMenu/NotificationMenu.spec.js
```diff
@@ -79,12 +79,26 @@ describe('NotificationMenu.vue', () => {
...
+                  },
+                },
+              },
             ],
           }
         }
       })
 
-      it('displays the total number of notifications', () => {
+      it('displays the unred number of notifications', () => {
```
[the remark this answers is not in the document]
Tirokk: Ups I wrote the past tense after its phonetic that is like „red“. Sorry!

webapp/components/notifications/NotificationMenu/NotificationMenu.vue
```diff
@@ -58,6 +104,15 @@ export default {
     totalNotifications() {
       return (this.notifications || []).length
     },
+    unredNotifications() {
+      let countUnread = 0
```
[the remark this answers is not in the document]
Tirokk: I like this ! 😍

webapp/components/notifications/NotificationMenu/NotificationMenu.vue
Tirokk: I did this locally …
Thanks!

webapp/components/notifications/NotificationMenu/NotificationMenu.vue
Tirokk: You are right @roschaefer ! It is no mutation, but a query!
This is a copy and paste left over. 😱

webapp/components/notifications/NotificationMenu/NotificationMenu.vue
Tirokk: All your examples @roschaefer are no really working, because if the `read` property changes in my code it should be equal as well. And therefore the new fetched notification should not be added to the beginning …

mattwr18:
Any idea @Tirokk why there are so many duplicate commits?

webapp/components/notifications/NotificationMenu/NotificationMenu.vue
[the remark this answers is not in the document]
Tirokk: But this does not function at all.
Don’t know why and will not spend hours of investigating for saving not on single line and add a bit of complexity …

webapp/components/notifications/NotificationMenu/NotificationMenu.vue
Tirokk: Yeah !!!

Tirokk:
I overtook @roschaefer suggestions.
I was mistaken in writing `unred` instead of `unread` in many places …

webapp/components/notifications/NotificationMenu/NotificationMenu.vue
```diff
@@ -50,32 +50,20 @@ export default {
   },
   watch: {
     notifications: {
+      immediate: true,
       handler(lastQueriedNotifications) {
         if (lastQueriedNotifications && lastQueriedNotifications.length > 0) {
-          // set this to be empty to get always called if a query comes with results from the backend
-          // has the sideeffect the handler is encouraged to be called again, but only once with no effect, because of the if above
-          this.notifications = []
```
Tirokk: Why you removed this line of code? After our discussion …

@mattwr18 You and I agreed yesterday to leave this line in the code with a specific comment, that I found out that this was needed for a resolution of a misfunctioning I saw, but that I can not exactly say at the moment what this was and we both can not find out right now.

Please don’t do that without talking to me again! 🙏🏼

mattwr18: It wasn't done on purpose and I have no idea how it was deleted... maybe I forgot to save the file before pushing up.
As I didn't do it mindfully, I cannot guarantee that I will not do it again, such it is in life.
Thanks for your understanding

webapp/components/notifications/NotificationMenu/NotificationMenu.vue
```diff
@@ -46,24 +82,39 @@ export default {
...
-        this.notifications = this.notifications.map(n => {
-          return n.from.id === markAsRead.from.id ? markAsRead : n
+        this.displayedNotifications = this.displayedNotifications.map(n => {
+          return this.equalNotification(n, markAsRead) ? markAsRead : n
         })
       } catch (err) {
         throw new Error(err)
       }
     },
+    equalNotification(a, b) {
```
roschaefer: ~I would not register this on the component if it's not being used in the template or elsewhere. You can assign the function in another function, see above :point_up:~

What I don't understand: Why is the check for `id` not enough?

webapp/components/notifications/NotificationMenu/NotificationMenu.vue
```diff
@@ -1,23 +1,35 @@
 <template>
-  <ds-button v-if="totalNotifications <= 0" class="notifications-menu" disabled icon="bell">
-    {{ totalNotifications }}
-  </ds-button>
+  <div
+    v-if="totalNotifications <= 0"
+    class="notifications-menu-pointer"
+    @click.prevent="updateNotifications"
+  >
+    <ds-button class="notifications-menu" disabled icon="bell">{{ unreadNotifications }}</ds-button>
```
roschaefer:
```suggestion
    <ds-button class="notifications-menu" disabled icon="bell">{{ unreadNotificationsCount }}</ds-button>
```

webapp/components/notifications/NotificationMenu/NotificationMenu.vue
```diff
@@ -1,23 +1,35 @@
 <template>
-  <ds-button v-if="totalNotifications <= 0" class="notifications-menu" disabled icon="bell">
-    {{ totalNotifications }}
-  </ds-button>
+  <div
+    v-if="totalNotifications <= 0"
```
roschaefer: Why do we have to add this extra `v-if` here? :thinking:

webapp/components/notifications/NotificationMenu/NotificationMenu.vue
```diff
@@ -29,13 +41,37 @@ export default {
...
     }
   },
   props: {
     placement: { type: String },
   },
+  watch: {
+    notifications: {
+      immediate: true,
+      handler(lastQueriedNotifications) {
+        if (lastQueriedNotifications && lastQueriedNotifications.length > 0) {
```
roschaefer: Prefer guard clause over if statement

```suggestion
  if (!(lastQueriedNotifications && lastQueriedNotifications.length > 0)) return
```

webapp/components/notifications/NotificationMenu/NotificationMenu.vue
```diff
@@ -29,13 +41,37 @@ export default {
...
+          lastQueriedNotifications.forEach(updatedListNotification => {
```
roschaefer: Use `filter` instead of `forEach`:

```js
const newNotifications = lastQueriedNotifications.filter(new => {
    return !this.displayedNotifications.find(old => sameSame(new, old))
])
```

webapp/components/notifications/NotificationMenu/NotificationMenu.vue
```diff
@@ -1,23 +1,35 @@
...
   <dropdown v-else class="notifications-menu" :placement="placement">
     <template slot="default" slot-scope="{ toggleMenu }">
-      <ds-button primary icon="bell" @click.prevent="toggleMenu">
-        {{ totalNotifications }}
+      <ds-button
+        primary
+        icon="bell"
+        @click.prevent="
+          toggleMenu()
+          updateNotifications()
```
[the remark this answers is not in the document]
roschaefer: Actually I prefer not to `updateNotifications` when i click on the notifications. It could happen that the list gets updated just when I want to click on a notification.

Can't we just let the notifications update on their own?

webapp/components/notifications/NotificationMenu/NotificationMenu.vue
```diff
@@ -29,13 +41,37 @@ export default {
...
+  },
+
   methods: {
+    async updateNotifications() {
+      try {
+        this.$apollo.queries.notifications.refetch()
+      } catch (err) {
+        throw new Error(err)
+      }
+    },
```
roschaefer: I would prefer to have this callback deleted. I believe it creates bad user experience, see above :point_up:

webapp/components/notifications/NotificationMenu/NotificationMenu.vue
```diff
@@ -29,13 +41,37 @@ export default {
...
+          lastQueriedNotifications.forEach(updatedListNotification => {
+            const sameNotification = this.displayedNotifications.find(oldListNotification => {
+              return this.equalNotification(oldListNotification, updatedListNotification)
+            })
+            if (!sameNotification) this.displayedNotifications.unshift(updatedListNotification)
+          })
+        }
+      },
+    },
+  },
```
roschaefer: @Tirokk you don't have to put this into a `watch` handler.

Vue-apollo has the `update` for graphql queries. We use it in a couple of other locations.  It's suitable for this task when you want to define **how** the `notifications` should be updated (in your case, you want to merge the new notifications with the old ones).

See here: https://vue-apollo.netlify.com/api/smart-query.html#options

We cannot use `fetchMore` as far as I know, because we're not using pagination here :thinking:

webapp/components/notifications/NotificationMenu/NotificationMenu.vue
```diff
@@ -29,13 +41,37 @@ export default {
...
+  watch: {
+    notifications: {
+      immediate: true,
+      handler(lastQueriedNotifications) {
+        if (lastQueriedNotifications && lastQueriedNotifications.length > 0) {
+          lastQueriedNotifications.forEach(updatedListNotification => {
+            const sameNotification = this.displayedNotifications.find(oldListNotification => {
+              return this.equalNotification(oldListNotification, updatedListNotification)
+            })
+            if (!sameNotification) this.displayedNotifications.unshift(updatedListNotification)
```
roschaefer: Instead of `unshift` I would use `concat` and then `sort` by `createdAt`.

https://developer.mozilla.org/de/docs/Web/JavaScript/Reference/Global_Objects/Array/concat
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/sort

Who knows if the notifications arrive in the right order?

webapp/components/notifications/NotificationMenu/NotificationMenu.vue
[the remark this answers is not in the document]
mattwr18: @roschaefer it might not be clear, but clicking on it to update only works if there are no `totalNotifications`, so it's not possible to update if you want to click on a notification

webapp/components/notifications/NotificationMenu/NotificationMenu.vue
mattwr18: there is a `v-if` because there is the additional functionality that the `NotificationsMenu`, if clicked on when there are no notifications, updates the notifications

webapp/components/notifications/NotificationMenu/NotificationMenu.vue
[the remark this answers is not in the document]
Tirokk: If I need to have a quick conversation with somebody this way I am able to get the notifications menu as quick updated as possible. Also in case the automatic update stops for any reason.

I discussed this with @datenbrei and we agreed to do it this way.

To have the menu not updated while it is open we could say:
```
 @click.prevent="
          await updateNotifications()
          toggleMenu()
 "
```
But then there will be a delay before the menu opens …
I my eyes there is no big possible harm if the open menu sometimes updates after I click on it.

PS: The `toggleMenu` function is undefined in any method as far I have tried …

webapp/components/notifications/NotificationMenu/NotificationMenu.vue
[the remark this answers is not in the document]
mattwr18: this might very well work @roschaefer, but does it really reduce complexity? 
`KISS`?

webapp/components/notifications/NotificationMenu/NotificationMenu.vue
mattwr18: when you say who knows, can you explain that? 
we are currently asking for them back with `orderBy: updatedAt`
doesn't that mean either we do know the order they arrive or something is broken in the functionality of either `neo4j-graphql-js` and/or `neo4j`?

webapp/components/notifications/NotificationMenu/NotificationMenu.vue
mattwr18: Also, @Tirokk, I submit as proof I didn't purposely remove the controversial line of code with comments to explain to the `PR` reviewer. 
Do you remember when we discussed and I changed this line of code to be `totalNotifications === 0`??
Must not have saved the file, or some other quirk with `git`

webapp/components/notifications/NotificationMenu/NotificationMenu.vue
Tirokk: Somebody can mention me in the same post again by editing. Then a new notification with the same `from.id` is created. But the first one can be still in the list and be read already.
The second unread one will not appear if you check only for `from.id `, even though it is in the database as separate notification with an other `read` value and a different `createdAt`.

webapp/components/notifications/NotificationMenu/NotificationMenu.vue
Tirokk: The extra `div` with this `v-if` is to switch between the disabled and enabled button and have the ability to update it on click even if disabled. Like @mattwr18 explained.

webapp/components/notifications/NotificationMenu/NotificationMenu.vue
Tirokk: done

webapp/components/notifications/NotificationMenu/NotificationMenu.vue
Tirokk: Done

webapp/components/notifications/NotificationMenu/NotificationMenu.vue
Tirokk: I like this! 😍
So we make a sports out of finding the shortest coding.
I just like to know the headline …

webapp/components/notifications/NotificationMenu/NotificationMenu.vue
Tirokk: Sorting is not necessary in my eyes.
Should be the right way anyhow and if not in my eyes its better to have the new notifications at the beginning.

webapp/components/notifications/NotificationMenu/NotificationMenu.vue
Tirokk: Now it is so perfect that I can't even test my code anymore on my own.
Because if I write myself a notification it updates immediately at sending the comment. 
😂🤣😅😰 TOP !!!
But I could use two browsers ???

Tirokk: Isn’t `refetch` enough?

webapp/components/notifications/NotificationMenu/NotificationMenu.vue
Tirokk: Why bad user experience?
Because of the possible error?
So I shall use `this.$apollo.queries.notifications.refetch()` directly?
Or get rid of the update at all? Like you said …

I don’t agree with the last, like I said above …

webapp/components/notifications/NotificationMenu/NotificationMenu.vue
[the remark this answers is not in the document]
Tirokk: This add complexity in my eyes and is bad understandable.
Like @mattwr18 I say `KISS`!

webapp/components/notifications/NotificationMenu/NotificationMenu.vue
Tirokk: Okay, I understand.

webapp/components/notifications/NotificationMenu/NotificationMenu.vue
roschaefer: I don't know exactly if it could happen in this very example. But since we merge the `read` notifications on the client which are not part of the response that comes from the server (thus not guaranteed to be sorted) I find it good practice to order the list on the client again.

Here's a (very speculative) real world example. Consider we are connected to the fediverse. A notification was created for the user on the other server.

1. Sb. mentions you on a remote server
2. You fetch the notifications
3. The notifications from the remote server arrive at your server
4. You fetch the notifications again. Now the old notification gets put on top even though the `createdAt` is older. This would break the order of the list by `createdAt`.

webapp/components/notifications/NotificationMenu/NotificationMenu.vue
roschaefer: I was confused by `update` earlier: It triggers two times. Feel free to debug it with `console.log`. The first call comes when apollo reads from the cache. As soon as the new results from the server arrives, `update` is called again. It might be that apollo is so smart that it updates the cache if you mark a notification as `read`. :man_shrugging:

roschaefer: Maybe `refetch` also works. I think `refetch` only triggers once. We use it for pagination.

webapp/components/notifications/NotificationMenu/NotificationMenu.vue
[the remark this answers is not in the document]
roschaefer: I think my head explodes right now.

Up to you. If you feel it's a great idea to update the list when you click on notifications - feel free to do it that way. I would prefer to implement the smallest solution first and get some user feedback before moving on the more sophisticated solution. But up to you. I have confidence in your decisions :wink:

webapp/components/notifications/NotificationMenu/NotificationMenu.vue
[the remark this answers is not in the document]
roschaefer: I thought a `fold` is not so complex but I might be wrong :neutral_face: 

Here's the concept on Wikipedia: https://en.wikipedia.org/wiki/Fold_(higher-order_function)

webapp/components/notifications/NotificationMenu/NotificationMenu.vue
roschaefer: Bad user experience because of the situation where you are about to click on a notification but in the very moment when you want to click on it, the list gets updated and you suddenly click on sth. else.

webapp/components/notifications/NotificationMenu/NotificationMenu.vue
roschaefer: You don't *have* to implement it. It was an intuition of mine, because we merge something sorted into something not sorted. Feel free to ignore this suggestion

mattwr18: I like your explanation, I hadn't thought of that. I think it doesn't hurt to do it either, I'm sure it's not expensive, especially since the list is more or less sorted anyways

Tirokk: I sorted now!

webapp/components/notifications/NotificationMenu/NotificationMenu.vue
Tirokk: I implemented now two new things and got rid of the click update. See my general comment of the changes 👇🏼

Tirokk:
I implemented now two new things and got rid of the click update:

1. Update is already triggered on hover over the menu button. This gives the update kind of an additional second to be done before trigger to open the menu by clicking.
2. Implemented the opening of the menu right after been updated, but latest after a small delay of two seconds after the click on the menu.

- Therefore the menu has kind of three seconds to be updated before opening.
- There will be always a danger of being updated while selecting a menu item, because the update interval runs in the background anyways.
(Would be exactly nice to start the `pollInterval` time by `zero` to avoid this right after opening the menu. Have not research that yet …)
- And the user will kind of never click right after opening the menu, when there is a rest danger of updating by the hover, because usually they have to read what is new in the notices.

webapp/components/notifications/NotificationMenu/NotificationMenu.vue
```diff
@@ -46,24 +72,86 @@ export default {
...
+        return NOTIFICATIONS_POLL_INTERVAL
+      },
+      update(data) {
+        const newNotifications = data.notifications.filter(newN => {
+          return !this.displayedNotifications.find(oldN => this.equalNotification(newN, oldN))
+        })
+        this.displayedNotifications = newNotifications
+          .concat(this.displayedNotifications)
+          .sort((a, b) => {
+            return a.createdAt === b.createdAt
```
roschaefer: Why not:
```suggestion
  return new Date(a.createdAt) - new Date(b.createdAt) // maybe reverse a with b, I'm not sure
```

Tirokk: Thanks for the hint @roschaefer ! 😍
Done

Tirokk:
I removed all the user triggered updates. 😅

Tirokk:
Since I pulled in the latest master I have locally an unrelated intermittent failing test in the backend Jest tests in file `src/schema/resolvers/shout.spec.js` named `shout and unshout posts › unshout › authenticated › another user's post` which I couldn’t solve till now:

Bildschirmfoto 2019-09-30 um 11 14 02

If I run this test only it’s not failing …

Tirokk:
Nice approvement @roschaefer 👍🏼

Issue #1466, opened 2019-09-04
🚀 [Feature] Image cropping


alina-beck:
## :rocket: Feature
reported by @ulfgebhardt 

I would like to control how images are cropped when I create a post. At the moment they are cropped automatically – and sometimes the most important part of the image is cut off:

![image](https://user-images.githubusercontent.com/11694483/64242978-6f8c6e00-cefe-11e9-8454-09c1e10ed7f8.png)

mattwr18:
we agreed in our internal meeting today with @alina-beck, @Tirokk, @datenbrei, @DennisHack, @ogerly, and I that for the time being we will allow free ratio cropping by the user with a max-height of say `2000px`, which can be changed, if desired.

This means that at the moment, there will be cases where another user must scroll down to see the additional content of the `Post`. We agreed that an additional issue could be created to autocrop on the `Post` page, and allow the users to click on the image to see the original size.

If you are mentioned above, please check what I've written for accuracy and your understanding of what we agreed on.

datenbrei:
agreed.

mattwr18:
basic image cropping added in https://github.com/Human-Connection/Human-Connection/pull/1666

PR #1666, opened 2019-09-21
🍰 Implement basic image cropping solution

vardyb:
#1466

- co-authored-by: mattwr18 <mattwr18@gmail.com>

## 🍰 Pullrequest
Adds the ability to crop an image when uploading:

### Issues
- relates #1466

### Todo
- [x] Improve interface so not full screen?
- [ ] Look at building tests


alina-beck:
So, I am testing this amazing new feature right now (love it! 💚) and have to say it feels very smooth – it's great that you managed to fit the cropping tool into the size of the create-post-card! 😍 

But there's an issue we probably still need to think about... When I crop an image like this:
Screenshot 2019-10-01 at 11 29 55

Here's what I see on the post page:
Screenshot 2019-10-01 at 11 30 12

While I would of course expect to see the image _just as I cropped it_.

I think right now we have hard-coded a `max-height` for images because we wanted to avoid suuuperlong images that would force you to scroll down three screens to reach the post content. And I still think we should limit the height of images somewhat to avoid that exact problem, although we could probably give quite a bit more flexibility.

Does the cropping tool allow us to restrict the aspect ratio to a certain range? That would be ideal because we have a fixed width and could then limit the height implicitly by only allowing certain aspect ratios – thus getting rid of the need for a `max-height`.

alina-beck:
hmm, not so easy @mattwr18. I think indeed that one fixed aspect ratio is not what we want! but I saw the package has `minCropBoxWidth` and `minCropBoxHeight` properties, although it says _'This size is relative to the page, not the image'_ whatever that means... 🤷‍♀

maybe we could play around with those values and see what happens? :grin:

mattwr18:
sure, would you be available for a pairing session later today/tomorrow @alina-beck ?
if not, I could play around with it, and take screen shots, or vice versa?

alina-beck:
@mattwr18 yes, let's have a look at it together! working on the embeds with Alex now, I'll let you know when I'm available. :smile:

mattwr18:
great! I've got a pairing in 30 minutes, but just let me know when you're ready!!

webapp/package.json
```diff
@@ -44,7 +44,8 @@
     ],
     "moduleNameMapper": {
       "^@/(.*)$": "<rootDir>/src/$1",
-      "^~/(.*)$": "<rootDir>/$1"
+      "^~/(.*)$": "<rootDir>/$1",
+      "\\.(css|less)$": "identity-obj-proxy"
```
roschaefer: What is that? :thinking:

mattwr18: it is for mocking CSS modules in Jest
so we are importing the CSS for cropperjs, but it breaks the tests, so we need to mock it
https://jestjs.io/docs/en/webpack#mocking-css-modules

mattwr18:
I don't understand that well either @roschaefer, here is some site that gives some of the best aspect ratios for popular social media https://buffer.com/library/ideal-image-sizes-social-media-posts
or maybe you prefer this one
https://www.shutterstock.com/blog/common-aspect-ratios-photo-image-sizes

Cropper, "By default, the crop box is free ratio." 
https://github.com/fengyuanchen/cropperjs

mattwr18:
hey @alina-beck I spent a couple hours trying to follow your great suggestions, but failed miserably hehe

the first things i tried to achieve was to flatten out the hierarchy from the `postdropzone` div to the `ds-icon` with the letter overlay, but I ran into the issue that when an image is added `vue-dropzone` adds a `dz-message`, I added the `justify-content` and `align-items` `center` to the #postdropzone, which worked before an image is added, but not after...

I also tried to call multiple methods to shorten the `transformImage` method, but it was getting more complicated trying to pass down everything each method needed. Finally, I tried the idea of using divs with `v-if`s instead of creating a `div`, but I just broke the functionality.

I would be happy to pair on it, or if someone else wanted to give it a go, but as we discussed, it might be nice to get the feature out to the user base and refactor later? don't know what others think...

alina-beck:
Hey @mattwr18 I think I just went down the same rabbit hole... 😊 I thought it would be a good idea to extract the `.hc-drag-marker` into its own `ImageUploadMarker` component and then use it on the profile page – which was no problem at all! – and in the teaser image dropzone. Well, _that_ didn't work because of the `.dz-message` thingy `vue-dropzone` is adding – as you already pointed out.

I am once again amazed by these packages claiming to allow the use of 'custom components' and then wrapping them in a `<div>` in a `<div>` in a `<span>`... good luck styling that! Do you know if this is a `vue` specific thing? I've never seen this done in `react` components, not this bad anyways. And it's _sooo_ frustrating. 😕 

So, I don't know. I'm still not very happy with the component code – creating elements on the fly, duplicating a lot of css – but I'm failing to improve it, same as you. Maybe someone else has an idea? Or we merge it as is to finally get the awesome cropping functionality into the project (and secretly hope we never need to touch that code again 😜)? What do you think?

mattwr18:
I think what I like most about it is the way it works, the code itself is a bit sloppy and I'm sure we could improve the styling, etc...
I feel like there's not a strong argument for delaying it from getting merged. Maybe we could refactor it and not use `vue-dropzone` later? or we could hope that someone will be motivated to refactor it, or that it just works without needing to touch it again...

aonomike:
I am not sure I have a huge opinion as I havent done much testing of the feature but from the reading I want to agree with both of you @mattwr18 and @alina-beck that we get this merged in :)

Tirokk:
To have the ability of cropping is wonderful in my eyes !!! 💫🚀
So it's great to have it. 😍

I didn’t cared for the code right now and just tried the functionality.
And already with my first try was not successful to reach my goal. We seem to have a fixed aspect ratio yet. This is not working for my picture at all:

- *May be for the future:* First is the cropping tool makes not a suggestion how the image could be cropped. See first screenshot.
- My picture is not a special case I think: With a fixed aspect ratio our users will have a lot of fun and we with the complains as well I fear. Like you see, I am not even able to crop my picture in a way to have all important parts in it and get rid of the parts I don’t want. See second screenshot.

No kind of fit at the beginning at all - Bildschirmfoto 2019-10-14 um 09 07 25

I am not even able to get the importent parts of my picture in the crop - Bildschirmfoto 2019-10-14 um 09 07 56

I have the feeling that we need a meeting where we discuss the goals of the cropping and the needs of certain aspect ratios.

mattwr18:
hey @Wolle (Hamburg) thanks for your comments on the functionality
:slight_smile:
do you have any suggestions on how to have free aspect ratio with our use case?
do you think it's a good idea to have no limits on the height of the image for example?
or do you think we should let the user have free ratio, but then autocrop their image?
do you agree with @Greg that we should autocrop and have a gradient that hopefully makes it clear the user should click on it to see the original?

Tirokk:
With cropping @mattwr18#1510 we need a meeting to discuss the many different needs and reasons may be …

„Users (post writers)“ perspective:
- The user will get rid of image parts they don't want.
- And the user like that all parts of the picture important for them can be seen by others.
- In my eyes it is questionable if a user is so annoyed if they go on a post page and sometimes can not read  the title right away, because the image is so long that the title is not visible in the window. Mostly the user has read the title already on the newsfeed page or in a post card somewhere …

"Developers" perspective:
- On the newsfeed page theoretically we are flexible for the aspect ratio. Accept may be for really high images. But they don't exist usually. So I would say KISS and doubt that this practically is an issue. And if it gets an issue, we could set an max aspect ratio for the hight later.
- On the post page it would be nice to have the title readable in the window always right away. But it will not happen always. Are there users have complaint about?
- Gregs solution to have a gradient on the image at the moment I don't agree so much. A gradient could be part of the image and may not be seen by the users as a fade away …
- Cropping can make very important parts of the image invisible. A textual underline in the image can be cropped away and destroys the effect of the teaser image at all. Even if you know you have to click on the image to see all parts the effect of an underline is not the same as it appears on the post page. Therefore I would be very careful with cropping.

mattwr18:
I'll have a look @roschaefer, but to do with the borders, I think there is a legitimate reason for that to be allowed... if a user wants a portrait image with white space on each side, for example..
is that right @alina-beck ??

mattwr18:
@roschaefer I have only been able to reproduce the error you pointed out by selecting more than one photo at the same time... if you add an image and then change your mind and add another, it replaces the image with the new one.

If you click on the image and select two or more images, they will both be shown, but I don't think we can do anything about this...

alina-beck:
really sorry @mattwr18 and @vardyb – just tested the latest commits and found yet more issues... 😞 

1. the upload button is hidden on screens smaller than `960px` (see screenshot below)
2. the `max-height` is ignored both by the `cropper` tool and on the `index` page (but works on the `post` page!)
3. the `cropper` sits in front of the header, so the header is hidden when scrolling while cropping (not a deal breaker but might be a quick win, so I'm including it in the list)

Screenshot 2019-10-14 at 15 59 40

the issue @roschaefer mentioned – about the _void_ – is probably rather easily solved by disabling to zoom in the cropping tool, don't you think? I'd like that, too. Not so much because I'm scared of the void but rather because I accidentally zoom in the picture all the time and it's a bit annoying. But maybe that's just me. :blush:

webapp/components/TeaserImage/TeaserImage.vue
```diff
@@ -178,7 +195,17 @@ export default {
...
+  position: relative;
+  width: 100%;
+  background-color: #000;
+}
+
+.crop-confirm {
+  position: absolute;
+  left: 10px;
+  top: 10px;
+  z-index: 9999;
```
alina-beck:
```suggestion
  z-index: 1;
```

this puts the button on top of the overlay but under the header ;)

webapp/pages/index.vue
```diff
@@ -190,6 +190,10 @@ export default {
 </script>
 
 <style lang="scss">
+.ds-card-image img {
+  max-height: 2000px;
```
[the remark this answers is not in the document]
mattwr18: yeah, there are some that are scoped, for other CSS reasons I guess :worried:

mattwr18:
I merged @roschaefer, but not because I don't like your ideas...

The button placement could be discussed, it doesn't matter too much to me
we are manually adding `eventListeners` to the button, so it would really be nice to try and refactor to use more vue-friendly html, but @alina-beck and I attempted and weren't immediately successful

I'll have a pairing session with @vardyb where I'd love to go over the changes, and how we got there, maybe he'll have fresh energy to continue improving the UX... I think it might be nice actually to have a preview of what the image would look like, and a button to crop, so that users are not forced to take this extra step, if they don't want...

Not sure how to test, though I don't like that at all..
Also, something to keep in mind is that `vue-dropzone`, which is what we base our upload off, is looking for `CO-MAINTAINERS`, since they are a bit overwhelmed.
https://github.com/rowanwins/vue-dropzone#vue-dropzone

PR #1668, opened 2019-09-22
Make Human Connection a Progressive Web App

roschaefer:
Following this tutorial here: https://pwa.nuxtjs.org/setup.html

## 🍰 Pullrequest
To my understanding, this is enough to let chrome users on Android add a "Shortcut" to their Home Screen. If we get a Google Account we can even add it to the Play Store.

### Issues

- None

### Todo

- [X] None


roschaefer:
Yes, found the lighthouse plugin, too! I wonder how we can integrate that into our CI/CD pipeline on several pages that require authentication

PR #1678, opened 2019-09-23
Improve comments output

mattwr18:
before, collpased:
Screenshot from 2019-09-23 09-03-00

uncollapsed:
Screenshot from 2019-09-23 09-03-10

after, collapsed:
Screenshot from 2019-09-23 09-03-31

uncollapsed:
Screenshot from 2019-09-23 09-03-39

### DIFF
--- a/webapp/components/Comment.vue
+++ b/webapp/components/Comment.vue
@@ -38,14 +38,14 @@
       </div>
       <div v-show="!openEditCommentMenu">
         <content-viewer
-          v-if="comment.content.length < 400"
+          v-if="$filters.removeHtml(comment.content).length < 180"
           :content="comment.content"
           class="padding-left"
         />
         <div v-else class="show-more-or-less-div">
           <content-viewer
             v-if="isCollapsed"
-            :content="$filters.truncate(comment.content, 400)"
+            :content="$filters.truncate(comment.content, 180)"
             class="padding-left text-align-left"
           />
           <span class="show-more-or-less">
--- a/webapp/package.json
+++ b/webapp/package.json
@@ -77,6 +77,7 @@
     "tippy.js": "^4.3.5",
     "tiptap": "~1.25.0",
     "tiptap-extensions": "~1.27.0",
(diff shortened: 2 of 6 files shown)


webapp/components/Comment.vue
```diff
@@ -38,14 +38,14 @@
...
-          v-if="comment.content.length < 400"
+          v-if="$filters.removeHtml(comment.content).length < 180"
           :content="comment.content"
           class="padding-left"
         />
         <div v-else class="show-more-or-less-div">
           <content-viewer
             v-if="isCollapsed"
-            :content="$filters.truncate(comment.content, 400)"
+            :content="$filters.truncate(comment.content, 180)"
```
roschaefer: Please refactor the 180 to be not such a magic number. You could put it into `constants/comments.js`.

webapp/package.json
```diff
@@ -77,6 +77,7 @@
     "tippy.js": "^4.3.5",
     "tiptap": "~1.25.0",
     "tiptap-extensions": "~1.27.0",
+    "trunc-html": "^1.1.2",
```
roschaefer: Very good. You chose the one which has more Github stars than the other two options.

mattwr18: hehe, I chose the one we use in the `excerptMiddleware`, but whoever chose it must have looked at the GitHub stars :smile_cat:

webapp/components/Comment.vue
Tirokk: I agree with @roschaefer . 😍

webapp/plugins/vue-filters.js
```diff
@@ -45,9 +46,9 @@ export default ({ app = {} }) => {
       if (length <= 0) {
         return value
       }
-      let output = value.substring(0, length)
+      let output = trunc(value, length).html
       if (output.length < value.length) {
-        output += '…'
+        output += ' …'
```
Tirokk: I really love this @mattwr18 ! 💫😍
The ellipse the most and for sure the space as well. 💥

PR #1701, opened 2019-09-24
fix the bug with scrolling post comments into view

### LINKED ISSUE #1273: 🐛 [Bug] TypeError: Cannot read property 'offsetTop' of null
## :bug: Bugreport
Clicking around on a post page can lead to this error page:

2019-08-15-213240_1920x1080_scrot

1. Click on a post with 0 comments
2. Click on "Comments"
3. Click on "More Info"
4. Click on "Comments" again
5. Boom! :boom: 

Screenshot - 2019-08-15T213336 977

### Expected behavior
Should see the comments

### Version & Environment
```
robert@e480 ~/D/h/Human-Connection> chromium --version
Chromium 76.0.3809.100 Arch Linux
robert@e480 ~/D/h/Human-Connection> cat /proc/version
Linux version 5.2.6-arch1-1-ARCH (builduser@heftig-3961) (gcc version 9.1.0 (GCC)) #1 SMP PREEMPT Sun Aug 4 14:58:49 UTC 2019
```
### Additional context
`git blame` shows me that the bug was introduced in 0a66ad56d. @Tirokk @ogerly never commit any code that has javascript globals like `window` or `document` without testing it for server side rendering. Feel free to remove the whole `watcher` block along with the link to comments - it's not that useful in my opinion.

vbelolapotkov:
## 🍰 Pullrequest
Move watcher from `post/_id` to `CommentsList`
Add scrolling comments into view on components list `mounted`
Add extra checks on `window` and `document` to avoid errors in ssr

### Issues
- fixes #1273 

### Todo
- [X] None


webapp/pages/post/_id/_slug/index.vue
```diff
@@ -193,7 +193,7 @@ export default {
 
     .ds-card-image {
       img {
-        max-height: 300px;
+        height: 300px;
```
mattwr18: code creep :fearful: 
I would prefer not having unrelated changes in `PR`s... we actually have talked, and there might even be an issue to remove the `max-height` altogether... enforcing a height of `300px` would be going in the opposite direction..... please revert this change, and let's make it in a separate ``PR`, please

webapp/components/FollowButton.vue
```diff
@@ -75,7 +75,7 @@ export default {
 
         const followedUser = follow ? data.followUser : data.unfollowUser
         this.$emit('update', followedUser)
-      } catch {
+      } catch (err) {
```
mattwr18: code creep :cold_sweat: 
I guess adding this doesn't hurt at all, even though we don't use `err`
and we are all certainly guilty of adding small changes we notice that have nothing to do with the ticket we are working on, but it's something we are trying to build a culture of not doing.

webapp/components/CommentList/CommentList.vue
```diff
@@ -47,6 +47,33 @@ export default {
...
+  watch: {
+    $route(to, from) {
+      // scroll inside the same page
+      if (to.hash === '#comments') {
+        this.scrollCommentsIntoView()
+      }
+    },
+  },
+  mounted() {
+    if (this.$route.hash === '#comments') {
```
mattwr18: do you have an example of when this would happen?
if someone sends another user a deep link with `#comments`?

webapp/pages/post/_id/_slug/index.vue
vbelolapotkov: @mattwr18 I wouldn't say this is a code creep. My bad of not providing the note on the intent of this change. So here it is. 

When you navigate to the `<post_url>#comments` from `<post_url>/more-info` there is a race condition takes place between loading post image and scrolling comments into view.
If the image is loaded after scrolling is finished it will push comments section 300px down which looks weird especially in mobile.

So fixing the image height solves this issue. 
I understand that this can affect the posts with images less than 300px in height but I doubt anybody uploads such a small post hero images. FYI, I've checked that posts without images are not affected.

Probably, a better solution would be using a placeholder image while the original one is being loaded.

webapp/components/FollowButton.vue
vbelolapotkov: @mattwr18 I agree this is not the part of the ticket. I understand and support the intention of building such a culture. So I need your advice here on how to properly solve this.

Here is the problem I've encountered.
The PR failed and I wasn't able to check in github (at least I haven't found any detailed report) which exactly tests failed. So I ran all the tests locally and fixed the failed ones :)

IMO this one micro change should be merged into master before the rest of PR.

P.s. I'm surprised how other PRs are not failing.

webapp/components/CommentList/CommentList.vue
vbelolapotkov: @mattwr18 yes, here is an example go to `More info` then to `Comments` (as described in #1273 ). Deep link is the case as well.

webapp/pages/post/_id/_slug/index.vue
mattwr18: Thank you for your explanation @vbelolapotkov... what would happen if we then implemented the no max-height of an image, which has been requested? the scrolling functionality would break?
Or we'd have to follow your suggestion of adding a placeholder image with a fixed height of 300px?

webapp/components/FollowButton.vue
mattwr18: Are you not able to access the builds in Travis? I either click on the link in the `PR` itself, or go to https://travis-ci.com/Human-Connection/Human-Connection/builds and sign in with my GitHub account and then I can see all the builds and investigate why it failed...
I think we can just leave the change here, I'm happy you understand and support the intention, I'm certainly guilty of code creep from time to time so feel free to call me out on it. I think just raising the conversation will help us in the long run.

webapp/components/CommentList/CommentList.vue
mattwr18: Thank you, I'll test that out :)

vbelolapotkov:
@Tirokk are you sure that screenshot is right?

webapp/mixins/scrollToAnchor.js
```diff
@@ -0,0 +1,20 @@
+import { scrollTo } from 'vue-scrollto'
+
+export default {
+  watch: {
+    $route(to, from) {
+      const anchor = to && to.hash
+      if (!this.checkAnchor(anchor)) return
+      setTimeout(() => {
+        scrollTo(anchor, 1000)
```
vbelolapotkov: @roschaefer IMO it's a good idea to add `vue-scrollto` as a module to 'nuxt.config.js' as described [here](https://github.com/rigor789/vue-scrollto#nuxtjs).

I'd pass default options:
```
{
  duration: 1000,
  offset: -100, // to compensate fixed navbar height
}
```

webapp/mixins/scrollToAnchor.js
```diff
@@ -0,0 +1,20 @@
+import { scrollTo } from 'vue-scrollto'
```
vbelolapotkov: @roschaefer I like the module and how it works. 👍

webapp/mixins/scrollToAnchor.js
```diff
@@ -0,0 +1,20 @@
...
+      const anchor = to && to.hash
+      if (!this.checkAnchor(anchor)) return
+      setTimeout(() => {
+        scrollTo(anchor, 1000)
+      }, 250)
+    },
+  },
+  mounted() {
+    const anchor = this.$route && this.$route.hash
+    if (!this.checkAnchor(anchor)) return
```
vbelolapotkov: @roschaefer how about some default behaviour when `checkAnchor` method is not defined in the component?

webapp/mixins/scrollToAnchor.js
mattwr18: From my limited experience working with `nuxt.scrollBehavior`, what I observed was that it is not called if you are on the same page, only on page transitions, so maybe the weird jump is not related to that??
https://router.vuejs.org/guide/advanced/scroll-behavior.html#scroll-behavior

vbelolapotkov:
@roschaefer perfect!

awesome

roschaefer:
@vbelolapotkov alright, thank you for your work, the solution and your reviews on this PR! :heart:

PR #1705, opened 2019-09-24
feat: 🍰 Set up Vue-Apollo Subscriptions

### LINKED ISSUE #1703: 🚀 [Feature] Add Vue-Apollo Subscriptions
## :rocket: Feature

We should set up `Subscriptions` for our site.
https://vue-apollo.netlify.com/guide/apollo/subscriptions.html#setup
With `Subscriptions`, we can update our `NewsFeed`, `NotificationsMenu`, etc... when there are changes

mattwr18:
### Issues

- fixes #1703 

### Todo

- [x] Test that server-side works
- [x] Set it up from the client side
- [x] Subscribe to `notifications` query
- [ ] Set it up as `wss` for safety


mattwr18:
@roschaefer do you want to have a look at this, maybe together, or should we just close this?

mattwr18:
this is ready for a `WIP` PR review @Human-Connection/hc-dev-team ...
maybe it is mergeable, but I noticed this comment on the `graphql-subscriptions` github page
https://github.com/apollographql/graphql-subscriptions
```
Note that the default PubSub implementation is intended for demo purposes. It only works if you have a single instance of your server and doesn't scale beyond a couple of connections. For production usage you'll want to use one of the PubSub implementations backed by an external store. (e.g. Redis)
```

It certainly seems to say that we shouldn't use this for production. Should we set it up with `Redis` then? what do you think @roschaefer ??

mattwr18:
Not nothing will happen, it will fallback to using the `PubSub` "for demo purposes" implementation.
Yes, it would need to be deployed with the `redis` helm chart.
I need to rebase this, and question I have is whether we want to commit fully to `Helm` and introduce the `redis` helm chart at the same time.
For me, this is not necessary because we are not maintaining the `helm` chart, so it is the same as using a `helm` chart for `cert-manager` or `grafana`, which is separate from our `human-connection` helm chart. 
What do you think?

PR #1756, opened 2019-09-29
1273 fix post page nav suggestions

roschaefer:
## 🍰 Pullrequest
@vbelolapotkov I was not sure if you mind when I push into your PR #1701 so I created this PR into your PR.
 
See my commit messages for explanations.


webapp/mixins/scrollToAnchor.js
```diff
@@ -0,0 +1,32 @@
...
+      window.scroll({
+        top,
+        left: 0,
+        behavior: 'smooth',
+      })
+    }, 250)
+  }
+}
+
+export default {
```
vbelolapotkov: @roschaefer there is a small flaw in this mixin. With the current implementation, `scrollToAnchor` will be called by each component on the page no matter what is the value of the `hash`. In our case with 10 comments on the page and `hash = '#comments'` we'll try to scroll 11 times. 
Please see suggestion below on how it can be solved. Not sure though if any of the app patterns are violated with that approach,

```suggestion
export default function(checkAnchor = () => true) {
  return {
    watch: {
      $route(to, from) {
        const anchor = to && to.hash
        checkAnchor.call(this, anchor) && scrollToAnchor(anchor)
      },
    },
    mounted() {
      const anchor = this.$route && this.$route.hash
      checkAnchor.call(this, anchor) && scrollToAnchor(anchor)
    },
  }
}
```

And to consume the mixin we'd use something like:
```javascript
// in the component
...
mixins: [scrollToAnchor(anchor => anchor === '#comments')
```

vbelolapotkov:
1273 fix post page nav suggestions:
- remove scrollBehavior from nuxt.config.js
- extract scroll to anchor behaviour into reusable mixin

PR #1757, opened 2019-09-29
fix email middleware transport config

vbelolapotkov:
## 🍰 Pullrequest
Set `ignoreTLS` option to true only when env variable `SMTP_IGNORE_TLS` is set to true.
Any other value will set the option to false.

### Issues
- None

### Todo
- [X] None


roschaefer:
thank you @vbelolapotkov :heart:

PR #1786, opened 2019-10-01
Update to 0.1.2

mattwr18:
Version bump


mattwr18:
I'm not sure why the `CHANGELOG` is merged... it's how the tool `github_changelog_generator` works, I guess.. I would prefer it the way you suggest as well, makes it much easier to see what was changed when... I'll have a look at the docs and see if there are flags that give us the desired output...

roschaefer:
@mattwr18 
```
rm CHANGELOG.md
git-generate-changelog --some --params --since-tag 0.1.1
```
Maybe?

mattwr18:
I could give it a go @roschaefer

mattwr18:
tried that out, and it gave the same output as before @roschaefer

mattwr18:
ok, I updated the release with the diff... but I guess I need to generate this directly before deploying the latest changes, not the day before cause there were a few `PR`s merged in that happened after this was generated

PR #1787, opened 2019-10-01
Update neo4j to Enterprise edition

mattwr18:
- accept license agreement

first step to setting up online backups for our database

### DIFF
--- a/neo4j/Dockerfile
+++ b/neo4j/Dockerfile
@@ -1,4 +1,4 @@
-FROM neo4j:3.5.11
+FROM neo4j:3.5.11-enterprise
 LABEL Description="Neo4J database of the Social Network Human-Connection.org with preinstalled database constraints and indices" Vendor="Human Connection gGmbH" Version="0.0.1" Maintainer="Human Connection gGmbH (developer@human-connection.org)"
 
 ARG BUILD_COMMIT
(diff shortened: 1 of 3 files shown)


neo4j/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM neo4j:3.5.11
+FROM neo4j:3.5.11-enterprise
```
roschaefer: But that would mean that people cannot run neo4j without license keys anymore? :thinking:

mattwr18: well, apparently they don't check the keys, you just have to agree to the license
see https://github.com/Human-Connection/Human-Connection/pull/1787/files#diff-4e5e90c6228fd48698d074241c2ba760R60

but also, we, since we are deploying to production, are agreeing to the commercial license and must provide the license number if asked, whereas others who are using it to help develop the site, would be agreeing to the developer's license https://neo4j.com/licensing/

but yeah, I think if people were wanting to set it up in different places/contexts, they might want to change this to the non-enterprise version

roschaefer: Seriously? It just works out of the box and we don't have to setup our license keys or anything like that?

mattwr18: it works with local docker dev... like if you run `docker-compose up` and go to localhost:7474 and log in, you can see in the bottom left that it says you are using `Enterprise` edition

PR #1788, opened 2019-10-01
Update Post query for blockedByUsers

### LINKED ISSUE #1663: 🐛 [Bug] Moderators and Admins can be blocked
## :bug: Bugreport

As a normal user I can block admins and moderators. But: If contributions get reported it must be possible for admins and moderators to see these. This is currently not the case. 

1. Write a post
2. Block the moderator
3. Somebody other reports the post
4. The post can not be seen/moderated by the moderator

### Expected behavior

If somebody blocks the moderator it may be generally ok. But if the moderator calls the post to be moderated the post must be visible for him.

### Version & Environment
 Type: [Desktop]      
 - OS: [Linux]      
 - Browser: [Firefox] 
 - Version [69]  
 - Device: [Notebook]  

### Additional context

There was a suggestion of Dennis, that a blocked user always can see any article, but just can't comment on them. Such a change would fix also this problem here. Because it never would be the way, that somebody can't see something just because being blocked.

mattwr18:
- Admin/Moderators see Posts even if they are blocked

### Issues

- fixes #1663 

### Todo

- [X] None

### DIFF
--- a/backend/src/schema/resolvers/users.js
+++ b/backend/src/schema/resolvers/users.js
@@ -22,6 +22,7 @@ export const getBlockedUsers = async context => {
 }
 
 export const getBlockedByUsers = async context => {
+  if (context.user.role === 'moderator' || context.user.role === 'admin') return []
   const { neode } = context
   const userModel = neode.model('User')
   let blockedByUsers = neode
@@ -30,7 +31,6 @@ export const getBlockedByUsers = async context => {
     .relationship(userModel.relationships().get('blocked'))
     .to('blocked', userModel)
     .where('blocked.id', context.user.id)
-    .where('blocked.role', 'user')
     .return('user')
   blockedByUsers = await blockedByUsers.execute()
   blockedByUsers = blockedByUsers.records.map(r => r.get('user').properties)
--- a/backend/src/schema/resolvers/users.js
+++ b/backend/src/schema/resolvers/users.js
@@ -30,6 +30,7 @@ export const getBlockedByUsers = async context => {
     .relationship(userModel.relationships().get('blocked'))
     .to('blocked', userModel)
     .where('blocked.id', context.user.id)
+    .where('blocked.user', 'user')
     .return('user')
   blockedByUsers = await blockedByUsers.execute()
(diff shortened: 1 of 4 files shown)


roschaefer:
Is this work-in-progress? I'm suprised this passes the build. I thought the `Post` query filters blocked content also for moderators or admins.

mattwr18:
It uses a `neode` query builder to filter out the `badIds`, but I added a line to the query builder to say that if you blocked a user, then filter it out, if not, show the posts for users who've blocked moderators or admins. https://github.com/Human-Connection/Human-Connection/pull/1788/files#diff-d7b7d7a5edf6685a8caa82f345180431R33

I did not add the same for users the moderator or admin has blocked, but, in my opinion, moderators and/or admins should not block users, otherwise it makes it more difficult to moderate the site

backend/src/schema/resolvers/users.js
```diff
@@ -30,6 +30,7 @@ export const getBlockedByUsers = async context => {
     .relationship(userModel.relationships().get('blocked'))
     .to('blocked', userModel)
     .where('blocked.id', context.user.id)
+    .where('blocked.role', 'user')
```
roschaefer: Ah, **that's** the line! Now I get it.

roschaefer: @mattwr18 I would prefer you could add a guard clause and return empty array if `context.user.role > 'user'` or something like that. It took me a while to understand that the additional `where('blocked.role')` is excluding *your own user node* from the result set, thus returning empty array. We could make this more explicit by returning empty array for this case.

roschaefer: A guard clause would also save us one unnecessary database call.

backend/src/schema/resolvers/users.js
```diff
@@ -22,6 +22,7 @@ export const getBlockedUsers = async context => {
 }
 
 export const getBlockedByUsers = async context => {
+  if (context.user.role === 'moderator' || context.user.role === 'admin') return []
```
roschaefer: We should probably export that check in `permissionsMiddleware` in the future

PR #1796, opened 2019-10-02
Max aspect ratio of 1:1 for Post index page

mattwr18:
- we are currently enforcing a 1:1 max ratio on the root path and would
like to maintain consistency

### DIFF
--- a/webapp/pages/post/_id/_slug/index.vue
+++ b/webapp/pages/post/_id/_slug/index.vue
@@ -197,7 +197,7 @@ export default {
 
     .ds-card-image {
       img {
-        height: 300px;
+        max-height: 710px;
         object-fit: cover;
         object-position: center;
       }


webapp/pages/post/_id/_slug/index.vue
```diff
@@ -197,7 +197,7 @@ export default {
 
     .ds-card-image {
       img {
-        height: 300px;
+        max-height: 710px;
```
roschaefer: Has this not been changed by @vbelolapotkov for a some reason? Code creep? :laughing:

mattwr18: it was changed by his `PR` yes... but as I pointed out to him, the reason I didn't like the change is cause we would need to change it when we deal with this issue. There is simply no way to enforce a 1:1 aspect ration with a fixed width unless we also limit the height.
I did double check the scroll functionality to check if it were broken in the way he described and it wasn't, but a lot changed on that `PR` before it was merged.
hehehe code creep is when you change code that doesn't have anything to do with the ticket, as I explained above... this is the _only_ change relevant to the issue :stuck_out_tongue:

mattwr18: Though, I think 710px is not actually a `1:1` ratio, since the `Post` page has a width of `910px`, I guess

mattwr18: As discussed in the issue about image cropping though, if we are inflexible with the width of the `Post`, then enforcing a max `1:1` means that the user would need to scroll to see anything other than the `Post` image

PR #1799, opened 2019-10-02
update lastActiveAt

### LINKED ISSUE #1305: 🚀 [Feature] Update `lastActiveAt` on every JWT token check
## :rocket: Feature
```gherkin
As an administrator
I would like to have a `lastActiveAt` attribute on every user account
In order to know how many users have been active at least once in the last month
```

For every authenticated action, the server has to check if the user is present in the database and is neither `deleted` or `disabled`. On these occasions we could also update a particular `lastActiveAt` attribute for analytical purposes. Having this information we could write cypher statements to check how many percent of the user base have been active in the last month, the last week, etc.

### Design & Layout
As a start, no user interface is required. Administrators have access to the database and could execute cypher statements to get the relevant data.

### Additional context
Here is the relevant code: https://github.com/Human-Connection/Human-Connection/blob/master/backend/src/jwt/decode.js#L15

Probably you just have to change:
```cypher
    MATCH (user:User {id: {id} })
    RETURN user {.id, .slug, .name, .avatar, .email, .role, .disabled, .actorId}
    LIMIT 1
```
to:
```cypher
    MATCH (user:User {id: {id} })
    SET user.lastActiveAt = datetime()
    RETURN user {.id, .slug, .name, .avatar, .email, .role, .disabled, .actorId}
    LIMIT 1
```

iylim:
### Issues

- None

### Todo

- [X] None

### DIFF
--- a/backend/src/jwt/decode.js
+++ b/backend/src/jwt/decode.js
@@ -14,7 +14,7 @@ export default async (driver, authorizationHeader) => {
   const session = driver.session()
   const query = `
     MATCH (user:User {id: $id, deleted: false, disabled: false })
-    SET user.lastActiveAt = datetime()
+    SET user.lastActiveAt = toString(datetime())
     RETURN user {.id, .slug, .name, .avatar, .email, .role, .disabled, .actorId}
     LIMIT 1
   `
--- a/backend/src/jwt/decode.js
+++ b/backend/src/jwt/decode.js
@@ -14,6 +14,7 @@ export default async (driver, authorizationHeader) => {
   const session = driver.session()
   const query = `
     MATCH (user:User {id: $id, deleted: false, disabled: false })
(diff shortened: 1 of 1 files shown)


backend/src/jwt/decode.js
```diff
@@ -14,6 +14,7 @@ export default async (driver, authorizationHeader) => {
   const session = driver.session()
   const query = `
     MATCH (user:User {id: $id, deleted: false, disabled: false })
+    SET user.lastActiveAt = datetime()
```
mattwr18: Actually, can you convert this `datetime()` to a string, at the moment we are doing this cause of a bug with one of the packages we are using for queries 
```suggestion
    SET user.lastActiveAt = toString(datetime())
```

roschaefer:
closed in favour of #1809

PR #1805, opened 2019-10-03
Update profile posts apollo query

### LINKED ISSUE #1754: 🐛 [Bug] Order contributions chronologically on profile page
## :bug: Bugreport

Posts are not ordered chronologically on the profile page. Applies to all three views: posts, commented and recommended.

1. Check the order of posts your profile page.

hc-posts-on-profile-page-1
hc-posts-on-profile-page-2
hc-posts-on-profile-page-3

### Expected behavior

All posts on the profile page should be ordered in chronological order.

### Version & Environment
 Type: Desktop      
 - OS: MacOS      
 - Browser: Firefox 
 - Version 69.0.1  
 - Device: MacBook

mattwr18:
- Make fetchMore orderBy equal to smart queries orderBy, 
- Make seed data more realistic,
- Avoid showing load more when there are not posts to load(less than 6 posts)

### Issues

- fixes #1754

### DIFF
--- a/backend/src/seed/seed-db.js
+++ b/backend/src/seed/seed-db.js
@@ -685,7 +685,7 @@ import { gql } from '../jest/helpers'
         return factory.create('User')
       }),
     )
-    
+
     await Promise.all(
       [...Array(30).keys()].map(() => {
         return factory.create('Post', {
--- a/backend/src/seed/seed-db.js
+++ b/backend/src/seed/seed-db.js
@@ -694,14 +694,14 @@ import { gql } from '../jest/helpers'
         })
       }),
     )
-    
+
     await Promise.all(
       [...Array(6).keys()].map(() => {
         return factory.create('Comment', {
           author: jennyRostock,
           postId: 'p2',
         })
-      })
+      }),
     )
 
     await Promise.all(
@@ -710,7 +710,7 @@ import { gql } from '../jest/helpers'
           author: jennyRostock,
           postId: 'p15',
         })
-      })
+      }),
     )
 
     await Promise.all(
@@ -719,7 +719,7 @@ import { gql } from '../jest/helpers'
           author: jennyRostock,
           postId: 'p4',
         })
-      })
+      }),
     )
 
     await Promise.all(
@@ -730,14 +730,14 @@ import { gql } from '../jest/helpers'
         })
       }),
     )
-    
+
     await Promise.all(
       [...Array(3).keys()].map(() => {
         return factory.create('Comment', {
           author: peterLustig,
           postId: 'p4',
         })
-      })
+      }),
     )
 
     await Promise.all(
@@ -746,7 +746,7 @@ import { gql } from '../jest/helpers'
           author: peterLustig,
           postId: 'p14',
         })
-      })
+      }),
     )
 
     await Promise.all(
@@ -755,7 +755,7 @@ import { gql } from '../jest/helpers'
           author: peterLustig,
           postId: 'p0',
         })
-      })
+      }),
     )
 
     await Promise.all(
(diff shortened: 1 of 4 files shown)


backend/src/seed/seed-db.js
```diff
@@ -106,27 +104,27 @@ import { gql } from '../jest/helpers'
...
-      f.create('Badge', {
+      factory.create('Badge', {
         id: 'indiegogo_en_rabbit',
         icon: '/img/badges/indiegogo_en_rabbit.svg',
       }),
-      f.create('Badge', {
+      factory.create('Badge', {
         id: 'indiegogo_en_wolf',
-        icon: '/img/badges/indiegogo_en_wolf.svg',
+        icon: '/img/badges/indiegogo_en_wolfactory.svg',
```
roschaefer: @mattwr18 what is a `wolfactory`? :wolf: :factory: 

:laughing:

webapp/pages/index.vue
```diff
@@ -50,7 +50,7 @@
       :infinite-scroll-throttle-delay="800"
       :infinite-scroll-immediate-check="true"
     >
-      <hc-load-more v-if="true" :loading="$apollo.loading" @click="showMoreContributions" />
```
roschaefer: What the heck? :astonished: :laughing:

webapp/pages/index.vue
```diff
@@ -151,7 +151,7 @@ export default {
           orderBy: this.sorting,
         },
         updateQuery: (previousResult, { fetchMoreResult }) => {
-          if (!fetchMoreResult || fetchMoreResult.Post.length < this.pageSize) {
+          if (!fetchMoreResult || fetchMoreResult.Post.length <= this.pageSize) {
```
roschaefer: I disagree. Most of the time, if there is more stuff to fetch, the length of `fetchMoreResult` is the same as `pageSize`. We have the edge case, that if your last page has the same number of items as `pageSize`, the load button is there when it shouldn't.

backend/src/seed/seed-db.js
mattwr18: haahahah global find and replace :rofl:

webapp/pages/index.vue
mattwr18: yeah, hehe `v-if="true"`

webapp/pages/index.vue
mattwr18: yeah, was a bit off on this logic hahah

webapp/pages/profile/_id/_slug.vue
```diff
@@ -375,14 +375,14 @@ export default {
...
-
       this.offset += this.pageSize
+
       PostQuery.fetchMore({
         variables: {
           offset: this.offset,
           filter: this.filter,
           first: this.pageSize,
-          orderBy: this.sorting,
+          orderBy: 'createdAt_desc',
```
roschaefer: Ouch! So `this.sorting` was always null? Is this the bug?

webapp/pages/profile/_id/_slug.vue
```diff
@@ -251,7 +251,7 @@
           </template>
         </masonry-grid>
         <div
-          v-if="hasMore"
+          v-if="hasMore && posts.length >= pageSize"
```
roschaefer: Is this the bug? Was the `hasMore` not enough? I can see that this condition will be `false` initially. But how relates that to the bug?

webapp/pages/profile/_id/_slug.vue
mattwr18: I feel like this could have been related to the bug, especially since I haven't been able to reproduce it after I changed this, but we would need to test it further, I guess... maybe on `nitro-staging`?

webapp/pages/index.vue
Tirokk: Oh may good 😬

PR #1806, opened 2019-10-03
fix invites count calculation for admin dashboard

### LINKED ISSUE #1633: 🐛 [Bug] Admin statistics entries are missing
## :bug: Bugreport

Admin statistics entries are missing. And invites seems wrong.

1. login in as an admin
2. go to the menus `Admin` item
3. there you see the problem

Bildschirmfoto 2019-09-18 um 12 40 06

And invites can not be zero, because I just before invited somebody. And we know there should be already invites imported.

Bildschirmfoto 2019-09-18 um 18 00 58

### Expected behavior

The missing statistics entries should be visible.

vbelolapotkov:
## 🍰 Pullrequest
This is a workaround solution for counting invites for admin dashboard suggested by @roschaefer in #1633 

### Issues
- fixes #1633 

### Todo
- [X] None

### DIFF
--- a/backend/src/schema/resolvers/statistics.js
+++ b/backend/src/schema/resolvers/statistics.js
@@ -9,7 +9,7 @@ export default {
           countPosts: 'Post',
           countComments: 'Comment',
           countNotifications: 'NOTIFIED',
-          countInvites: 'InvitationCode',
+          countEmails: 'EmailAddress',
           countFollows: 'FOLLOWS',
           countShouts: 'SHOUTED',
         }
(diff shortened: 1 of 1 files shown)


backend/src/schema/resolvers/statistics.js
```diff
@@ -28,6 +28,11 @@ export default {
           const stat = statistics[mapping[key]]
           response[key] = stat ? stat.toNumber() : 0
         })
+
+        /*
+         * Note: invites count is calculated this way because invitation codes are not in use yet
```
roschaefer: We learned the day before yesterday that we open registration for non-donators in two weeks from now. So, Invitation codes will not be necessary anymore. (Just as a side note)

vbelolapotkov: Thanks, good to know about invitation codes.

PR #1808, opened 2019-10-03
Display user email for administrators

### LINKED ISSUE #1704: 🚀 [Feature] Make Email of a User visible for Moderators
## :rocket: Feature

A moderator needs to contact people. For now he can't. Therfore in the admin panels "Users" lookup we need this email address ether (and best) as a small icon which when be clicked it is copied to the clipboard if possible or with a hover.

### Design & Layout

The emails of users should **not** be directly listed, because of privacy.
user-lookup

### Additional context

As long as we do moderation by ourselfes this is ok. Later, if community moderation comes, we need ether an NDA for everybody or we need another method of communication

aonomike:
## 🍰 Pullrequest

- Display user email to administrators
- Refactor Permission middleware tests
- Add tests to check email of user only revealed to admin and owner

### Issues

fixes #1704 

### Images

Screen Shot 2019-10-05 at 5 29 31 PM

### DIFF
--- a/webapp/pages/admin/users.vue
+++ b/webapp/pages/admin/users.vue
@@ -33,6 +33,16 @@
             <b>{{ scope.row.name | truncate(20) }}</b>
           </nuxt-link>
         </template>
+        <template slot="email" slot-scope="scope">
+          <nuxt-link
+            :to="{
+              name: 'profile-id-slug',
+              params: { id: scope.row.id, slug: scope.row.slug },
+            }"
+          >
+            <b>{{ scope.row.email }}</b>
+          </nuxt-link>
+        </template>
         <template slot="slug" slot-scope="scope">
           <nuxt-link
             :to="{
@@ -128,6 +138,7 @@ export default {
               id
               name
               slug
+              email
               role
               createdAt
               contributionsCount
--- a/webapp/pages/admin/users.vue
+++ b/webapp/pages/admin/users.vue
@@ -34,14 +34,9 @@
           </nuxt-link>
         </template>
         <template slot="email" slot-scope="scope">
-          <nuxt-link
-            :to="{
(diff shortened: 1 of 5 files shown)


webapp/pages/admin/users.vue
```diff
@@ -128,6 +138,7 @@ export default {
               id
               name
               slug
+              email
```
roschaefer: At the moment, the backend does not allow anybody except the owner to see the `email` of the user. You will need to change permissions in the backend.

I think I would still feel safe, if only the owner *or* an admin user can see the email.

aonomike: Done, thanks @roschaefer

aonomike:
Screen Shot 2019-10-05 at 5 29 31 PM

webapp/pages/admin/users.vue
```diff
@@ -33,6 +33,11 @@
             <b>{{ scope.row.name | truncate(20) }}</b>
           </nuxt-link>
         </template>
+        <template slot="email" slot-scope="scope">
+          <a :href="`mailto:${scope.row.email}`">
+            <b>{{ scope.row.email }}</b>
```
roschaefer: Some of those emails are quite long. Maybe we should think about how to fit the data into the layout? E.g. at some places in the app, we use `| truncate(20)` filter.

Although I'm OK to merge this and wait for our admins to complain about it :wink:

mattwr18: That's what I thought as well, @datenbrei mentioned that they might be copying and pasting the links, so maybe it's more difficult if it's truncated??

aonomike: Thanks @mattwr18 @roschaefer 
Let me know if I need to update this :D

PR #1814, opened 2019-10-04
Implement public registration

### LINKED ISSUE #1785: 🚀 [Feature] Register as a new User
## :rocket: Feature
~~~gerkhin
As an interested person
I'd like to register to the network
to use it
~~~
We need the following additional things (two of them not in the image):

- [ ] Dialogue, I'm more than 18 Years old
- [ ] I'm not a company but a natural person
- [ ] Some kind of simple capture (without Google)
- [ ] **Grayed out**: I join as a non-profit or sharity organization

### Design & Layout
register-new-user

### Additional context
Later we need to add a more complex capture and also some mechanism to prevent exsessive registrations

roschaefer:
## 🍰 Pullrequest
This will allow us to enable public registration with an environment variable `PUBLIC_REGISTRATION=true`. It can be disabled by setting it to `false`.

### Issues

- fix #1785 

### Todo

- [X] None


webapp/components/PasswordReset/Request.spec.js
```diff
@@ -18,7 +18,7 @@ describe('Request', () => {
         success: jest.fn(),
         error: jest.fn(),
       },
-      $t: jest.fn(),
+      $t: jest.fn(t => t),
```
mattwr18: we just ran into this issue today, on `ReportModal.spec.js`... would love to refactor all the `$t: jest.fn()` like this

webapp/components/PasswordReset/Request.vue
```diff
@@ -124,17 +102,9 @@ export default {
   },
 }
 </script>
-<style>
+<style scoped>
 .login-image {
   width: 90%;
   max-width: 200px;
 }
```
mattwr18: if this is not used anymore, we should remove it, although I would like to know the motive for removing the image from these forms... they seems a bit bland

webapp/components/Registration/CreateUserAccount.vue
```diff
@@ -1,117 +1,107 @@
...
-          src="/img/sign-up/nicetomeetyou.svg"
-        />
-      </ds-space>
-      <ds-space>
-        <ds-heading size="h3">
-          {{ $t('registration.create-user-account.title') }}
-        </ds-heading>
-      </ds-space>
+  <div v-if="response === 'success'">
+    <sweetalert-icon icon="success" />
```
mattwr18: it would be nice to wrap these in `<transition name="ds-transition-fade">` to keep it consistent across the code base

webapp/pages/password-reset.vue
```diff
@@ -20,3 +25,10 @@ export default {
   },
 }
 </script>
+<style lang="scss" scoped>
+img {
+  padding-left: 50px;
+  padding-right: 50px;
+  max-width: 200px;
+}
+</style>
```
mattwr18: same here, if there is no img, why is there css?

webapp/components/PasswordReset/Request.spec.js
roschaefer: I know.. we learn everyday. From now on I will always check if `wrapper.contains('some.translation.key')` instead of `expect(mocks.$t).toHaveBeenCalled('some.translation.key')`

webapp/components/PasswordReset/Request.vue
roschaefer: I did not really like the logo on the password reset form and would have preferred a "lock" icon or image that distinguishes the form from the signup form etc. And I think the motivation for removing it was that I had troubles to get the layout right.

PR #1840, opened 2019-10-08
Allow admins to pin a post

### LINKED ISSUE #1205: 🚀 [Feature] Create a pinned Post for administration
## :rocket: Feature
```gherkin
Feature: Pin a post
  As an admin
  I want to pin a post so that it always appears at the top
  In order to make sure all network users read it - e.g. notify people about security incidents, maintenance downtimes
```

### Design & Layout
Screenshot - 2019-10-17T005347 639

### Additional context
**Limitation of one pinned post maximum:**
We have evidence from the legacy alpha, that reaching all users is a feature that gets abused for marketing. On legacy alpha, it was the system notifications - originally intended to be rarely used for exceptional situations like planned downtimes - which got heavily abused for marketing purposes. Pinning a post is great for the admin but gives terrible user experience.

Therefore, we decided to go with a limitation of one pinned post max at the same time. That will leave enough space on the index page for the actual (desired) content and also create an incentive to think before you pin a post (because you have only one).

**Database:**
A `PINNED` relation with a property `pinnedAt` should point from this admin to all its pinned Posts. We could then order by `pinnedAt` on the post and not run into any performance issues as we do with "Most popular" or "Most commented" orderings. The `pinnedAt` would be `null` for all unpinned posts and they would come last.

mattwr18:
- at the moment, it is created as a property of Post, but will be
refactored to create a relationship[:PINNED] from the admin to the Post

- Co-authored-by: kachulio1 <jngugi88@gmail.com>

### Issues

- fixes #1205 

### Todo

- [x] Test-Drive the Refactor of backend to create a relationship [:PINNED] instead of adding it as an attribute of `Post` node
- [x] Add attributes to [:PINNED], like `createdAt`
- [x] Frontend implementation
- [x] Styling


backend/src/schema/resolvers/posts.spec.js
```diff
@@ -357,6 +374,78 @@ describe('CreatePost', () => {
...
+                  role: 'admin',
+                },
+              },
+            },
+            errors: undefined,
+          }
+
+          await expect(mutate({ mutation: createPostMutation, variables })).resolves.toMatchObject(
+            expected,
+          )
```
roschaefer: I would add a test case here to ensure we have only one pinned post at a time. E.g. when you pin a post and there was another pinned post already, the previously pinned post gets unpinnned.

backend/src/schema/resolvers/posts.js
```diff
@@ -29,10 +28,26 @@ const filterForBlockedUsers = async (params, context) => {
...
+    const filteredPostsArray = []
+    Object.keys(params.filter).forEach(key => {
+      filteredPostsArray.push({ [key]: params.filter[key] })
+    })
+    filteredPostsArray.push(pinnedPostFilter)
+    params.filter = { OR: filteredPostsArray }
+  }
+  return params
+}
+
```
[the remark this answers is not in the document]
mattwr18: tried to, but couldn't get it to work the other way

backend/src/schema/resolvers/posts.js
```diff
@@ -111,14 +128,17 @@ export default {
       return post
     },
     UpdatePost: async (_parent, params, context, _resolveInfo) => {
-      const { categoryIds } = params
```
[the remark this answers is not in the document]
mattwr18: done, thanks for the reminder, I think it's much cleaner

backend/src/schema/resolvers/posts.js
```diff
@@ -131,15 +151,41 @@ export default {
...
+        `
+
+        await session.run(cypherDeletePreviousRelations)
+
+        updatePostCypher += `
+          MATCH (user:User {id: $userId}) WHERE user.role = 'admin'
+          MERGE (user)-[:PINNED {createdAt: toString(datetime())}]->(post)
+          WITH post
+          `
+      }
```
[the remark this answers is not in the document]
mattwr18: done

backend/src/schema/resolvers/posts.js
```diff
@@ -211,10 +225,68 @@ export default {
...
     },
+    pinPost: async (_parent, params, context, _resolveInfo) => {
+      let pinnedPost
+      const { driver, user } = context
+      const session = driver.session()
+      const { id: userId } = user
+      let writeTxResultPromise = session.writeTransaction(async transaction => {
+        const deletePreviousRelationsResponse = await transaction.run(
+          `
+          MATCH ()-[previousRelations:PINNED]->(post:Post)
```
Tirokk: Why not `User`?
```suggestion
          MATCH (:User)-[previousRelations:PINNED]->(post:Post)
```

backend/src/schema/resolvers/posts.js
```diff
@@ -211,10 +225,68 @@ export default {
...
+          DELETE previousRelations
+          RETURN post
+        `,
+          { params },
```
Tirokk: `params` is not used here, as far I see …

backend/src/schema/resolvers/posts.js
```diff
@@ -211,10 +225,68 @@ export default {
...
+    pinPost: async (_parent, params, context, _resolveInfo) => {
+      let pinnedPost
+      const { driver, user } = context
+      const session = driver.session()
+      const { id: userId } = user
+      let writeTxResultPromise = session.writeTransaction(async transaction => {
+        const deletePreviousRelationsResponse = await transaction.run(
+          `
+          MATCH ()-[previousRelations:PINNED]->(post:Post)
+          DELETE previousRelations
```
Tirokk: What happens with the posts are not pinned anymore?
They appear as usual post?
That would be annoying, or?

backend/src/schema/resolvers/posts.js
```diff
@@ -211,10 +225,68 @@ export default {
...
+      }
+      return pinnedPost
+    },
+    unpinPost: async (_parent, params, context, _resolveInfo) => {
+      let unpinnedPost
+      const session = context.driver.session()
+      const writeTxResultPromise = session.writeTransaction(async transaction => {
+        const unpinPostTransactionResponse = await transaction.run(
+          `
+          MATCH ()-[previousRelations:PINNED]->(post:Post {id: $params.id})
```
Tirokk: Why not `User` here?
```suggestion
          MATCH (:User)-[previousRelations:PINNED]->(post:Post {id: $params.id})
```

backend/src/schema/resolvers/posts.spec.js
```diff
@@ -550,6 +562,361 @@ describe('UpdatePost', () => {
...
+        beforeEach(async () => {
+          await factory.create('Post', {
+            id: 'only-pinned-post',
+            author: admin,
+          })
+          await mutate({ mutation: pinPostMutation, variables })
+          variables = { ...variables, id: 'only-pinned-post' }
+          await mutate({ mutation: pinPostMutation, variables })
+          pinnedPost = await neode.cypher(
+            `MATCH ()-[relationship:PINNED]->(post:Post) RETURN post, relationship`,
```
Tirokk: Would love to have `User` here …
```suggestion
            `MATCH (:User)-[relationship:PINNED]->(post:Post) RETURN post, relationship`,
```

webapp/components/ContentMenu.vue
```diff
@@ -175,6 +200,7 @@ export default {
         },
       })
     },
+    pinPost() {},
```
Tirokk: This has no functionality and is superfluous!?

backend/src/schema/resolvers/posts.js
mattwr18: Cause it's not important what or who pinned it

backend/src/schema/resolvers/posts.js
mattwr18: Yes, what should happen with them? They are part of the network... Why should they be disabled or deleted? Or do you mean something else

backend/src/schema/resolvers/posts.js
mattwr18: same as :point_up_2:

webapp/components/ContentMenu.vue
mattwr18: good catch! thanks!

backend/src/schema/resolvers/posts.js
mattwr18: good catch, thanks

mattwr18:
@Tirokk I have no idea what this means `As menu symbol a pin and a striped pin would be great.`
do you have a link to the icon in the styleguide?

As for your comment `As the banderole color of the pinned post instead of blue I would suggest red or something like this and may be a pin symbol in front of it?`

I don't dislike your idea to differentiate the ribbon from the others, I didn't handle the styling @roschaefer did, and I'm waiting for @alina-beck to have a look since she has posted some ideas on the issue, which you might or might not have seen. I don't like the idea of using `red`, prefer the `yellow` @alina-beck suggested, but as I said, not too fussy.

I guess if we don't feel it's ready to merge, we can work on making it look different/adding more automated tests for it... up to you all.

backend/src/schema/resolvers/posts.js
aonomike: And how long does it take before an admin can unpin a post? I was thinking of a case that one admin has just pinned a post, and immediately after another admin pins a different post.

backend/src/schema/resolvers/posts.js
Tirokk: ???
Why then in file `backend/src/schema/types/type/Post.gql`, line 19:

```
pinnedAt: String @cypher(
      statement: "MATCH (this)<-[pinned:PINNED]-(:User) WHERE NOT this.deleted = true AND NOT this.disabled = true RETURN pinned.createdAt"
    )
```

In my eyes it is a kind of code quality if we are as good as specific we can be. That’s why we are writing test to nail the things more down, as an example.

For sure at the moment only users can pin posts. But that only users can do something is true  for a lot of things. Only users can emote to posts, but we query for the label `User` as well like in file "backend/src/schema/resolvers/posts.js“, line 315:
```
        emotionsCount: '<-[related:EMOTED]-(:User)',
```

mattwr18: Sure, I updated it already... 
I don't think it's important, but if you do, I'm happy to update it, as I showed with my actions.

backend/src/schema/resolvers/posts.js
mattwr18: You can pin another `Post` straight after @aonomike... We shouldn't have an issue like you described cause we have very few admins that will pin posts, and hopefully will not abuse pinning them.
If we do have an issue, we can address it.

webapp/locales/de.json
```diff
@@ -354,6 +354,7 @@
   },
   "post": {
     "name": "Beitrag",
+    "pinned": "Gepinnt",
```
Tirokk: Here I would prefer a more funktional description from the users perspective like `System info` or similar …
If the user reads `Pinned` or `Gepinnt` then they may think: Oh I haven’t pinned a post …
Or?

webapp/locales/en.json
```diff
@@ -355,6 +355,7 @@
   },
   "post": {
     "name": "Post",
+    "pinned": "Pinned",
```
Tirokk: Same same … 👆🏼

mattwr18:
we have this styling at the moment, but if anyone wants to change it, please open an issue and put in a `PR`, for now it is more than serviceable in our mind
Screenshot from 2019-10-21 13-56-54

PR #1853, opened 2019-10-09
Added createdAt date for follow and shout

drone-cloud:
### Issues

- None

### Todo

- [X] None


drone-cloud:
I need to look into why the build is failing.

backend/src/schema/resolvers/follow.spec.js
```diff
@@ -129,6 +129,16 @@ describe('follow', () => {
...
+
+      // Test to make sure FOLLOWS relationship has "createdAt" date.
+      const relation = await neode.cypher(
+        'MATCH (user:User {id: {id}})-[relationship:FOLLOWS]->(followed:User) WHERE relationship.createdAt IS NOT NULL RETURN relationship',
+        { id: 'u1' },
+      )
+      const relationshipProperties = relation.records.map(
+        record => record.get('relationship').properties.createdAt,
+      )
+      expect(relationshipProperties[0]).toEqual(expect.any(String))
```
mattwr18: can you please place this into it's own `it` block? 
we should avoid multiple `expect` in the same it block

roschaefer: Usually you would try to have one assertion per spec. You can easily create a new spec by saying:
```js
it('adds `createdAt` to `FOLLOW` relationship', async () => { ... })
```

Redundancy is not a problem with software tests.

backend/src/schema/resolvers/shout.spec.js
```diff
@@ -100,6 +100,18 @@ describe('shout and unshout posts', () => {
...
+          'MATCH (user:User {id: $userId1})-[relationship:SHOUTED]->(node {id: $userId2}) WHERE relationship.createdAt IS NOT NULL RETURN relationship',
+          {
+            userId1: 'current-user-id',
+            userId2: 'another-user-post-id',
+          },
+        )
+        const relationshipProperties = relation.records.map(
+          record => record.get('relationship').properties.createdAt,
+        )
+        expect(relationshipProperties[0]).toEqual(expect.any(String))
```
mattwr18: same as above... you can have an `it` block where you add a description for the test instead of a comment
then run `await mutate({ mutation: mutationShoutPost, variables })`
then this code you added

backend/src/schema/resolvers/follow.spec.js
mattwr18: you can run `await mutate({ mutation: mutationFollowUser, variables })` then this code

backend/src/schema/resolvers/shout.spec.js
roschaefer: Same as above :point_up: in this test case we already have two assertions - because of a weakness in the `shout` resolver: Instead of returning the shouted post, we return `Boolean`. That's why we have to send another graphql query just to make sure it really got shouted by the current user.

backend/src/schema/resolvers/follow.spec.js
drone-cloud: @mattwr18 -  Much appreciated your review. Implemented changes as you requested.

drone-cloud: @roschaefer I like your suggestion - "Redundancy is not a problem with software tests." I was thinking in terms of code.
Much appreciated your review. Implemented changes as you requested.

backend/src/schema/resolvers/shout.spec.js
drone-cloud: Thanks again for review @mattwr18 . it's done as well.

drone-cloud: Thanks @roschaefer for review. I made changes as requested.

drone-cloud:
> you are good
> 
> @drone-cloud this looks good! I give it a "Request Changes" code review though. If you want, we can merge it as it is, but I thought I give you the chance to improve your code before it gets into `master`.



> you are good
> 
> @drone-cloud this looks good! I give it a "Request Changes" code review though. If you want, we can merge it as it is, but I thought I give you the chance to improve your code before it gets into `master`.

Thank you immensely.

PR #1872, opened 2019-10-11
 🍰 Fix - maintaining sorting after navigation

### LINKED ISSUE #1733: 🐛 [Bug] Sorting is not maintained after page navigation
## :bug: Bugreport

Currently the user loses their sorting when they navigate to a `Post`, for example.

1. After logging in, you should be on the root path, if not navigate there by clicking on the Human Connection logo
2. click on the sorting dropdown in the top right of the news feed... it should read `Newest`
3. change the sorting to `Oldest`
4. after it has sorted, click on one of the Posts
5. navigate back to the index page with the browser's back button
6. notice the sorting changes back to `Newest`

### Expected behavior

A user's sorting option should be maintained till the change it, or a hard refresh
maybe, it should be persistent across sessions(?)
short fix, would be until a hard refresh or a different session, then another ticket could be opened if we want to support default sorting options

nimitbhargava:
## 🍰 Pullrequest
Restoring posts order even after navigation

### Issues
- fixes #1733


mattwr18:
hey @nimitbhargava, you might already have seen, but it's failing the build because of linting, and because we need to update some of the component tests to tell them about our `getters`, `mutations`... as you pointed out

roschaefer:
Hey @nimitbhargava sorry for taking over this PR. Last Friday my intention was to reduce the list of open PRs. As discussed in our daily standup I will do the following in the future:

* Write a comment in a PR before adding commits to it, so the author can respond
* Ask in the Discord if sb. knows the state of the PR, e.g. you paired with @mattwr18 and he could have told me that you are still keen to finish this PR - if I just should have asked in the Discord.

Issue #1708, opened 2019-09-24
🚀 [Feature] [Moderation] See reported Content with additional information


datenbrei:
## :rocket: Feature

If a user reported content and gave additional information (choose from list and free from field) then the moderator should have an extra button to see this information. 

It's up to the UI-Designer, if the category of tos-violation is shown in the list of reports or in an own (popup) window.

### Design & Layout

It's up to the UI-Designer, if the category of tos-violation is shown within the list of reports or in an own (popup) window. The free form message should be restricted in length 1000 chars? and maybe shown as a hover, if no own popup windows isused.

### Additional context

- [x] It’s fixed by PR #1797

PR #1878, opened 2019-10-11
🍰 Refactor Database for Reporting with specific information

Tirokk:
## 🍰 Pullrequest

Refactor database structure from `(resource)<-[:REPORTED]-(report)<-[:REPORTED]-(submitter:User)` to simpler `(resource)<-[:REPORTED]-(submitter:User)` structure.

### Issues

- relates #1707
- relates #1708
- and relative to PR #1797

### Todo

- [x] Cypher commands to convert the data of the database to this new structure. See file `deployment/legacy-migration/maintenance-worker/migration/neo4j/dbManipulations/20191011_changeReportNodeToRelationOnly.cql`

### Additional Informations

- Accidentally included a frontend change for adding the created at to the moderators report list to this PR.
- Removed the limit of the report list length. So we have to add pagination in the future. (talked about that with @datenbrei )


deployment/legacy-migration/maintenance-worker/migration/neo4j/dbManipulations/20191011_changeReportNodeToRelationOnly.cql
```diff
@@ -0,0 +1,7 @@
+MATCH (submitter:User)-[:REPORTED]->(report:Report)-[:REPORTED]->(resource)
+DETACH DELETE report
+CREATE (submitter)-[reported:REPORTED]->(resource)
+SET reported.createdAt = toString(datetime())
+SET reported.reasonCategory = 'other'
+SET reported.reasonDescription = '!!! Created automatically to ensure database consistency! Date-time is this creation date and time.'
+RETURN reported
```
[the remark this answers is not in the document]
mattwr18: this didn't work for me... I get a `Permission denied` error...
it works if I use `cypher-shell` directly, which is probably what I would do

backend/src/schema/resolvers/reports.js
```diff
@@ -87,6 +81,60 @@ export default {
           break
       }
 
+      return response
+    },
+  },
+  Query: {
+    reports: async (_parent, _params, { driver, _req, _user }, _resolveInfo) => {
```
mattwr18: if we are not using `req` and `user` there is no reason to destructure them and then put an `_`
this is a good practice for `_parent` cause it is the first parameter and we are not using it, but we don't have to destructure anything more than driver, if we are not using it

backend/src/schema/resolvers/reports.js
```diff
@@ -1,25 +1,18 @@
...
 export default {
   Mutation: {
-    report: async (
-      _parent,
-      { resourceId, reasonCategory, reasonDescription },
-      { driver, _req, user },
-      _resolveInfo,
-    ) => {
-      const reportId = uuid()
+    report: async (_parent, params, { driver, _req, user }, _resolveInfo) => {
```
mattwr18: same as below :point_down:

backend/src/schema/index.js
```diff
@@ -35,14 +35,14 @@ export default applyScalars(
...
             'Comment',
-            'Report',
             'Statistics',
             'LoggedInUser',
             'Location',
             'SocialMedia',
             'User',
             'EMOTED',
             'NOTIFIED',
+            'REPORTED',
```
mattwr18: I guess that you want to exclude the `REPORTED` query as well, otherwise we have one for `reports` and another for `REPORTED`

backend/src/schema/resolvers/reports.js
```diff
@@ -87,6 +81,60 @@ export default {
...
+      if (!dbResponse) return null
+
+      const response = []
+      dbResponse.forEach(ele => {
+        const { report, submitter, resource, type } = ele
+
+        const responseEle = {
+          ...report.properties,
+          resource: resource.properties,
+          resourceId: resource.properties.id,
```
mattwr18: why do we need `resource` and also `resourceId` if the resource has the id already?

mattwr18: why do we need `resource` at all actually, aren't we checking the type and returning the resources' properties based on which type it is?

backend/src/schema/resolvers/reports.spec.js
```diff
@@ -1,37 +1,55 @@
 import { GraphQLClient } from 'graphql-request'
```
mattwr18: would love to refactor this test to remove the use of `graphql-request`, but maybe that's not your priority at the moment... we could help with that, @aonomike is getting awesome at this... maybe in a separate `PR` though?

backend/src/schema/index.js
Tirokk: Yes, genau!

backend/src/schema/resolvers/reports.spec.js
Tirokk: You are right @mattwr18 . I set the priority to get this PR done. Would love if @aonomike jumps into that … 😍

deployment/legacy-migration/maintenance-worker/migration/neo4j/dbManipulations/20191011_changeReportNodeToRelationOnly.cql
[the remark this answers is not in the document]
Tirokk: I would love to have this minimum temporarily in the code base. For transparency and may be as a conversion example for later generations?

But for sure must not be in the legacy-migration. I couldn’t find quicke a better place in the short time … I like to ask and than forgot.
Do you have a suggestion where to put it? @mattwr18 

It only works in the cypher-shell. Correct. Where you have tried this out?

backend/src/schema/resolvers/reports.js
Tirokk: I forgot to delete the code here after I find out it can not be supported now … Thanks for the hint @mattwr18 
Done …

backend/src/schema/resolvers/reports.js
Tirokk: Yes, you are right. @mattwr18 
Done …

backend/src/schema/index.js
mattwr18: sorry, you marked as resolved, but didn't exclude `REPORTED` for queries... maybe by accident?

deployment/legacy-migration/maintenance-worker/migration/neo4j/dbManipulations/20191011_changeReportNodeToRelationOnly.cql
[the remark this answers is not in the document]
mattwr18: I tried running it from the command line, which can't be done since it requires `cypher-shell`

I think it would be good to convert it to a script, if we want it in the code base.
see `neo4j/db_setup.sh`, it's part of the database and it's manipulation, so it seems like a good place for it.
You can copy and paste a lot of the code from `neo4j/ds_setup.sh`, It is meant to be used once or twice, and it cannot be called directly from the remote server, so we'd still need to open the file locally and copy and paste it. I like that you linked to the example of where you found the way to do it, maybe we don't need an example in our code base, if there are docs with them?

backend/src/schema/index.js
Tirokk: I did locally not yet pushed up.
You mean to exclude it from the queries list, I suppose?

backend/src/schema/types/type/REPORTED.gql
```diff
@@ -0,0 +1,32 @@
+type REPORTED {
+  createdAt: String
+  reasonCategory: String
```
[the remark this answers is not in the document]
Tirokk: Yes, I forgot and did it locally yet. Push it up all together …

backend/src/schema/resolvers/reports.spec.js
```diff
@@ -287,3 +291,227 @@ describe('report', () => {
...
+    })
+
+    it('roll "user" gets no reports', async () => {
+      authenticatedUser = await user.toJson()
+      const { data, errors } = await query({ query: reportsQuery })
+      expect(data).toEqual({
+        reports: null,
+      })
+      expect(errors[0]).toHaveProperty('message', 'Not Authorised!')
+    })
```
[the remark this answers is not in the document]
Tirokk: Thanks for the nice hint 😍, I have refactored yet locally …

backend/src/schema/resolvers/reports.spec.js
```diff
@@ -287,3 +291,227 @@ describe('report', () => {
...
+        author: moderator,
+        id: 'p2',
+        categoryIds,
+        content: 'Already seen post mention',
+      }),
+      factory.create('Post', {
+        author: user,
+        id: 'p3',
+        categoryIds,
+        content: 'You have been mentioned in a post',
```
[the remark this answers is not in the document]
Tirokk: 😂 Yes, unreflected copy and paste …
Thanks for you pointing that out !!! 😘
Changed locally yet …

backend/src/schema/resolvers/reports.spec.js
```diff
@@ -1,37 +1,55 @@
...
+    `
+    client = new GraphQLClient(host, {
+      headers,
+    })
+    return client.request(reportMutation, variables)
+  }
+
   beforeEach(async () => {
-    returnedObject = '{ id }'
+    returnedObject = '{ createdAt }'
```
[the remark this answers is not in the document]
Tirokk: I have done locally …
Very good hint @mattwr18 !!! 😎

mattwr18:
@Tirokk, it was not translated because I sent a graphql request giving a category that does not exist as a translatable key... "something"

Tirokk:
Yeah, that makes sense @mattwr18 !

mattwr18:
but should have been addressed by using an enum, so shouldn't be an issue going forward, great @Tirokk

backend/src/schema/resolvers/reports.spec.js
```diff
@@ -20,7 +20,7 @@ describe('report mutation', () => {
 
   const action = () => {
     reportMutation = gql`
-      mutation($resourceId: ID!, $reasonCategory: String!, $reasonDescription: String!) {
+      mutation($resourceId: ID!, $reasonCategory: ReasonCategory!, $reasonDescription: String!) {
```
Tirokk: Thanks that you have seen this and the related ones. 👍🏼 I was obviously in a hurry …
Sorry for that!

mattwr18: no need to apologize, we got each others' backs

backend/src/schema/resolvers/reports.spec.js
```diff
@@ -480,7 +475,6 @@ describe('reports query', () => {
             submitter: expect.objectContaining({
               id: 'user1',
             }),
-            resourceId: 'auth1',
```
Tirokk: I was in a hurry and not concentrated … 😬

backend/src/middleware/validation/validationMiddleware.js
```diff
@@ -57,11 +57,40 @@ const validateUpdatePost = async (resolve, root, args, context, info) => {
...
+  return resolve(root, args, context, info)
+}
+
 export default {
   Mutation: {
     CreateComment: validateCommentCreation,
     UpdateComment: validateUpdateComment,
     CreatePost: validatePost,
     UpdatePost: validateUpdatePost,
+    report: validateReport,
```
Tirokk: Great done @mattwr18

backend/src/schema/resolvers/reports.spec.js
mattwr18: well, it works for queries, but didn't work for mutations, we could easily add it to the mutations as well, I just didn't feel like it's necessary, and didn't hear back from you so I just thought I'd make the change and revert if people wanted me to

backend/src/middleware/validation/validationMiddleware.js
mattwr18: glad you like it, I need to go back and do this for some of my `PR`s hahaha

Tirokk:
You made a wonderful work @mattwr18 and have seen some things I have forgotten or overseen.
Great thanks !!! 😎

Tests are all successful.

But I can not report anything anymore. Don’t know why.
And a moderator has an empty report list:

Bildschirmfoto 2019-10-14 um 21 40 45

mattwr18:
I also forgot to change the mutation in `webapp/graphl`, so that is probably why it failed the build...
did you re-seed the database?

neo4j/change_report_node_to_relationship.sh
```diff
@@ -0,0 +1,26 @@
...
+
+echo "
+MATCH (submitter:User)-[:REPORTED]->(report:Report)-[:REPORTED]->(resource)
+DETACH DELETE report
+CREATE (submitter)-[reported:REPORTED]->(resource)
+SET reported.createdAt = toString(datetime())
+SET reported.reasonCategory = 'other'
+SET reported.reasonDescription = '!!! Created automatically to ensure database consistency! createdAt is when the database manipulation happened.'
+RETURN reported;
+" | cypher-shell
```
roschaefer: Interesting approach!

webapp/pages/moderation/index.vue
```diff
@@ -82,6 +82,14 @@
           {{ scope.row.submitter.name }}
         </nuxt-link>
       </template>
+      <!-- createdAt -->
+      <template slot="createdAt" slot-scope="scope">
```
roschaefer: Seems to be a deprecated API: https://vuejs.org/v2/guide/components-slots.html#Scoped-Slots-with-the-slot-scope-Attribute

mattwr18:
@roschaefer merging as it's into a feature branch and I have changes you requested there, including these ones

Issue #1634, opened 2019-09-18
🐛 [Bug] Put the imprint page directly on the page again


Tirokk:
## :bug: Bugreport

Put the imprint page directly on the page again.
It was removed here #1615 …

### Steps to reproduce the behavior
1. login
2. click on the imprint link on the footer
3. you are directed to an external page

### Expected behavior

You should be directed to an internal page.

datenbrei:
Closed, see https://github.com/Human-Connection/Human-Connection/pull/1883

PR #1883, opened 2019-10-12
Imprint page is back in the system again

ogerly:
## 🍰 Pullrequest
 Imprint page is back in the system again

### Issues
https://github.com/Human-Connection/Human-Connection/issues/1634
 
- None

### DIFF
--- a/webapp/components/PageFooter/PageFooter.vue
+++ b/webapp/components/PageFooter/PageFooter.vue
@@ -2,9 +2,7 @@
   <div id="footer" class="ds-footer">
     <a href="https://human-connection.org" target="_blank" v-html="$t('site.made')"></a>
     <span>-</span>
-    <a href="https://human-connection.org/impressum/" target="_blank">
-      {{ $t('site.imprint') }}
-    </a>
+    <nuxt-link to="/imprint">{{ $t('site.imprint') }}</nuxt-link>
     <span>-</span>
     <nuxt-link to="/terms-and-conditions">{{ $t('site.termsAndConditions') }}</nuxt-link>
     <span>-</span>
--- a/webapp/nuxt.config.js
+++ b/webapp/nuxt.config.js
@@ -46,6 +46,7 @@ export default {
       'registration-enter-nonce',
       'registration-create-user-account',
       'pages-slug',
+      'imprint',
       'terms-and-conditions',
       'code-of-conduct',
       'changelog',
--- a/webapp/pages/imprint.vue
+++ b/webapp/pages/imprint.vue
@@ -1,0 +1,83 @@
+<template>
+  <div>
+    <ds-space>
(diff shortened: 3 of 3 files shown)


datenbrei:
There is no legal reason to have the imprint on the same website or domain. For legal advice see here (German): https://www.it-recht-kanzlei.de/Thema/impressum-tmg.html?page=1#sect_3

Also the text of the imprint should not be part of the software repository at all, because it is highly specific to the organization, in this case specific to Human Connection.

I once decided about this, so I'll close this issue.

Later there could and should be two things: Adaptable basic HTML pages and a freely configurable footer, which also may be implemented as a free to fill out one line HTML field in the Admin panel, so that it can be easily adapted to the whishes of the specific company.

ogerly:
there must be a planning with the issues ... i only waste time!

Tirokk:
What I do not understand @datenbrei :
We discussed that in a meeting with you and @roschaefer said that the imprint has to be on the page directly and we agreed to put the imprint back on the page.
So as @ogerly asked me what this issue is about, that exists since weeks, I explained him.

I get in a role where I suggest issues to other programmers and they are superfluid or they are silently already picked up by other devs.
That makes no fun and I like to know where I have to send the questioners for issues that we get less frustrated?

datenbrei:
I was told by @ogerly that @roschaefer had complaints about the original decision to have the imprint as a link to our homepage. I heard about that so I think, I shouldn't have been part of such a meeting or decision. Well, though, I remember, that there had been complaints about that. If I would have been part of the meeting I would not have been astonished to heae so.

Anyways, you are right, we should document these decisions, because I have the same feelings: Having things discussed, where later other imlementations are done, best example: The pinned post (where Robert had not been part of that discussion)

So my suggestion would be: Let us just document the decisions with the corresponding issue immediatly, when they have been done.

mattwr18:
That's not the way I remember it @datenbrei... @roschaefer for sure was part of the `Monday` standup where we talked about not limiting pinned posts, and we knew that he had voiced his opposition to that in the past, but in that meeting _did_ not repeat his concerns, or let us know that he felt so strongly about them. It might be that we didn't give proper time or make sure that he was given space to voice them, but nonetheless he didn't.

I also am quite certain that you were part of the meeting where @roschaefer said the reason we want to have the `imprint` on the site is that we want as much as possible on the site of our production app, and that others who might host it on other severs in the future would benefit from that.

To your other point, I agree that we need to do a better job of coming to a consensus and then having a written copy of it somewhere.

mattwr18:
did we decide to close this @ogerly @datenbrei @Tirokk @roschaefer ??

roschaefer:
As discussed in the daily standup - we decided to close this. We do it later when we make the software configurable for hosters

PR #1901, opened 2019-10-15
1873 -WITHOUT CHECKBOX

### LINKED ISSUE #1873: 🐛 [Bug] Add Reference ToS and Data privacy Statement to Register Dialogue
## :bug: Bugreport
When opening the registration dialogue there is not link or reference the the terms of ervice and the data privacy statement.
register

1. Open tregistration
2. See missing reference

### Expected behavior
Before registering you should see a reference to the above information

### Version & Environment
 Type: [all]      
 - OS: [all]      
 - Browser: [all] 
 - Version [all]  
 - Device: [all]  

### Additional context
It is a legal need, that anybody is informed **before** registration. ;-)

ogerly:
## 🍰 Pullrequest
reference to terms of use, data protection, minimum age at added

# WITH CHECKBOX https://github.com/Human-Connection/Human-Connection/pull/1905

### Issues

- fixes #1873

FireShot Capture 088 - Human Connection - Human Connection - localhost
FireShot Capture 087 - Human Connection - Human Connection - localhost

### DIFF
--- a/webapp/locales/en.json
+++ b/webapp/locales/en.json
@@ -31,6 +31,9 @@
         "title": "Join Human Connection!",
         "form": {
           "description": "To get started, enter your email address:",
+          "terms-and-condition": "Our <a href=\"/terms-and-conditions\"><ds-text bold color=\"primary\" > terms of conditions!</ds-text></a> apply!",
+          "data-privacy": "Our <a href=\"https://human-connection.org/datenschutz/\" target=\"_blank\"><ds-text bold color=\"primary\" > Data privacy </ds-text></a> apply!",
+          "minimum-age": "You must be at least 18 years old!",
           "invitation-code": "Your invitation code is: <b>{code}</b>",
           "errors": {
             "email-exists": "There is already a user account with this email address!",
--- a/webapp/components/Registration/Signup.vue
+++ b/webapp/components/Registration/Signup.vue
@@ -30,6 +30,16 @@
         name="email"
         icon="envelope"
       />
+      <ds-list size="base">
+        <ds-list-item>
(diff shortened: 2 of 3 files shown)


webapp/locales/en.json
```diff
@@ -31,6 +31,9 @@
         "title": "Join Human Connection!",
         "form": {
           "description": "To get started, enter your email address:",
+          "terms-and-condition": "Our <a href=\"/terms-and-conditions\"><ds-text bold color=\"primary\" > terms of conditions!</ds-text></a> apply!",
```
roschaefer: I would not use exclamation marks. (!)

webapp/locales/en.json
```diff
@@ -31,6 +31,9 @@
...
+          "data-privacy": "Our <a href=\"https://human-connection.org/datenschutz/\" target=\"_blank\"><ds-text bold color=\"primary\" > Data privacy </ds-text></a> apply!",
+          "minimum-age": "You must be at least 18 years old!",
```
roschaefer:
```suggestion
          "minimum-age": "I confirm to be at least 18 years old",
```

webapp/locales/en.json
ogerly: That's agreed with martin. 
because you have to put a checkmark at the registration .. there is only a hint with a link to the infosource at this place. 

at this place no little checkers should be set. 

but no worries I wanted to put them also there ran. 

please talk to martin about such suggestions.

mattwr18:
agree with @roschaefer, I think actually to follow the law, it cannot be something passive, but they must actively choose to accept the terms and condtions

datenbrei:
ok, agreed. Then we have to check two times, but anyways ...

PR #1903, opened 2019-10-15
Highlight and expand linked comment

### LINKED ISSUE #1204: 🚀 [Feature] Expand and mark comment if one jumps to it by URL hashtag
## :rocket: Feature

- If a user gets a notification as an example, because they are mentioned in a comment, and they jump to the comment by clicking on the notification it would be wishful to expand the comment, because otherwise the mention could be in an invisible part of the comment.
- To get the focus of the user to the hashtaged comment it would be wishful to have a short (yellow) blinking at it like it was in the old Alpha.

### Additional context

To implement this functionality the notification by mention the user in a comment have to be implemented. See #1017 .

alina-beck:
## 🍰 Pullrequest
Highlights and expands a comment when it is targeted by an anchor link (e.g. when a user clicks on a comment notification)

### Issues
- fixes #1204 

### Dear Reviewers...
@Tirokk I decided to highlight the comment with a green border instead of the yellow blinking you suggested – please let me know if that's ok! Other reviewers' opinions are also appreciated, of course. 😉 

🕙 Please let me know what you think of the animation duration for the green border! It's set to `4s` which feels very high but I'm on a slow connection and wouldn't see it at all, otherwise.

### DIFF
--- a/webapp/components/Comment/Comment.spec.js
+++ b/webapp/components/Comment/Comment.spec.js
@@ -32,9 +32,6 @@ describe('Comment.vue', () => {
         truncate: a => a,
         removeHtml: a => a,
       },
-      $route: {
-        hash: '',
-      },
       $scrollTo: jest.fn(),
       $apollo: {
         mutate: jest.fn().mockResolvedValue({
--- a/webapp/components/Comment/Comment.spec.js
+++ b/webapp/components/Comment/Comment.spec.js
@@ -32,6 +32,9 @@ describe('Comment.vue', () => {
         truncate: a => a,
         removeHtml: a => a,
       },
+      $route: {
+        hash: '',
+      },
       $scrollTo: jest.fn(),
       $apollo: {
         mutate: jest.fn().mockResolvedValue({
--- a/webapp/components/Comment/Comment.vue
+++ b/webapp/components/Comment/Comment.vue
@@ -69,9 +69,14 @@ import scrollToAnchor from '~/mixins/scrollToAnchor.js'
 
 export default {
   mixins: [scrollToAnchor],
-  data: function() {
+  data() {
+    const anchor = `commentId-${this.comment.id}`
+    const isTarget = this.$route.hash === `#${anchor}`
+
     return {
-      isCollapsed: true,
+      anchor,
+      isTarget,
+      isCollapsed: !isTarget,
       openEditCommentMenu: false,
     }
   },
@@ -96,9 +101,6 @@ export default {
       user: 'auth/user',
       isModerator: 'auth/isModerator',
     }),
-    isTarget() {
-      return this.$route.hash === `#${this.anchor}`
-    },
     isLongComment() {
       return this.$filters.removeHtml(this.comment.content).length > 180
     },
@@ -108,9 +110,6 @@ export default {
       }
 
       return this.comment.content
-    },
-    anchor() {
-      return `commentId-${this.comment.id}`
     },
     displaysComment() {
       return !this.unavailable || this.isModerator
--- a/webapp/components/Comment/Comment.vue
+++ b/webapp/components/Comment/Comment.vue
@@ -43,12 +43,15 @@
         />
       </div>
       <div v-else>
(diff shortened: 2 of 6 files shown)


mattwr18:
I :heart: this refactor!! the code was much more complex than it needed to be! 
One comment I had, is that I didn't really notice the green border, to be honest... 
Like after testing it out a few times, I did notice it, but when I showed it to Paula, she also didn't notice it until I told her what to look for...

alina-beck:
Ok, fair point! So would you prefer a change in background color, as @Tirokk originally suggested?

mattwr18:
Not necessarily, I was thinking more like if there could be a thicker border, maybe?

webapp/components/Comment/Comment.vue
```diff
@@ -89,9 +69,14 @@ import scrollToAnchor from '~/mixins/scrollToAnchor.js'
 
 export default {
   mixins: [scrollToAnchor],
-  data: function() {
+  data() {
+    const anchor = `commentId-${this.comment.id}`
+    const isTarget = this.$route.hash === `#${anchor}`
```
roschaefer: I would prefer not to have `this.$route` in components. Is it possible to pass this as a prop?

webapp/components/Comment/Comment.vue
```diff
@@ -43,35 +43,15 @@
...
-          class="padding-left text-align-left"
-        />
-        <div class="show-more-or-less-div">
-          <span class="show-more-or-less">
-            <a v-if="!isCollapsed" class="padding-left" @click="isCollapsed = !isCollapsed">
-              {{ $t('comment.show.less') }}
-            </a>
-          </span>
-        </div>
+        <content-viewer :content="commentContent" class="comment-content" />
```
roschaefer: Wow, nice refactoring! :astonished:

webapp/components/Editor/ContentViewer.vue
```diff
@@ -31,6 +31,9 @@ export default {
       }),
     }
   },
+  beforeUpdate() {
+    this.editor.setContent(this.content)
```
roschaefer: Could you give me an idea what the effect of this is? :thinking:

webapp/components/Comment/Comment.spec.js
```diff
@@ -32,6 +32,9 @@ describe('Comment.vue', () => {
         truncate: a => a,
         removeHtml: a => a,
       },
+      $route: {
+        hash: '',
```
roschaefer: Would be obsolete, if `Comment.vue` wouldn't access `this.$route`.

webapp/components/Editor/ContentViewer.vue
alina-beck: Yep! So, we're using the `tiptap editor` to display comments. The problem was that when the `content` prop changed the `ContentViewer` didn't update so the comment would always stay in its initial state (i.e. it could not be collapsed or expanded).

To solve this I had to a) use the `content` prop within the `<template>` to trigger an update at all – so I set it as a `key` and b) update the editor with the new `content` (which is happening above).

This was not necessary before because we were using three different instances of the `ContentViewer`.

webapp/components/Comment/Comment.spec.js
alina-beck: true – thanks for pushing up your solution, I like it! :blush:

webapp/components/Editor/ContentViewer.vue
roschaefer: Thanks, sounds more like the VueJS way. :+1:

alina-beck:
@Tirokk thank you for taking the time to review! I didn't think @roschaefer would be so fast to merge this in. :wink:

Since this PR is closed now let's test it on `staging` and decide if the animation needs to be changed. (I think you're not seeing anything because the page takes too long to load and scroll down to the comment.) The duration and style can easily be fixed in another PR. :smile:

I've also put in a new PR that uses constants and increases the truncate limit as we discussed before the daily – see #1925

PR #1905, opened 2019-10-15
1873 -WITH CHECKBOX

### LINKED ISSUE #1873: 🐛 [Bug] Add Reference ToS and Data privacy Statement to Register Dialogue

ogerly:
## 🍰 Pullrequest
reference to terms of use, data protection, minimum age at added  WITH CHECKBOXES

# WITHOUT CHECKBOX https://github.com/Human-Connection/Human-Connection/pull/1901 

### Issues
- fixes #1873

FireShot Capture 090 - Human Connection - Human Connection - localhost
FireShot Capture 089 - Human Connection - Human Connection - localhost

### DIFF
--- a/webapp/components/Registration/Signup.vue
+++ b/webapp/components/Registration/Signup.vue
@@ -30,15 +30,24 @@
         name="email"
         icon="envelope"
       />
-      <ds-list size="base">
-        <ds-list-item>
-          <ds-text v-html="$t('components.registration.signup.form.terms-and-condition')"></ds-text>
-        </ds-list-item>
-        <ds-list-item>
-          <ds-text v-html="$t('components.registration.signup.form.data-privacy')"></ds-text>
-        </ds-list-item>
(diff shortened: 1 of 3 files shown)


mattwr18:
step in the right direction, but good catch by @roschaefer to point out that we need to have it on a page that all users hit...
I wonder if the issue could not have said this, so you wouldn't have spent time needlessly :woman_shrugging: :

datenbrei:
This is why I proposed to have it here just as an informal text, see #1901

PR #1925, opened 2019-10-16
Improved comment truncation

### LINKED ISSUE #1795: 🐛 [Bug] Shortened text is shown longer than full text
## :bug: Bugreport
1. If a text needs no shortening, it should not have an option to be shown full.
2. If 

1. Write a comment, which is three lines long
2. Save
3. View: It can be shown full although it is already fully visible.
Normally shortened version:
short

Normally full version:
long

### Expected behavior
The short version should not be longer than the full version

### Version & Environment
 Type: [Desktop]      
 - OS: [Linux]      
 - Browser: [Firefox] 
 - Version [69]  
 - Device: [Notebook]  

### Additional context
It may have to do with the shortened version showing dots, which opnes a new line.

alina-beck:
## 🍰 Pullrequest

- use constants for comment truncation, as suggested by @Tirokk in PR #1903 
- only truncate comments with more than 300 characters

At the moment also comments with 182 characters are shortened to 180 characters – so it looks weird when expanding and collapsing. When we increase the difference (`300` is my suggestion) we make sure to truncate only when it makes a noticeable difference.

as @datenbrei kindly pointed out this also fixes #1795 :wink:


datenbrei:
This may be the same issue as #1795

PR #1926, opened 2019-10-16
🍰 Add Cypher statement for ordering

### LINKED ISSUE #1924: 🐛 [Bug] Moderators reports list does not sort
## :bug: Bugreport

Moderators reports list does not sort.

Bildschirmfoto 2019-10-16 um 12 40 39

1. log in as moderator
2. in the users menu click on moderator
3. see the list

### Expected behavior

List should sort descent.

Tirokk:
#1314  🍰 Pullrequest

Add Cypher statement for ordering.

- minor db manipulation text change

### Issues

- fixes #1924

### Todo

- [X] None

### DIFF
--- a/backend/src/schema/resolvers/reports.js
+++ b/backend/src/schema/resolvers/reports.js
@@ -1,8 +1,9 @@
 export default {
   Mutation: {
-    report: async (_parent, params, { driver, user }, _resolveInfo) => {
+    report: async (_parent, params, context, _resolveInfo) => {
       let createdRelationshipWithNestedAttributes
       const { resourceId, reasonCategory, reasonDescription } = params
+      const { driver, user } = context
       const session = driver.session()
       const writeTxResultPromise = session.writeTransaction(async txc => {
         const reportRelationshipTransactionResponse = await txc.run(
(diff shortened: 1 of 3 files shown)


backend/src/schema/resolvers/reports.js
```diff
@@ -58,54 +59,68 @@ export default {
...
+      const { driver } = context
       const session = driver.session()
-      const res = await session.run(
-        `
+      let response
+      let orderByClause
+      switch (params.orderBy) {
+        // case 'createdAt_asc':
+        //   orderByClause = 'ORDER BY report.createdAt ASC'
+        //   break
```
mattwr18: I don't understand why this is commented out?
maybe because there is no UI to allow `moderators` to order how they want, but I am guessing.
If that is the case, I don't see the need to leave it commented out, since there is more than one UI... we have the playground, for example.

if there is no `orderBy` it just comes in whatever order the database happens to return it?

Tirokk: Okay. @mattwr18 
I was uncertain how to handle this.

> if there is no orderBy it just comes in whatever order the database happens to return it?

Yes! That was the cause I’ve seen that no ordering was applied.

Tirokk:
That exactly happened to me (and you) 😅😂

Issue #1830, opened 2019-10-07
🚀 [Feature] [Moderation] Enforce the moderators to first report before they can disable


Tirokk:
## :rocket: Feature

Users, posts, and comments should be only possible to disable if they are reported before.

- This is important for transparency. A moderator can not disable nodes without reason and knowledge of other moderators.
- This makes the human interface less complex. See Additional context …

### Design & Layout

- For moderators the context menu should have no item `Disable Contribution` (and other nodes) if the node is not reported.
- As a hint only for moderators the context menu report item `Report Post` (and other nodes) should be named `Report Post (necessary for disabling)`.

### Additional context

Then we need no extra list for disabled nodes.

### Todo

- [x] Backend should only disable nodes that are reported already.
- [x] Change the database: All disabled nodes get a report relation with the submitter as the disabler, the `reasonCategory` as `other` and the `reasonDescription` as `Automatically created for database consistency …`.

datenbrei:
Very good. This way we always have a protocol about why something was disabled, when it was done and who has done so. This is very important if we think about users trusting in us.

Tirokk:
This is or can be still necessary.
Please prove.

Tirokk:
This issue has been already totally fulfilled by PR List and protocol moderation #1954 and can be closed.
Or? @mattwr18

Tirokk:
I close this, because it is in my eyes totally fulfilled.

PR #1954, opened 2019-10-18
List and protocol moderation

### LINKED ISSUE #1710: 🚀 [Feature] [Moderation] Protocol and list Moderation Decisions
## :rocket: Feature

When reported content arrives in the moderator's panel the moderator has two options and should protocol these as states of the moderation request:

1. Protocol: A morderator has startet to work on it
2. Protocol: Disabled the user, contribution or comment
2. Protocol: Decided to NOT disable the content or user.

The "Protocol" is the list in the Moderation panel.

### Design & Layout

The decisions should be shown as a status in the list of reports. maybe as an icon or as text.

### Additional context

It would be good to limit the history to 4 weeks or so, older entries should later just show up in some kind of archive.

Related:
- 🚀 [Feature] [Moderation] When reporting Content Users can add specific information #1707
- 🚀 [Feature] [Moderation] See reportet Content with additional information #1708 
- 🚀 [Feature] [Moderation] Send Feedback Email to reporting User #1709

Tirokk:
## 🍰 Pullrequest

List and protocol moderation.

### Issues

- fixes #1710
- relates #1830

### Todo

- [x] design moderators reports list
- [x] fine design of moderators reports list (with tags and CSS) (we can leave it as is, for now in my opinion)
- [x] write frontend tests
- [x] write a storybook story #1999 (can be)
- [x] script to restructure the database data
  - [ ] Should it contain a counter-test that checks if all DECIDED and REPORTED have been converted?
- [x] investigate how to refactor the backend best
- [x] decide if all decisions should stay in the database: YES, all decisions should stay in the database!
- [x] implement backend mutations
  - [x] validates in backend that review can only be made if a report exists
- [x] refactor mutations again to next new database structure with a `Claim` node
- [x] write backend tests


ogerly:
i see now :)

backend/src/middleware/validation/validationMiddleware.js
```diff
@@ -59,13 +59,23 @@ const validateUpdatePost = async (resolve, root, args, context, info) => {
...
+  const { user, driver } = context
+  if (resourceId === user.id) throw new Error('You cannot review yourself!')
   const session = driver.session()
   const reportQueryRes = await session.run(
     `
-      MATCH (:User {id:$submitterId})-[:REPORTED]->(resource {id:$resourceId}) 
-      RETURN labels(resource)[0] as label
+      MATCH (resource {id: $resourceId})
+      WHERE resource:User OR resource:Post OR resource:Comment
+      OPTIONAL MATCH (:User)-[filed:FILED]->(:Report {closed: false})-[:BELONGS_TO]->(resource)
```
roschaefer: Are there ever situations in which ~we have a `:Report` without a user who filed it or~ no `author:User` who `:WROTE` the resource?

backend/src/middleware/validation/validationMiddleware.js
```diff
@@ -59,13 +59,23 @@ const validateUpdatePost = async (resolve, root, args, context, info) => {
...
+  const { resourceId } = args
+  const { user, driver } = context
+  if (resourceId === user.id) throw new Error('You cannot review yourself!')
   const session = driver.session()
   const reportQueryRes = await session.run(
     `
-      MATCH (:User {id:$submitterId})-[:REPORTED]->(resource {id:$resourceId}) 
-      RETURN labels(resource)[0] as label
+      MATCH (resource {id: $resourceId})
+      WHERE resource:User OR resource:Post OR resource:Comment
```
roschaefer: Nice

backend/src/schema/resolvers/moderation.js
```diff
@@ -1,41 +1,51 @@
...
-      if (!resource) return null
-      return resource.id
+      try {
+        const cypher = ` 
+            MATCH (moderator:User {id: $moderatorId})
+            MATCH (resource {id: $params.resourceId})<-[:BELONGS_TO]-(report:Report {closed: false})
+            WHERE resource:User OR resource:Post OR resource:Comment
+            MERGE (report)<-[review:REVIEWED]-(moderator)
+            ON CREATE SET review.createdAt = $dateTime, review.updatedAt = review.createdAt
+            ON MATCH SET review.updatedAt = $dateTime
```
roschaefer: Very nice

backend/src/schema/resolvers/moderation.js
```diff
@@ -1,41 +1,51 @@
...
+            SET review.disable = $params.disable, review.closed = $params.closed
```
roschaefer: ~Ah, so closing and reviewing can be the same action?~
After the reading below I ask myself, why do we need to store the `review.closed`? It's implicit because the most recent review must be the closing review.

webapp/components/_new/features/ReportList/ReportList.vue
```diff
@@ -0,0 +1,150 @@
...
+      })
+    },
+  },
+  apollo: {
+    reportsList: {
+      query: reportsListQuery(),
+      update({ reports }) {
+        this.reports = reports
+        this.allReports = reports
+        this.unreviewedReports = reports.filter(report => !report.reviewed)
```
roschaefer: I changed my mind. This is certainly something that we need to `filter` on the backend. Otherwise I see this:

Screenshot - 2019-11-30T014846 489

roschaefer:
By the way: This is a perfect example that the lines between a generic and a feature component are blurry. Why is `ReportList` a feature component but `ReportsTable` is a generic components? :thinking: @alina-beck

backend/src/middleware/validation/validationMiddleware.js
mattwr18: I don't think so... I think this `OPTIONAL MATCH` is because the resource is not always a `Post` or `Comment` that was written by someone. If it is then we are checking that a moderator cannot review there own `Post|Comment`, see lines 100-105

webapp/components/_new/features/ReportList/ReportList.vue
mattwr18: not sure how you got into this situation, as I never got here...
did you seed it with the new seed data?
can be that you had unconverted relationships?
I thought about filtering it in the `backend`, but it's challenging because you need to check if a `report` has any `reviewed`... I can check out the filters betters, but one benefit of this way is that we grab one list, filter it and save the values to variables, and we minimize the number of backend calls... 
where should it error out, if the data looks correctly in the database?

mattwr18:
@roschaefer, this `PR` is still a `[WIP]`, and I intend on pairing with @alina-beck on the frontend, as well as adding component tests and stories, still I like that you are giving feedback... and even got in on the fun.

No one said it was simple or always straightforward when a component should be placed in `generic`/'features', but that is partly the point. The idea is that we as a team, and before that the `PR` creator, think about these decisions more critically. 

For example, I placed the `ReportList` in `/features` because it involves communicating with the `backend` to retrieve data related to `reports`... `ReportsTable` just presents that data, though it does manage internal state. For the `NotificationsTable`, I did the communication with the `backend` in the `page` component itself, and considered doing that here as well since now the `page` component just renders the `ReportList` component, so maybe I overdid it- something I would love to get feedback on from @alina-beck, as well as others.

Something else challenging for me was that I would love to be able to have a `Table` component that is indifferent about the data it receives- just displays it.. then it would truly be a "dumb" component. 

This would be awesome because then we could use it for both the `NotificationsTable` and the 
`ReportsTable`, however, this assumes that we receive the same data structure, which we don't.
`Notifications` has `from` and `to`, whereas `Report` has `filed`/`resource`... I thought about refactoring that to make them equal, but the `Notification` comes `from` a `Post|Comment` and `to` a `User`, whereas a `Report` is `filed` by a `User` and `belongsTo` a `resource(Post|Comment|User)`... this is just one example, there would surely be other areas of difficulty.

backend/src/schema/resolvers/moderation.js
mattwr18: Actually, maybe we don't need... I think that @Tirokk was explicitly setting the `disable`/`closed` on the `REVIEWED` relationship when I starting refactoring. Maybe we can remove `closed` though since if the `Report` is `closed`, as you say it can only be the last `review` that closed it.

webapp/locales/en.json
```diff
@@ -482,7 +482,7 @@
     "reports": {
       "empty": "Congratulations, nothing to moderate.",
       "name": "Reports",
-      "typeRowHeadline": "Type",
+      "typeRowHeadline": "Current status",
```
mattwr18: I'd have to check, but I guess this is not the `Current status`... by `Type`, I think @Tirokk was referring to the `type` that used to be returned from the `backend` as in the `type of Resource that was reported`... in the `NotificationsTable`, I didn't have a header for this column...
I'll see what @alina-beck thinks about the `layout` though, since I trust her judgement more than any of ours :stuck_out_tongue_closed_eyes:

roschaefer:
I'm also not really happy about the composition:
`ReportList` is rendering a single `ReportTable`, apparently `List` is bad naming here. Then the `ReportTable` renders multiple `<thead>` and `<tbody>` for each report :astonished: . When I click on the button to show the review history, all reports get opened.

mattwr18:
No, all reports don't get shown, more details about the report gets shown... There can only be one open report on the same resource at a time, but there can be many people who have filed a report against a resource for multiple different reasons.
Yes, I agree we could probably come up with a better naming convention for `ReportList`... It is a list of reports, but that is in the form of a table, then @Tirokk programmed it to have a table nested inside the `ReportTable` for all the info for `FILED`...
I think we could start by just having the page component fetch the data from the `backend` and then have the `ReportsTable`,or change it to a`List`, maybe... Let's get @alina-beck feedback...

mattwr18:
wow @alina-beck !!! I'm super impressed by the code you contributed, and even more impressed by the look and feel of the `moderation` page!!! :green_heart: 
Thank you sooo much!!! :blush: 
check it out @HC-Team 
Screenshot from 2019-12-03 19-10-47

I'll just work on the stories/tests now

webapp/components/features/ReportList/ReportList.vue
```diff
@@ -6,13 +6,13 @@
...
     </div>
-    <reports-table :reports="reports" @confirm="confirm" />
+    <reports-table :reports="reports" @confirm="openModal" />
   </ds-card>
 </template>
 <script>
 import { mapMutations } from 'vuex'
 import DropdownFilter from '~/components/DropdownFilter/DropdownFilter'
-import ReportsTable from '~/components/_new/features/ReportsTable/ReportsTable'
+import ReportsTable from '~/components/features/ReportsTable/ReportsTable'
```
roschaefer: Oh, yes, I like removing `_new`.

PR #1968, opened 2019-10-21
fix: Only one ellipse is displayed

### LINKED ISSUE #1893: 🐛 [Bug] Double ellipses for truncated strings
## :bug: Bugreport

We refactored to use an npm library `trunc-html`, and now we have double `ellipses`

1. write a comment more than 180 characters
2. you should see show more link to click on and two sets of ellipses
3. you should see the same for long users names/long post titles
4. ...
5. Profit

### Expected behavior

we should only have one set of ellipses

At the moment, we are using `vue-filters` for our truncating...
```
 if (output.length < value.length) {
   output += ' …'
 }
```
this bit of code is superfluous
see https://github.com/Human-Connection/Human-Connection/blob/master/webapp/plugins/vue-filters.js#L42

Mogge:
### Issues

- fix #1893  

### Todo

- [X] None

### DIFF
--- a/webapp/plugins/vue-filters.js
+++ b/webapp/plugins/vue-filters.js
@@ -31,11 +31,7 @@ export default ({ app = {} }) => {
       if (length <= 0) {
         return value
       }
-      let output = trunc(value, length).html
-      if (output.length < value.length) {
-        output += ' …'
-      }
-      return output
+      return trunc(value, length).html
     },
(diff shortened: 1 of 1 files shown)


roschaefer:
@Mogge if you write "fix #4711" in the PR description, that issue will be closed auto-magically when this PR gets merged.

PR #1975, opened 2019-10-21
Add notifications page with All Notifications

### LINKED ISSUE #1974: 🚀 [Feature] Notifications page with all notifications
## :rocket: Feature

```
As a user
I'd like to have a page listing all my notifications
So that I can find old notifications and the posts related to them
```

mattwr18:
### Issues

- fixes #1974 

### Todo

- [x] Link always visible in NotificationMenu, but styling is not pretty
- [x] other links? where? Added to the AvatarMenu as suggested by @alina-beck 
- [x] sorting options label is not translatable, we can follow the pattern @roschaefer introduced for the sortingOptions for `Post`, but it involves extracting everything to `Vuex`, so is it a priority? Not a priority?
- [x] Add tests


mattwr18:
Not really happy with the design of this, I started off trying to list the notifications where a max of them would be showed per page, but don't know how to make the whole row clickable in a table, ended up with this so far... 
Screenshot from 2019-10-21 19-47-16

alina-beck:
Regarding the latest design – it's way better than before! 😍 
image

1. Is the `Content` truncated? Because for longer posts or comments this could get messy otherwise. If that's too complicated I'd also be fine with not displaying the content at all, although it definitely is nice to have
2. `Type` is actually `Post title`, isn't it? Or are there cases where it is something else?
3. It would be awesome to somehow distinguish `read` from `unread` notifications. Otherwise the sorting doesn't do much good. (Maybe a checkmark or something, similar to Github notifications? Not necessarily interactive, though. That's probably a whole new feature. :wink:)
4. Can I also sort by `date` and `user`? That would be _amazing_! 😎 (Sorry, I'm just going by the design, haven't checked out the branch...)

mattwr18:
hey, thanks @alina-beck, regarding your questions
1) it is showing the `contentExcerpt`, which is truncated at 180 characters
2) I changed it to `Post`, but didn't update the `en.json` cause it's always the Post's title
3) I think I can add the same styling as the `NotifcationsMenu` to add opacity and it will be easier to identify when it's showing all notifications which are read and unread.... I can also have another column for the status... I didn't understand what you meant by the sorting doesn't do much good(?) 
Currently, a user can select to see just the Unread/Read posts and it will filter the table to show just those they want to see... this to me is a nice feature and works... maybe it's cause of the wording... it doesn't sort actually, it filters
4) similar to above, at the moment there is no sorting, just filtering, but that could be changed to sorting, if that seems more intuitive

alina-beck:
got it! I thought you said sorting (which would be a nice-to-have feature for the future) but if it's a filter then I think it is clear enough and there's no need to add an extra column. :smile:

mattwr18:
@alina-beck one thing I am certainly not happy about is that the link in the NotificationMenu is at the bottom of the menu, and I would like it to always be shown... do you know any CSS tricks that I can use? I have searched the internet a little bit and can keep looking, but maybe you know some magic.

alina-beck:
You could try giving the link `position: fixed` (and a background color). I think even more important is to have the link available somewhere else, too – otherwise I can only open the page when I have _new_ notifications. But where? Hmm...

Maybe in the `avatar menu`, alongside `my profile` and `settings`? Then we could even completely hide the notifications button from the header when there are no new notifications, oder?

And the label translation issue – I'd say it's not a priority.

mattwr18:
For `If I’m in the sorting Alle (All) I would like to see if I read a notification already.` I would like that as well, do you know how we can add this opaque styling to a single row in our `ds-table`?

2) I'll have to have a look at what you mean
3) If you click on the `Post` title it does take you to the comment where you were mentioned
4) It is not very noticeable in the `NotificationsMenu` either, probably less so, but I do like your idea of adding an icon as a column

Tirokk:
Response to ☝🏼

1. I think you can add a class like in the notification item for opacity to all elements of a row.
2. Okay.
3. If I click on a posts title I would await to get to the post and not to the comment where I am mentioned, if the comment content with the mention is afterwards in the same row … !?
4. I would add the icons in front of what belongs to a post (maybe only once) and in front of what belongs to a comment …

mattwr18:
1) it's not so simple, @Tirokk 
2) ...
3) I don't know what you mean, this is the same functionality as the NotificationsMenu
4) yep

mattwr18:
actually, I can add a class to every column and then add the opaque styling, but I'm sure @alina-beck knows a better way... 
Screenshot from 2019-10-25 11-34-26

Tirokk:
3. We can have a more detailed functionality here if we want. If you like to have the same functionallity as in the Notifications Menu then it would be more like: Click on a whole row and that links to the notification reason …

Your picture looks got to me …

mattwr18:
you can't just easily add a link around the whole row @Tirokk

mattwr18:
I'm not worried how it looks, I am concerned that I must add a class to each and every column... I would prefer to add it to one column and then target the `tr` element with scss, but I don't know if that is even possible

mattwr18:
@alina-beck I tried to follow your advice, but maybe it's hard to explain the issue... or I'm doing something wrong... for now I've put it at the top, which I don't like, but don't know how to change in a way that means the link is always visible
Screenshot from 2019-10-25 12-38-12

mattwr18:
big thank to @alina-beck for helping with the styling of the link 
Screenshot from 2019-10-28 17-37-22


really nice to see @aonomike and @mindcodemediator for joining and helping with the component testing of the `AvatarMenu`

mattwr18:
we now have a paginated notifications page! Also, I am trying to follow @alina-beck suggestions in https://github.com/Human-Connection/Human-Connection/pull/2075/files

Screenshot from 2019-10-29 18-13-24
Screenshot from 2019-10-29 18-13-44

I did this by extracting `NotificationsDropdownFilter`, `NotificationsTable`, and `Paginate` into their own components. 
I'm sure I didn't follow your suggestions perfectly @alina-beck, but maybe you can point out some ways to improve it in a code review?

backend/src/schema/resolvers/notifications.js
```diff
@@ -41,13 +40,15 @@ export default {
         default:
           orderByClause = ''
       }
-
+      const offset = args.offset ? `SKIP ${args.offset}` : ''
+      const limit = args.first ? `LIMIT ${args.first}` : ''
```
roschaefer: Our GraphQL schema saves us here, because you must enter an Integer. But if that wouldn't be the case, we would have a cypher injection vulnerability here.

roschaefer:
@mattwr18 I love the design, too! It would be great, if the notification page would also update the notification counter. Since the notification page and the notification counter are not on a component tree, I would suggest to use VueX here.

I'm too tired to contribute that right now, but I think that's what I'm going to do tomorrow.

webapp/components/AvatarMenu/AvatarMenu.vue
```diff
@@ -0,0 +1,161 @@
...
+      color: $text-color-link-active;
+    }
+  }
+  nav {
+    margin-left: -$space-small;
+    margin-right: -$space-small;
+    margin-top: -$space-xx-small;
+    margin-bottom: -$space-xx-small;
+    a {
+      padding-left: 12px;
```
alina-beck: Please no negative margins and no styling of HTML tags! In this case `nav` has the class `.ds-menu`, so if you need to override styles (😱 😭) do it for the class instead. Same for `a` – the ones targeted here are `.ds-menu-item-links`.

PS: I know this is not your code @mattwr18, just copy-🍝. But you have the chance to make it better! :smile:

webapp/components/NotificationsTable/NotificationsTable.vue
```diff
@@ -0,0 +1,105 @@
...
+    markNotificationAsRead(notificationSourceId) {
+      this.$emit('markNotificationAsRead', notificationSourceId)
+    },
+  },
+}
+</script>
+<style lang="scss">
+.notification-status {
+  opacity: 0.6; /* Real browsers */
+  filter: alpha(opacity = 60); /* MSIE */
```
alina-beck: Oh yeah? Cool trick! Haven't seen this before. 😎 And the filter is not applied in other browsers, on top of the `opacity` setting?

Btw, according to [caniuse](https://caniuse.com/#search=opacity) IE understands opacity from version 9.

webapp/components/Paginate/Paginate.spec.js
```diff
@@ -0,0 +1,71 @@
...
+        wrapper = Wrapper()
+        nextButton = wrapper.findAll('.ds-button').at(0)
+        expect(nextButton.attributes().disabled).toEqual('disabled')
+      })
+
+      it('is not disabled if hasNext is true', () => {
+        expect(nextButton.attributes().disabled).toBeUndefined()
+      })
+
+      it('emits back when clicked', async () => {
```
alina-beck:
```suggestion
      it('emits next when clicked', async () => {
```

...I guess?

webapp/components/Paginate/Paginate.vue
```diff
@@ -0,0 +1,26 @@
...
+  methods: {
+    back() {
+      this.$emit('back')
+    },
+    next() {
+      this.$emit('next')
+    },
+  },
+}
+</script>
```
alina-beck: Absolutely ❤️ this component! I wish they could all look like this – neat, small, concise, reusable! :smile:

webapp/components/Paginate/Paginate.vue
```diff
@@ -0,0 +1,26 @@
+<template>
+  <ds-flex direction="row-reverse">
+    <ds-flex-item width="50px">
+      <ds-button @click="next" :disabled="!hasNext" icon="arrow-right" primary />
+    </ds-flex-item>
+    <ds-flex-item width="50px">
+      <ds-button @click="back" :disabled="!hasPrevious" icon="arrow-left" primary />
```
alina-beck: styling-wise I think these could just be regular buttons – without the `primary` attribute

webapp/layouts/default.vue
```diff
@@ -67,52 +67,7 @@
...
-                          </ds-menu>
-                          <hr />
-                          <nuxt-link class="logout-link" :to="{ name: 'logout' }">
-                            <ds-icon name="sign-out" />
-                            {{ $t('login.logout') }}
-                          </nuxt-link>
-                        </div>
-                      </template>
-                    </dropdown>
+                    <avatar-menu placement="top" />
```
alina-beck: woohoo! 🎉

webapp/locales/de.json
```diff
@@ -171,7 +171,18 @@
...
-    "comment": "Kommentar"
+    "comment": "Kommentar",
+    "title": "Benachrichtigungen",
+    "pageLink": "Alle Benachrichtigungen",
+    "post": "Beitrag",
+    "user": "Benutzer",
+    "content": "Inhalt",
+    "sortingLabel": {
+      "all": "Alle",
+      "read": "Lesen  ",
```
alina-beck:
```suggestion
      "read": "Gelesen",
```

webapp/pages/notifications/index.vue
```diff
@@ -0,0 +1,106 @@
+<template>
+  <ds-card space="small">
+    <ds-flex>
+      <ds-flex-item :width="{ lg: '85%' }">
+        <ds-heading tag="h3">{{ $t('notifications.title') }}</ds-heading>
+      </ds-flex-item>
+      <ds-flex-item class="sorting-dropdown">
```
alina-beck: Here the styleguide messes things up again... If you don't give the `ds-flex-item` a `width` it will have a `flex-grow: 1` inline style so it will take up the whole remaining width of the flexbox, making it impossible to align it right.

As a workaround I would suggest:
1. give the `ds-flex-item` a `width="100px"`
2. remove the `sorting-dropdown` class
3. give the `ds-flex` a class with `justify-content: space-between;`

mattwr18:
> great teamwork
> 
> Great teamwork everyone! I think this can be merged now, oder?

I would say no @roschaefer ... in my mind, we need to at a minimum add component tests for the 3 or so components that have not been tested yet... also, I woke up thinking that the `NotificationsDropdownFilter` could be refactored to be more generic in line with @alina-beck 's suggestion that components should be stateless, and therefore easier to reuse.

This `PR` is taking longer than originally intended, but I think I can attempt to get it finished this weekend.

webapp/components/NotificationsDropdownFilter/NotificationsDropdownFilter.vue
```diff
@@ -0,0 +1,69 @@
...
+      selected: this.$t('notifications.sortingLabel.all'),
+      sortingOptions: [
+        { label: this.$t('notifications.sortingLabel.all'), value: null },
+        { label: this.$t('notifications.sortingLabel.read'), value: true },
+        { label: this.$t('notifications.sortingLabel.unread'), value: false },
+      ],
+    }
+  },
+  computed: {
+    routes() {
```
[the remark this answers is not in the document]
mattwr18: haha, good point! we refactored this together, and it's much better now... classic started off one way, then got refactored and it didn't make sense to map it any longer :stuck_out_tongue:

webapp/components/NotificationsDropdownFilter/NotificationsDropdownFilter.vue
```diff
@@ -0,0 +1,69 @@
...
+        return {
+          label: option.label,
+          value: option.value,
+        }
+      })
+      return routes
+    },
+  },
+  methods: {
+    sortNotifications(option, toggleMenu) {
```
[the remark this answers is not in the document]
mattwr18: yes :smiley:

webapp/components/AvatarMenu/AvatarMenu.vue
mattwr18: I have removed this, but unfortunately the style is still being applied because of this https://github.com/Human-Connection/Human-Connection/blob/master/webapp/assets/styles/imports/_tooltip.scss#L100

Could we maybe refactor that in another `PR` @alina-beck ?

webapp/components/NotificationsTable/NotificationsTable.vue
mattwr18: refactored, and using design token to boot!

webapp/components/Paginate/Paginate.spec.js
mattwr18: yup copy :spaghetti:

webapp/pages/notifications/index.vue
mattwr18: great trick!

webapp/components/NotificationList/NotificationList.spec.js
```diff
@@ -3,8 +3,8 @@ import NotificationList from './NotificationList'
 import Notification from '../Notification/Notification'
 import Vuex from 'vuex'
 import Filters from '~/plugins/vue-filters'
-
 import Styleguide from '@human-connection/styleguide'
+import { notifications } from '~/components/utils/Notifications'
```
roschaefer: Hm, I see what you're doing here. Personally I would prefer to have a folder `NotificationList.spec/` where I would place the fake data. Now you're using this in many locations, also from other components.

I don't mind having it that way. Probably I would have put that in one components folder and simply imported it in another components spec.

backend/src/schema/types/type/NOTIFIED.gql
```diff
@@ -1,4 +1,5 @@
 type NOTIFIED {
+  id: ID!
   from: NotificationSource
   to: User
   createdAt: String
```
Tirokk: Exclamation mark?

```
  createdAt: String!
  updatedAt: String!
```

webapp/components/AvatarMenu/AvatarMenu.spec.js
```diff
@@ -0,0 +1,168 @@
...
+        })
+
+        it('displays a total of 5 links', () => {
+          const allLinks = wrapper.findAll('.ds-menu-item')
+          expect(allLinks).toHaveLength(5)
+        })
+      })
+    })
+  })
+})
```
Tirokk: Wonderful 😍

webapp/components/AvatarMenu/AvatarMenu.story.js
```diff
@@ -0,0 +1,17 @@
...
+
+storiesOf('AvatarMenu', module)
+  .addDecorator(withA11y)
+  .addDecorator(helpers.layout)
+  .addDecorator(StoryRouter())
+  .add('dropdown', () => ({
+    components: { AvatarMenu },
+    store: helpers.store,
+    template: '<avatar-menu placement="top" />',
+  }))
```
Tirokk: Wonderful 😍

webapp/components/NotificationMenu/NotificationMenu.vue
```diff
@@ -41,54 +46,39 @@ export default {
...
-          return !this.displayedNotifications.find(oldN => this.equalNotification(newN, oldN))
-        })
-        this.displayedNotifications = newNotifications
-          .concat(this.displayedNotifications)
-          .sort((a, b) => {
-            return new Date(b.createdAt) - new Date(a.createdAt)
-          })
-        return data.notifications
+      update({ notifications }) {
+        return unionBy(notifications, this.notifications, notification => notification.id).sort(
```
Tirokk: Wow. This is even shorter !!! 😍

webapp/components/Paginate/Paginate.vue
Tirokk: Cool stuff. I like this as well 😍

webapp/layouts/default.vue
Tirokk: 💋

webapp/components/NotificationMenu/NotificationMenu.vue
mattwr18: thanks to @roschaefer ;)

Tirokk: 👍🏼

PR #1999, opened 2019-10-22
[WIP] [Systems-Development-and-Frameworks] Storybook for reported content 

roschaefer:
## 🍰 Pullrequest

This is the new exercise for the course "Systems Development and Frameworks". As a user you can report content that violates the community guidelines. It will show up in such a list:

Screenshot - 2019-10-22T214308 150

What we didn't thought of is that multiple users could report the same thing. So a moderator should see all reports grouped by resource, e.g. one post and a list of all users who reported the post and when.

Relates to #1710 .

#### See moderation page:

1. Visit https://nitro-staging.human-connection.org/
2. Log in as `moderator@example.org` and password `1234`.
3. Click on the avatar menu and choose `Moderation`.

#### Where is the code that renders the list?

It is here: https://github.com/Human-Connection/Human-Connection/blob/master/webapp/pages/moderation/index.vue

#### How can I report someting?

1. Visit https://nitro-staging.human-connection.org/
2. Log in as any user, you can choose `huey@example.org` or `dewey@example.org` etc. and password `1234`
3. Click on the context menu of a post or comment or the context menu on the user profile page. You must not be this user or must not be the author of this post or comment.

###  Task 1: Refactor

All the code which is relevant to show a single line in the table should be moved into this component:

https://github.com/Human-Connection/Human-Connection/blob/storybook_for_reported_content/webapp/components/ReportedContent/ReportedContent.vue

Import the component on the moderation page component :point_up: 

Refactoring is started in #1954 now.

### Task 2: Write a storybook story

You can find the storybook story here: https://github.com/Human-Connection/Human-Connection/blob/storybook_for_reported_content/webapp/components/ReportedContent/ReportedContent.story.js

It would be great if you could write data to show the `ReportedContent` component if the content is a `Post` a `Comment` or a `User` respectively.

#### Task 2.1: Implement the moderation "log"

As described in the screenshot above :point_up: we didn't really think of a resource being reported multiple times. The desired behaviour now is to group all reports by resource and also show if a moderator has disabled or enabled a resource. This list is called "log" (don't know if that's a good name :thinking:). But I sketched the data in here: https://github.com/Human-Connection/Human-Connection/blob/storybook_for_reported_content/webapp/components/ReportedContent/ReportedContent.story/data.js

So the goal is to list this log just below the row with the reported resource.

### Task 3: Write a component test

You can find a blueprint for a component tere here:
https://github.com/Human-Connection/Human-Connection/blob/storybook_for_reported_content/webapp/components/ReportedContent/ReportedContent.spec.js

In the test you could e.g. test:

* links to the reported content
* given the content is not disabled
  * it shows a dash in the column `Disabled by`
* given the content is disabled
  * it shows who disabled the content

### DIFF
--- a/webapp/components/ReportedContent/ReportedContent.spec.js
+++ b/webapp/components/ReportedContent/ReportedContent.spec.js
@@ -1,0 +1,52 @@
+// import { mount } from '@vue/test-utils'
+// import ReportedContent from './ReportedContent.vue'
+
+let resource // eslint-disable-line no-unused-vars
+
+/**
+ * This file is just a sketch. Feel free to change it as needed.
+ */
+
+describe('ReportedContent', () => {
+  describe('given a post', () => {
+    beforeEach(() => {
+      resource = {
+        /* ? */
+      }
+    })
+
+    it.todo('links to the post')
+  })
+
+  describe('given a comment', () => {
+    beforeEach(() => {
+      resource = {
+        /* ? */
+      }
+    })
+
+    it.todo('links to the post of the comment')
+  })
+
+  describe('given the resource is not disabled', () => {
+    beforeEach(() => {
+      resource = {
+        disabled: false,
+        // ...
+      }
+    })
+
+    it.todo('shows a dash in the column "Disabled by"')
+  })
+
+  describe('given the resource is disabled', () => {
+    beforeEach(() => {
+      resource = {
+        disabled: true,
+        // ...
+      }
+    })
+
+    it.todo('shows who disabled the content')
+  })
+})
--- a/webapp/components/ReportedContent/ReportedContent.story.js
+++ b/webapp/components/ReportedContent/ReportedContent.story.js
@@ -1,0 +1,59 @@
+import { storiesOf } from '@storybook/vue'
+import { withA11y } from '@storybook/addon-a11y'
+import helpers from '~/storybook/helpers'
+import {
+  onePostReportedTwoTimes,
+  reportedPostDisabledByModerator,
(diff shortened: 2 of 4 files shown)


vbelolapotkov:
Hey @roschaefer!  I'd like to work on this one if it's not urgent.

vbelolapotkov:
@roschaefer I've checked the moderation page component and found it's impossible to extract table row into own component without replacing `ds-table` with something different. The reason is inside `ds-table` component ([src code is here](https://github.com/visualjerk/vue-cion-design-system/blob/master/src/system/components/data-display/Table/Table.vue)) it provides the slots for row columns but not the whole row.

So I see the following options here:
1. Submit a PR into the design system repo to add row slot. Wait until it's merged and use it.  Pros: we stay consistent with design system. Cons: it will take time, probably long.

2. Replace `ds-table` with similar `hc-table` where row slot is available. Found example [here](https://medium.com/binarcode/understanding-scoped-slots-in-vue-js-db5315a42391). Pros: fast and simple solution. Cons: broken consistency with design system.

3. Use 2 as a temporary solution, submit the PR to design system, switch back to `ds-table` once our PR is merged and released.

Please let me know what do you think?

P.s. IMO option 3 makes sense :)

roschaefer:
@Tirokk you're working on #1710 do you want to merge this storybook?

roschaefer:
We have three storybooks  now for `ReportsTable.vue`, `ReportsList.vue` and `FiledReportsTable`, so I believe we can close this.

PR #2014, opened 2019-10-23
[WIP] Pinned post is first

### LINKED ISSUE #2013: 🐛 [Bug] Pinned posts don't stay in the first position
## :bug: Bugreport
sometimes pinned posts don't end up in first position and move even further down when adding new posts:
image

1. pin a post
2. go to index
3. add a new post
4. go back to index
5. Profit

### Expected behavior
a pinned post should always be in first position on the index page

alina-beck:
## 🍰 Pullrequest
Add a class to pin the pinned post to the top of the grid.

### Issues
fixes #2013

### DIFF
--- a/webapp/pages/index.vue
+++ b/webapp/pages/index.vue
@@ -209,7 +209,7 @@ export default {
         return {
           filter: this.finalFilters,
           first: this.pageSize,
-          orderBy: ['pinned_asc', this.orderBy],
+          orderBy: this.orderBy,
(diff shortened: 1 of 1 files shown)


alina-beck:
Added `[WIP]` so nobody will merge this accidentally – the problem with this fix is that it only puts the `pinned post` first when it is in the first few fetched results.

alina-beck:
this issue was already solved by #1998

PR #2015, opened 2019-10-23
Update to version 0.1.7

mattwr18:


mattwr18:
it's ok... it's still a work in progress, but I think I'm learning some things... so I expect it to get better in the future @roschaefer

PR #2016, opened 2019-10-23
Hide shout count on user profile unless my profile

### LINKED ISSUE #1995: 🚀 [Feature] Shouts should be hidden from other users
## :rocket: Feature
brought up by Dennis

In the future users should be able to decide if they want their comments and shouts to be public or private (see #1747). Until then all shouts should be hidden from other users and only be visible on one's own profile page.

mattwr18:
- we were asked to hide the shouts from other users by Dennis Hack, he
believes people will be more likely to shout a post if it is anonymous.
This only works from our UI to our backend... further steps would need
to be taken to truly make it anonymous. Also, we were asked to implement
it in such a way as to give the user the choice to show their shouts to
others.

Co-authored-by: Alina Beck <alina.beck@mail.com>

### Issues

- fixes #1995

### DIFF
--- a/webapp/pages/profile/_id/_slug.vue
+++ b/webapp/pages/profile/_id/_slug.vue
@@ -172,7 +172,7 @@
           <ds-grid-item class="profile-top-navigation" :row-span="3" column-span="fullWidth">
             <ds-card class="ds-tab-nav">
               <ul class="Tabs">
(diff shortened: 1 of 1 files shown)


mattwr18:
thanks to @alina-beck, we didn't break the styling with this feature

mattwr18:
Screenshot from 2019-10-23 14-01-15
Screenshot from 2019-10-23 14-01-29

PR #2023, opened 2019-10-24
Remove duplicate pinned post record

### LINKED ISSUE #2022: 🐛 [Bug] Pinned Post appears twice on page
## :bug: Bugreport

Once where it is supposed to, and once where it would be ordered by the createdAt date

1. go to the site 
2. pin a post, if you are an admin, if not there must be a pinned post to replicate the bug
3. make note of the time and title of the pinned post
4. scroll down till you see it a second time
5. Profit

### Expected behavior

Screenshot from 2019-10-23 23-55-58

mattwr18:
- Maintain consistent orderBy for fetchMore

### Issues

- fixes #2022

### DIFF
--- a/webapp/pages/index.vue
+++ b/webapp/pages/index.vue
@@ -146,7 +146,7 @@ export default {
           offset: this.offset,
           filter: this.finalFilters,
           first: this.pageSize,
-          orderBy: this.orderBy,
+          orderBy: ['pinned_asc', this.orderBy],
         },
         updateQuery: (previousResult, { fetchMoreResult }) => {
           if (!fetchMoreResult || fetchMoreResult.Post.length < this.pageSize) {


roschaefer:
Do we loose the pinned post feature through this? Don't we stop ordering the pinned post at the first position with this?

mattwr18:
no @roschaefer, we maintain the pinned post, and remove duplicate records when fetchMore is called

roschaefer:
OK, `fetchMore`. Still I don't understand why the pinned post appears again. It shouldn't matter if you order by the `pinned` attribute on pages > 1.

So here is my problem:

1. On the first page, you order by `pinned` and `createdAt` and get posts at index 0..11 with 0 being the pinned post
2. On the next page you order by `createdAt` and get the posts 12..23. Post at index 0 was the pinned post in step 1 but instead the post formerly at index 1 gets shifted to index 0.
3. What about the post at index 11? Do we miss it?

mattwr18:
no @roschaefer, we don't lose any `Post`s

Issue #1413, opened 2019-08-28
🐛 [Bug] Background is scrolling when overlay is open


alina-beck:
## :bug: Bugreport
When the filter menu or notifications dropdown is open I can still scroll in the background – which is a huge issue when the screen is not big enough to display the full overlay because then I have to scroll to the bottom of the page _before_ I can scroll within the overlay.

![ezgif com-optimize(1)](https://user-images.githubusercontent.com/11694483/63853449-a3154880-c992-11e9-883d-6bed94724093.gif)

### Steps to reproduce the behavior
1. open the filter menu or notifications dropdown
2. resize your screen so the overlay doesn't fit
3. start scrolling and see how long it takes you to reach the bottom of the overlay

### Expected behavior
Background scrolling should be disabled as soon as overlays are opened.

PR #2048, opened 2019-10-27
overflow: hidden on body removed when popover

### LINKED ISSUE #1857: 🐛 [Bug] Hover on Avatar disables Scrollbar
## :bug: Bugreport
When you move your mouse over an avatar/user picture a tooltip appears and the scrollbar on the right disappearss. Because of this the whole window content moves a bit to the right.

Normal view with normal scrollbar
ok

Bad view with missing scrollbar
not-ok

1. Put your mouse over an article's author picture

### Version & Environment
 Type: []      
 - OS: []      
 - Browser: [] 
 - Version []  
 - Device: []

leternoz:
## 🍰 Pullrequest

The CSS property `overflow: hidden` was set on the `body` when the pointer hover the avatar. I don't know if this was done on purpose, but by removing it, now it doesn't hide the scrollbar.

Screenshot_20191028_000643

### Issues
- fixes #1857

### Todo

- [X] None

### DIFF
--- a/webapp/assets/styles/main.scss
+++ b/webapp/assets/styles/main.scss
@@ -52,7 +52,6 @@ $easeOut: cubic-bezier(0.19, 1, 0.22, 1);
 
 body.dropdown-open {
   height: 100vh;
-  overflow: hidden;
 }
 
 blockquote {


alina-beck:
Ah, I think we added `overflow: hidden` to the body to make sure there was no weird scroll behavior when the `notifications menu` and `categories filter` was open! Check this bug: #1413 

Do you think there is any other way to fix this? Or maybe we need a setting to only set `overflow: hidden` for the notifications menu and categories filter? Or use different components altogether? Hmm...

leternoz:
I tested on Chromium 78 and Firefox 70, it doesn't seem to have any weird behavior with the scrolling.  

Otherwise, I think, if we would like to have different behavior on `hover` for the `notification menu` and `categories filter`, and the avatar's `post card`, we would have to add an input to the `dropdown` component, to indicate if we want the `overflow: hidden` or not when it is opened.
Human Connection - Human Connection

mattwr18:
@leternoz @roschaefer ... will we close this then? or leave it open?

leternoz:
Sorry for the late answer. This PR is not fixing the problem, and I don't think I will have the time to work on it. The required solution is a bit more complex, it would require further analysis to evaluate the depth of the changes to bring.
So in the meanwhile, I suggest to close this PR.

roschaefer:
cry

That's sooo sad! But I will close as @leternoz suggested.

@leternoz please tell us if we can help you out. I would love not to loose your effort unmerged. Have you found your way into https://human-connection.org/discord already?

Issue #782, opened 2019-06-07
🚀 [Feature] Country of Contribution?


Tirokk:
## :rocket: Feature

Should the User have the possibility to select the related Country of the Contribution?
See below …

### Design & Layout

<img width="812" alt="Alpha - Create Post - Country - Bildschirmfoto 2019-06-06 um 09 13 14" src="https://user-images.githubusercontent.com/25344101/59100811-74ee1280-8927-11e9-822c-60c918ad9095.png">

### Additional context

- Do we need to have this information to show the Users Contributions of his/her Country?
- The Language is not enough to select the appropriate Contributions for a User. English is the Language of different countries, as an example.
- Otherwise, can we handle this without knowing the related Country?

God first issue, a bit more sophisticated!

ogerly:
the language used in the frontend should be the default language. 
you could omit it if the contribution language is always the language of the frontend. 
there are a lot of points. if the point stays in then you should never use it as a user. 
you could simply change the language in the frontend and all messages in this language will be filtered.

ogerly:
that is too much for a user according to my opinion. the tags do this more efficiently. for reasons it would be a 6th day.

roschaefer:
I think I agree with @ogerly : [KISS](https://en.wikipedia.org/wiki/KISS_principle) please :kissing_heart: 

Let's wait until we have users from many countries and we get the first 3 feature requests to have some sort of filtering by country.

sushidave:
@Tirokk I support this feature request because:

- it helps users to find local information
- it supports users to collaborate and act locally/regionally. Keep in mind that one of HC's major goals is to help its users to take action!

**User experience:**
Geo-tagging should be optional for users, so IMHO it shouldn't be too complicated.

**Data fields:**
In order to use the location information properly in queries (for searches, maps, filtering, personal dashboard, etc.) a geo-location field would be required. Notice that geo-location will be used with other HC features such as user profile, events and projects already.

**Example:**
No need to reinvent the weel! Please see #266 how the French FOSS social network "Communecter" does it.

**Milestone:**
NitroAlpha Enhancements is fine

appinteractive:
Just as a reminder, we already have profile locations. By that you could find already users nearby. Locations are more relevant for events and Projekts then posts. 🤔

sushidave:
@appinteractive Thanks for your opinion. Please mind that this feature request is not about helping users to find users nearby but to better find relevant information. 

[...]

PR #2059, opened 2019-10-28
🍰 First Implementation Of Filtering Posts By Language

### LINKED ISSUE #2045: 🚀 [Feature] Filter Posts by language
## :rocket: Feature

grafik

I think that entering HC as non German speaker may cause the impression of a pure German Network, because until now nearly all the posts you see are in German.
To avoid this, it should be possible to filter the post by language, and may be set the default of the filter to the selected language.

### Design & Layout
As indicated in the screenshot. May be it is better placed below the emotes.

### Additional context
Thinking about the possiibilty to save filters.

Mogge:
## 🍰 Pullrequest
grafik

First implementation.

### Issues

- fixes #2045

### Todo

- [ ] Finding a way to display flags of the languages. See comments of #2045 .
- [ ] Add tests.

### DIFF
--- a/webapp/components/FilterPosts/FilterPosts.spec.js
+++ b/webapp/components/FilterPosts/FilterPosts.spec.js
@@ -3,6 +3,9 @@ import VTooltip from 'v-tooltip'
 import Styleguide from '@human-connection/styleguide'
 import Vuex from 'vuex'
 import FilterPosts from './FilterPosts.vue'
+import locales from '~/locales'
+import orderBy from 'lodash/orderBy'
+
 const localVue = createLocalVue()
 
 localVue.use(Styleguide)
@@ -12,6 +15,8 @@ localVue.use(Vuex)
 let mutations
 let getters
 
+const languages = orderBy(locales, 'name')
+
 describe('FilterPosts.vue', () => {
   let mocks
   let propsData
@@ -20,6 +25,8 @@ describe('FilterPosts.vue', () => {
   let environmentAndNatureButton
   let democracyAndPoliticsButton
   let happyEmotionButton
+  let englishButton
+  let spanishButton
 
   beforeEach(() => {
     mocks = {
@@ -54,6 +61,8 @@ describe('FilterPosts.vue', () => {
       'posts/RESET_CATEGORIES': jest.fn(),
       'posts/TOGGLE_CATEGORY': jest.fn(),
       'posts/TOGGLE_EMOTION': jest.fn(),
+      'posts/TOGGLE_LANGUAGE': jest.fn(),
+      'posts/RESET_LANGUAGES': jest.fn(),
     }
     getters = {
       'posts/isActive': () => false,
@@ -64,6 +73,7 @@ describe('FilterPosts.vue', () => {
       'posts/filteredCategoryIds': jest.fn(() => []),
       'posts/filteredByUsersFollowed': jest.fn(),
       'posts/filteredByEmotions': jest.fn(() => []),
+      'posts/filteredLanguagesCodes': jest.fn(() => []),
     }
     const openFilterPosts = () => {
       const store = new Vuex.Store({ mutations, getters })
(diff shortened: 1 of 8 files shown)


Tirokk:
Hey @Mogge !!! Great work! 🚀💫
Thanks a lot for your contribution. 👍🏼

That looks and functions very good in my opinion @Mogge . Even if some things are still missing.
Tests for example. I added a checkbox to the PRs description.

I was first not aware that I have to scroll down in the filters menu to see the buttons.

And it is kind of discussable if this is at least the right place to have it. Or should it be a general setting of the users profile? @datenbrei

webapp/locales/en.json
```diff
@@ -87,7 +87,11 @@
     },
     "followers": {
       "label": "Users I follow"
-    }
+    },
+      "language": {
```
Tirokk: Please format with correct indent. ☝🏼

webapp/locales/de.json
```diff
@@ -86,7 +86,11 @@
     },
     "followers": {
       "label": "Benutzern, denen ich folge"
-    }
+    },
+      "language": {
```
Tirokk: Please format with correct indent.

Tirokk:
You have some build errors @Mogge .
Please click on details and then on `The Build`:
See https://travis-ci.com/Human-Connection/Human-Connection/builds/133810500

Tirokk:
And the flags are not absolutely related to the flags of the countries as well. @Mogge @datenbrei 
English language should have the English or the USA flag as an example … ?

Additional we need related countries on the posts and be able to filter for them as well. See #782

Mogge:
The problem with the flags is even greater in Spanish. I do not like flags at all. But it would be nice to have a graphical representation for each language, better than the flag icon I use now.
I will check the errors and try to write some tests.
I think it is more important to filter by language than by country by now. We want to have HC attractive for non-German speakers. When they enter the web and only see German post they may think it is only a German community. So may be by default we should show only posts in the chosen language to new users.

datenbrei:
@Tirokk What would a country-specific post be? One about a country, one from a country? Related to a country? From a specific author of such a country? Or an author living in a country? 

First I would stick with just language, later first we have to decide, what we really want to do with it.

datenbrei:
@Mogge and @Tirokk For now I'd take the english flag. It is the english language, spoken all over the world.

Mogge:
I pushed a new version with the tests added. When I ran yarn test I had no errors.

webapp/components/FilterPosts/FilterPosts.spec.js
```diff
@@ -97,13 +107,31 @@ describe('FilterPosts.vue', () => {
       expect(mutations['posts/TOGGLE_CATEGORY']).toHaveBeenCalledWith({}, 'cat4')
     })
 
+    it('calls TOGGLE_LANGUAGE when clicked', () => {
+      const wrapper = openFilterPosts()
+      englishButton = wrapper
+        .findAll('button.language-buttons')
+        .at(languages.findIndex(l => l.code === 'en'))
```
mattwr18: I really like your way of doing this! one thing I complain about with vue-test-utils over enzyme for example is that there is not `filterWhere`... but I really like how you've done this... I need to refactor some of my tests that look for the button in a position, which is easy to break if someone adds another button, for example.

webapp/components/FilterPosts/LanguageFilterMenuItems.vue
```diff
@@ -0,0 +1,103 @@
...
+      <ds-flex-item
+        :width="{ base: '0%', sm: '0%', md: '0%', lg: '3%' }"
+        id="languages-menu-divider"
+      />
+      <ds-flex v-for="language in locales" :key="language.code" class="languages-menu">
+        <ds-flex class="languages-menu">
+          <ds-flex-item width="100%" class="language-menu-item">
+            <ds-button
+              class="language-buttons"
+              icon="flag"
```
mattwr18: As I stated in the comments, I would prefer using the same country codes for these buttons as the `LocaleSwitch`, using flags of countries is problematic, as discussed, but also the same icon doesn't make a lot of sense to me either.

Mogge:
grafik

As agreed with @mattwr18 I changed the flag icon to the language code.
A pinned post in a language that is not in the filter stays pinned and visible to the user.
I think about a solution for the language filter in the user profile page, as it is not supposed that a user changes the language filter a lot. The idea is adding a `languageFilter property to the user, which is always applied when filtering posts.

webapp/components/FilterPosts/LanguageFilterMenuItems.vue
Mogge: grafik

roschaefer:
@mattwr18 are we good to go here?

mattwr18:
I would be ok to merge this, but if I am being picky @Mogge, I do ask myself why the buttons are rectangular when all the other filter buttons are circular... maybe I am being pedantic... what do you think @alina-beck?

alina-beck:
@mattwr18 @Mogge I thought the same when I saw this... We _are_ picky. :grin: But right now the style guide only allows `icons` for circular buttons so unless we want to save all language codes as svg icons I don't see that this can be solved. I am, however, planning to add this functionality when migrating the styleguide – and buttons will be one of the first things on my list! :)

PR #2062, opened 2019-10-28
Send only one notification for mention and comment

### LINKED ISSUE #1953: 🐛 [Bug] Two Notifications for just one Comment
## :bug: Bugreport
If somebody comments on my post and also mentions my with @myself then I get two notifications for just one comment.

double

1. A user writes a comment and mentions me
2. I get two notifications (see above)

### Expected behavior
For every comment arriving there should be just one notification.

KapilJ22:
### Issues

- fixes #1953

### Todo

- [X] None


KapilJ22:
> I just wanted to let you know I have had a look, and tested it out @KapilJ22
> I would like to discuss this in our daily standup before approving since we need to make a decision on which reason to prioritize with the notification...

Sounds good @mattwr18 !

KapilJ22:
> good enough
> 
> Great! @KapilJ22 from my commit message:
> 
> ```
> refactor: small refactoring for readability
> 
> @KapilJ your solution is right now the best solution, I think. Probably
> the ideal solution would be if we could implement the `CreateComment`
> resolver in such a way that it is doing eager loading of the
> `comment->post->author` relationship and resolving a commment object
> which has the required objects all set. Then you wouldn't have to
> refetch all the stuff.
> 
> But I think this is OK for now :+1:
> ```

Thanks for your review and refactoring @roschaefer - you are awesome :).

PR #2073, opened 2019-10-29
1931 - after successful login the saved language of the user is set

### LINKED ISSUE #1931: 🚀 [Group] [Feature]  Language setting for the user
## :rocket: Group - Feature
#  Language setting for the user 
after successful login the saved language of the user must set

# DB: User.language 

- at registration
 - [x] When registering the user the set language must be saved  #1929

- and in the user settings
- [x] the user can change his language setting to the user setting  #1930

[x] after successful login the saved language of the user is set

ogerly:
## 🍰 Pullrequest
after successful login the saved language of the user is set

### Issues GROUP
- fixes #1931

### DIFF
--- a/webapp/pages/login.vue
+++ b/webapp/pages/login.vue
@@ -26,7 +26,7 @@ export default {
   },
   methods: {
     handleSuccess() {
-      this.$i18n.set(this.user.locale)
+      this.$i18n.set(this.user.locale || 'en')
       this.$router.replace(this.$route.query.path || '/')
     },
   },
--- a/webapp/pages/login.vue
+++ b/webapp/pages/login.vue
@@ -25,8 +25,8 @@ export default {
     }
   },
   methods: {
-    handleSuccess() {      
-      this.$i18n.set(this.user.locale)   
+    handleSuccess() {
+      this.$i18n.set(this.user.locale)
       this.$router.replace(this.$route.query.path || '/')
     },
   },
--- a/webapp/pages/login.vue
+++ b/webapp/pages/login.vue
@@ -25,12 +25,8 @@ export default {
     }
   },
   methods: {
-    handleSuccess() {
-      const currentLocale = this.$i18n.locale()
-      const userSettingLocale = this.user.locale
-      if (currentLocale !== userSettingLocale) {
-        this.$i18n.set(userSettingLocale)
-      }
+    handleSuccess() {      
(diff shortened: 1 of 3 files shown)


webapp/store/auth.js
```diff
@@ -46,6 +46,9 @@ export const getters = {
   termsAndConditionsAgreed(state) {
     return state.user && state.user.termsAndConditionsAgreedVersion === VERSION
   },
+  locale(state) {
+    return state.locale
```
mattwr18: should we have a fallback here @ogerly ? most of the getters I have seen in or store have `|| null`, etc...

webapp/pages/login.vue
```diff
@@ -7,19 +7,30 @@
...
   asyncData({ store, redirect }) {
     if (store.getters['auth/user'].termsAndConditionsAgreedVersion === VERSION) {
       redirect('/')
     }
   },
   methods: {
     handleSuccess() {
+      const currentLocale = this.$i18n.locale()
+      const userSettingLocale = this.user.locale
+      if (currentLocale !== userSettingLocale) {
```
roschaefer: I think the `if` statement is quite unnecessary here, isn't it?

webapp/pages/login.vue
```diff
@@ -7,19 +7,30 @@
...
+        this.$i18n.set(userSettingLocale)
+      }
```
roschaefer: How about just setting:
```js
        this.$i18n.set(this.user.locale)
```
And that's it?

webapp/store/auth.js
roschaefer: @ogerly I think you can remove this. You're consuming `this.user.locale` so this code here is never used.

webapp/pages/login.vue
```diff
@@ -7,19 +7,26 @@
...
+  },
   asyncData({ store, redirect }) {
     if (store.getters['auth/user'].termsAndConditionsAgreedVersion === VERSION) {
       redirect('/')
     }
   },
   methods: {
-    handleSuccess() {
+    handleSuccess() {      
+      this.$i18n.set(this.user.locale)
```
roschaefer: What about a default here?
```suggestion
      this.$i18n.set(this.user.locale || 'en')   
```

mattwr18: @ogerly can you please add this?

ogerly:
sorry

PR #2077, opened 2019-10-29
[WIP] 🍰 Refactor reported content

vbelolapotkov:
## 🍰 Pull request
- Move reported content into its own component
- Replace table with a responsive list

### Issues
- relates #1999 
- relates #1954 

### Todo
- [ ] Display reports list for each resource
- [ ] Cleanup styling


vbelolapotkov:
@Tirokk I've created this PR as an early suggestion for UI refactoring and layout improvements to make the view responsive. Please let me know what do you think and whether it makes sense to continue in that direction.

Tirokk:
@vbelolapotkov Hey, wonderful that you stepped into this. 🚀💫😎

Unfortunately I already refactored the moderators report/decisions list a week ago. And it is still in progress to get more complex.
Was kind of a bad timing as @roschaefer broad up this idea.

I would love to get with you on this when my refactoring takes place in an excepting way.
Then we can have a pairing?
What you think?

vbelolapotkov:
@Tirokk thanks for the update. I assumed something like that so decided to pause after the first step. 

I like the idea of pairing so just let me know when you're finished with refactoring. Have fun with refactoring and let me know if I can help somehow ;)

P.s. I'd close this PR as it is not relevant anymore.

Issue #1851, opened 2019-10-09
🚀 [Feature] Make Tags below Posts  clickable as Hashtags are


datenbrei:
## :rocket: Feature
You can click on a hashtag, but not on the tags below the post, which are made out of them.

Also the **Tag below should be clickable, as the hashtag is**.

### Design & Layout
![tags](https://user-images.githubusercontent.com/13457963/66474220-239a8f00-ea91-11e9-92d6-828547b58b4f.png)

### Additional context

![Bildschirmfoto 2019-10-25 um 22 03 29](https://user-images.githubusercontent.com/25344101/67616070-99626280-f7d4-11e9-97ee-6bdf0619fd20.png)

Should have links like here:

![Bildschirmfoto 2019-10-26 um 09 43 55](https://user-images.githubusercontent.com/25344101/67616136-64a2db00-f7d5-11e9-92c6-65403c9b3f7d.png)

mcdead154:
Hi, i would like to work on this issue.

DennisHack:
Hi mcdead154, 
yes, sure! Would be good to make an appointment for an short briefing together on Zoom. @datenbrei 
Thanks in advance! 
Dennis

mcdead154:
Sure, I've made an account on Zoom. Briefing would be good for me to have a better understanding of issue and application in overall.

Tirokk:
@mcdead154 are you still interested to work on this issue?

mcdead154:
Yes, I would like to work on this.

roschaefer:
@mcdead154 are you still on this issue? Because I would like to use it as an exercise for some computer science students tomorrow.

mcdead154:
@roschaefer yes I work on this I just waited for some response since @DennisHack wrote we will have a briefing. But I'll make a PR on this in this sprint.

datenbrei:
Hi @mcdead154 , I'm sorry for this, I didn't get and get a notification although I have switched on that I get notifications for discussions I take part on.

For this minor feature we don't need a briefing. It is just, that the same happens, if you click on a tag below the article as if you click on the hashtag in the article itself.

mattwr18:
closed by https://github.com/Human-Connection/Human-Connection/pull/2091

PR #2091, opened 2019-10-30
1851 tags clickable

mcdead154:
## 🍰 Pullrequest

Added feature, clickable tags below post, same as hashtags are.
### Issues

- relates #1851 

### Todo

- [X] None
image


webapp/components/Hashtag/Hashtag.vue
```diff
@@ -1,5 +1,7 @@
 <template>
-  <ds-tag>#{{ id }}</ds-tag>
+  <ds-tag>
+    <a v-bind:href="`/?hashtag=${id}`" target="_blank">#{{ id }}</a>
```
roschaefer:
```suggestion
    <nuxt-link :to="hashtagUrl">#{{ id }}</nuxt-link>
```
and implement a `hashtagUrl` computed property :wink:

mcdead154:
@roschaefer I have tried nuxt link, but hashtag opens a new tab on click, and I haven't find a way for nuxt link to open a blank tab, so I have used an <a> tag instead

roschaefer:
`nuxt-link` does not need to open in a new tab. It's used when we link within our app. If we link to somewhere outside of our app, we would use `<a href="...">`. Among other things it does link prefetching and we would loose that optimization.

roschaefer:
It might be that the editor uses `<a>` tags. But that's not on purpose. It's because we did not figure out yet how to render `nuxt-link`s inside an editor node.

PR #2100, opened 2019-10-31
lower case the email

### LINKED ISSUE #2057: 🐛 [Bug] Password-Reset/Registration Dialogue is Case-Sensitive
## :bug: Bugreport
When users want to reset their passwords or want to register and they use emails with uppcercase in it, these can not be found and you get an email which says:  _You requested a password reset but unfortunately we couldn't find an account associated with your e-mail address. Did you maybe use another one when you signed up?_

wrong-email

1. Open reset Password dialogue
2. Type in your emial with upper case letters

Screenshot - 2019-10-29T171429 064

Screenshot - 2019-10-29T171929 698

### Expected behavior
The given email for reset should be converted to lowercase

### Version & Environment
any

Kachulio1:
## 🍰 Pull request

Sets the user email to lower case to avoid case sensitive mismatch 

### Issues

- None

### Todo

- [X] None

### DIFF
--- a/backend/src/schema/resolvers/passwordReset.js
+++ b/backend/src/schema/resolvers/passwordReset.js
@@ -34,7 +34,8 @@ export default {
   Mutation: {
     requestPasswordReset: async (_parent, { email }, { driver }) => {
       const nonce = uuid().substring(0, 6)
-      return createPasswordReset({ driver, nonce, email })
+      const lowerCasedEmail = email.toLowerCase()
+      return createPasswordReset({ driver, nonce, lowerCasedEmail })
     },
     resetPassword: async (_parent, { email, nonce, newPassword }, { driver }) => {
       const session = driver.session()


backend/src/schema/resolvers/passwordReset.js
```diff
@@ -34,7 +34,8 @@ export default {
   Mutation: {
     requestPasswordReset: async (_parent, { email }, { driver }) => {
       const nonce = uuid().substring(0, 6)
-      return createPasswordReset({ driver, nonce, email })
+      const lowerCasedEmail = email.toLowerCase()
+      return createPasswordReset({ driver, nonce, lowerCasedEmail })
```
roschaefer: This is not going to work as `email` and not `lowerCaseEmail` gets desctructured here: https://github.com/Human-Connection/Human-Connection/blob/master/backend/src/schema/resolvers/passwordReset.js#L5

Kachulio1: Yep, good catch will fix that.

roschaefer: https://github.com/Human-Connection/Human-Connection/pull/2118 :point_left:

roschaefer:
I'm going to close this in favour of #2118 hope you don't mind @Kachulio1 ?

PR #2101, opened 2019-10-31
1747 show shouts

### LINKED ISSUE #1747: 🚀 [Feature] Article's Shout displayed in Profile or not
## :rocket: Feature
If somebody shouts an article it should be possible to decide, if this shout should be also displayed in the profile or not.

### Additional context
This is a privacy option. Somebody could just shout non-political articles for example so that nobody could guess his political direction from it.

Kachulio1:
## 🍰 Pull request
Allows the user to set shouts either public or private

### Issues
- fixes #1747 

### Todo
- [X] None


alina-beck:
Hey @Kachulio1! Matt said you want some frontend support on this? Do you have any implementation details in mind already?

I think we would need to:
- [ ] add this to the settings page
- [ ] toggle the `shouts` tab on the user profile depending on the setting
- [ ] display a list of shouted posts when the `shouts` tab is selected

Did I miss anything?

Kachulio1:
@alina-beck sorry have been tied up with personal things. sorry for the late response.
should we sync on Monday about things I should work on the backend?

alina-beck:
Hey @Kachulio1, no problem! I think it's done, actually. We already did fetch `shouted posts` because we had the feature before, so I just needed to toggle them on the profile. And the `mutation` you set up worked fine. :+1:

backend/src/schema/resolvers/users.js
```diff
@@ -177,6 +177,7 @@ export default {
         'termsAndConditionsAgreedVersion',
         'termsAndConditionsAgreedAt',
         'allowEmbedIframes',
+        'showShoutsPublicly',
```
mattwr18: Very important that we not forget to add this! Great job catching this issue before it bites us @kachulio

PR #2108, opened 2019-11-01
Fix: Poll Interval

### LINKED ISSUE #2102: 🐛 [Bug] Notifications query ignores poll interval and queries infinitely
## :bug: Bugreport
When you switch the language, you get infinite requests of the notifications. As if the notifications query would ignore the polling interval (I knew the poll interval would bite us again :disappointed: ). If you refresh the page, it's back to normal.

1. Visit https://nitro-staging.human-connection.org/
2. Login with an account
3. Open developer tools, switch to network tab
4. Switch the language
5. See way too many requests

@datenbrei this could very well be the reason for our high CPU utilization recently.

### Expected behavior
Notifications query might be refetched at most once until it's back to the normal poll frequency.

### Additional context
2019-11-01-004335_1920x1080_scrot

ogerly:
## 🍰 Pullrequest
Fix: Poll Interval

### Issues
- fixes #2102

### DIFF
--- a/webapp/components/notifications/NotificationMenu/NotificationMenu.vue
+++ b/webapp/components/notifications/NotificationMenu/NotificationMenu.vue
@@ -78,9 +78,7 @@ export default {
       query() {
         return notificationQuery(this.$i18n)
       },
-      pollInterval() {
-        return NOTIFICATIONS_POLL_INTERVAL
-      },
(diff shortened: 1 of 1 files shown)


ogerly:
I feel so much more :)

PR #2109, opened 2019-11-01
Fix:2107 Block User on MainPage Popup

### LINKED ISSUE #2107: 🐛[Bug] block the main page of a user in the popup
## :bug: Bugreport
In the listing of all contributions, when poppping up over a user name, one cannot block the user. 
please set a function on the button that the user can be blocked at this place.

Bildschirmfoto von 2019-11-01 08-53-43

ogerly:
## 🍰 Pullrequest
on the overview page of the posts / home page. a user can be blocked directly in the popup. 

### Issues
- fixes #2107

### DIFF
--- a/webapp/components/User/User.vue
+++ b/webapp/components/User/User.vue
@@ -79,7 +79,7 @@
             />
           </ds-flex-item>
           <ds-flex-item :width="{ base: 1 }">
-            <ds-button fullwidth>
+            <ds-button fullwidth @click="block(user)">
               <ds-icon name="user-times" />
             </ds-button>
           </ds-flex-item>
@@ -98,6 +98,7 @@ import HcFollowButton from '~/components/FollowButton'
 import HcBadges from '~/components/Badges'
 import HcAvatar from '~/components/Avatar/Avatar.vue'
 import Dropdown from '~/components/Dropdown'
+import { Block } from '~/graphql/settings/BlockedUsers'
 
 export default {
   name: 'HcUser',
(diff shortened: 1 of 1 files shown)


webapp/components/User/User.vue
```diff
@@ -148,6 +149,13 @@ export default {
       this.user.followedByCount = followedByCount
       this.user.followedByCurrentUser = followedByCurrentUser
     },
+    async block(user) {
+      await this.$apollo.mutate({ mutation: Block(), variables: { id: user.id } })
+      this.$toast.success(user.name + 'geblockt')
```
alina-beck: can we use a translation string instead of `geblockt`?

webapp/components/User/User.vue
```diff
@@ -148,6 +149,13 @@ export default {
...
+      setTimeout(() => {
+        this.$router.go()
```
alina-beck: what is happening here @ogerly? why is the router involved?

webapp/components/User/User.vue
```diff
@@ -79,7 +79,7 @@
             />
           </ds-flex-item>
           <ds-flex-item :width="{ base: 1 }">
-            <ds-button fullwidth>
+            <ds-button fullwidth @click="block(user)">
```
roschaefer: Honestly, I would like to get this button completely removed. It's enough to have it on the user profile page.

This button exists since the early days of our nitro code base. It never worked and this has been discovered just a few days ago. Apparently not too many users expect the block button in the popup.

alina-beck: Oh, didn't know that! I'm fine with removing it – @ogerly what do you think?

ogerly: dennis has asked me to do this. Please discuss this with dennis.

ogerly:
Delete Block Button in MainPage PostCards >>> https://github.com/Human-Connection/Human-Connection/pull/2193

PR #2118, opened 2019-11-04
Fix: Email is Case-Sensitive

### LINKED ISSUE #2057: 🐛 [Bug] Password-Reset/Registration Dialogue is Case-Sensitive

roschaefer:
## 🍰 Pullrequest
This was much more difficult than expected because of all those places where we use emails.

### Issues

- fix #2057

### Todo

- [x] Use backend server response for submit messages (it's kind of done, one of the graphql mutation is not responding with the created object)


webapp/components/PasswordReset/Request.vue
```diff
@@ -68,8 +69,11 @@ export default {
     }
   },
   computed: {
+    email() {
+      return normalizeEmail(this.formData.email)
+    },
     submitMessage() {
-      const { email } = this.formData
+      const { email } = this
```
roschaefer: I think this is the graphql mutation which has Boolean as return type. Thus we cannot take the normalized email from the backend but have to normalize it client-wise.

PR #2140, opened 2019-11-06
Thumbnail in the contribution overview change from 2000px to max-height 357.33px

ogerly:
## 🍰 Pullrequest
Thumbnail in the contribuion overview with change from 2000px to max-height 357.33px
in the article itself you can see the picture in full size

FireShot Capture 177 - Human Connection - Human Connection - localhost

# before it looked like this

FireShot Capture 178 - Human Connection - Human Connection - localhost

### DIFF
--- a/webapp/pages/index.vue
+++ b/webapp/pages/index.vue
@@ -220,7 +220,7 @@ export default {
 
 <style lang="scss">
 .ds-card-image img {
-  max-height: 2000px;
+  max-height: 357.33px;
   object-fit: contain;
 }
 


mattwr18:
we made this change to add 2000px after discussing this with myself, @alina-beck, @DennisHack
it could possibly be decreased, but certainly not down to 350px... I guess we need to discuss further

ogerly:
ok

Tirokk:
We have decided to have big pictures in the newsfeed and in the post itself @ogerly .
For me it looks nice and I love it !!!

Very rear people are doing that. So let them their freedom …

ogerly:
close

Issue #2074, opened 2019-10-29
:star2: [EPIC] Styleguide Migration


alina-beck:
## :star2: EPIC – Styleguide Migration

We want to transition from relying on [Nitro Styleguide](https://github.com/Human-Connection/Nitro-Styleguide) to having all design tokens and components within our main repo. The following items need to be migrated:

### Design Tokens
- [ ] icons
- [ ] logos
- [ ] color: used ~ 52 times in 12 files
- [ ] text-color: used ~ 13 times in 8 files
- [ ] background-color: used ~ 20 times in 6 files
- [ ] border-color: used ~ 4 times in 3 files
- [ ] border-size: used ~ 6 times in 2 files
- [ ] border-radius: used ~ 10 times in 7 files
- [ ] font-size: used ~ 7 times in 5 files
- [ ] font-space: used ~ 1 times in 1 file
- [ ] font-family: used ~ 1 times in 1 file
- [x] font-weight: not used
- [x] line-height: not used
- [x] letter-spacing: not used
- [x] opacity: not used
- [ ] space: used ~ 63 times in 14 files
- [ ] size: used ~ 33 times in 2 files
- [ ] box-shadow: used ~ 7 times in 6 files
- [x] duration: not used
- [x] ease: not used
- [ ] z-index: used ~ 4 times in 4 files
- [x] media-query: not used

### Layout Components
- [ ] ds-card: used ~ 55 times in 43 files
- [ ] ds-container: used ~ 20 times in 17 files
- [ ] ds-flex + ds-flex-item: used ~ 127 times in 31 files
- [ ] ds-grid + ds-grid-item: used ~ 9 times in 5 files
- [ ] ds-modal: used ~ 4 times in 4 files
- [x] ds-page: not used
- [x] ds-page-title: not used
- [ ] ds-placeholder: used ~ 2 times in 2 files
- [ ] ds-section: used ~ 6 times in 6 files
- [ ] ds-space: used ~ 112 times in 43 files
- [ ] ds-spinner: used ~ 4 times in 3 files

### Typography Components
- [x] ds-chip: not used
- [x] ds-code: not used
- [ ] ds-heading: used ~ 20 times in 18 files
- [ ] ds-icon: used ~ 33 times in 21 files
- [ ] ds-logo: used ~ 2 times in 2 files
- [ ] ds-tag: used ~ 4 times in 4 files
- [ ] ds-text: used ~ 66 times in 18 files

### Data Input Components
[...]

PR #2159, opened 2019-11-07
Migrate design tokens

alina-beck:
## 🍰 Pullrequest
Migrates all design tokens from the Nitro Styleguide into the main repo and includes them in Storybook. Even adds a tab so you can see them there:
Screenshot 2019-11-07 at 11 11 50

### Issues
- relates #2074


webapp/storybook/config.js
```diff
@@ -22,6 +22,20 @@ Vue.component('v-popover', {
   template: '<div><slot>Popover Content</slot></div>',
 })
 
+// Setup design token addon
+const scssReq = require.context('!!raw-loader!~/view/styles', true, /.\.scss$/)
+const scssTokenFiles = scssReq
+  .keys()
+  .map(filename => ({ filename, content: scssReq(filename).default }))
```
roschaefer: That's quite interesting. I wonder in which places we could re-use that. Did you copy this from somehwere @alina-beck?

alina-beck: Yep, this came from the [design token addon](https://github.com/UX-and-I/storybook-design-token) for storybook.

I don't think it's possible to change token values _in_ the storybook interface but you should be able to change them in the `.scss` file and see the change reflected in storybook immediately.

PR #2160, opened 2019-11-07
🍰 2119-Fix Contribution  consistent form input validation

### LINKED ISSUE #2119: 🚀 [Feature] Create Post – consistent form input validation
## :rocket: Feature
Validate all fields of the `create post` form in a consistent way. (Ideally, the same system can then be applied to all forms on the network.)

Screenshot 2019-11-05 at 10 44 14

### User Problem
It is unclear which parts of the `create post` form are required and what kind of input is valid. Many users don't know that they have to select `categories` or that a video by itself is not enough content (see #1876).

### Implementation
I would like it if the `title` field and the main `content` editor could be validated with visible feedback _after_ the user entered their input and left the field (i.e. onBlur). Validating input too early (when the user just started typing) or too late (when the user tried to save the post) can lead to frustration.

Open questions / challenges:
- Since images are optional they probably don't need validation?
- It is unclear if the third-party editor allows for the behavior proposed above
- We need a clear message to explain to users that the length of the URL of an embed link does not count toward the minimum content length
- We need to come up with a solution for `categories`
- Is the `language` required? Should it be pre-filled?
- In addition to the validation (after user input) we could also mark required fields (before user input) – but how?

### Design & Layout
The visual feedback for `title` and `content` could look like this:
Screenshot 2019-11-05 at 10 52 39

Question: Do we also want to give feedback for valid content (e.g. a green border and a checkmark)?

### Validation (How can we make sure that this feature solves the above problem?)
- no more bug reports about videos as posts not being accepted
- we could measure if users are faster when creating posts (are any analytics tools set up?)
- developers can see if new contributors are less confused when creating a post in a pairing session

ogerly:
## 🍰 Pullrequest
Fix consistent form input validation

 
 
ezgif com-video-to-gif (2)

# Behaviour 1
FireShot Capture 188 - Human Connection - Human Connection - localhost

# Behaviour 2
FireShot Capture 187 - Human Connection - Human Connection - localhost

# Behaviour 3
FireShot Capture 186 - Human Connection - Human Connection - localhost

### Issues
- fixes #2119


Tirokk:
Hey, chick @ogerly ,
I will make a review tomorrow morning.
The only thing I say right away to the design is:
I would only make left align of the language menu and not a right align.

ogerly:
@Tirokk  I pulled the select bar over the whole box so it comes across better. 
on mobile largest it also runs over the whole box. and the importance is even more clear. 
i can do it left justified and 50% flex
i needed it a bit bigger because of the german text that was broken in the beginning and stood over

ogerly:
this PR is closed and here it goes on https://github.com/Human-Connection/Human-Connection/pull/2189

PR #2169, opened 2019-11-08
1967 component tests content view

### LINKED ISSUE #1967: 🤖 [TESTS] Add component tests for ContentMenu
## :robot: Tests

#AddMoreComponentTests

Mogge:
### Issues
- fixes #1967

### Todo

- [X] There is a lot of repeating code, that can be condensed.


webapp/components/ContentMenu/ContentMenu.spec.js
```diff
@@ -0,0 +1,543 @@
...
+        expect(items).toHaveLength(1)
+        expect(items.at(0).text()).toBe('report.contribution.title')
+        expect(
+          items
+            .at(0)
+            .find('span.ds-menu-item-link')
+            .attributes('to'),
+        ).toBe('/report-contribution-title')
+        items.at(0).trigger('click')
+        expect(openModalSpy).toHaveBeenCalledWith('report')
```
mattwr18: instead of setting up the `method` openModal as a spy, can we set up `$store.commit` as a mock, and check that is called with specific data?

webapp/components/ContentMenu/ContentMenu.spec.js
```diff
@@ -0,0 +1,543 @@
...
+        expect(wrapper.emitted('showEditCommentMenu')).toEqual([[true]])
+        expect(items.at(1).text()).toBe('comment.menu.delete')
+        expect(
+          items
+            .at(1)
+            .find('span.ds-menu-item-link')
+            .attributes('to'),
+        ).toBe('/comment-menu-delete')
+        items.at(1).trigger('click')
+        expect(openModalSpy).toHaveBeenCalledWith('delete')
```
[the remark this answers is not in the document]
Mogge: I reduced the expectations to only one per it block. Now I am only checking that the methods are called correctly, and skipped all the tests of text, etc.

webapp/components/ContentMenu/ContentMenu.spec.js
Mogge: I tried, but it is getting very complicated because y also have to mount the Modal component. I think we have to rely on the tests of the Modal component that the commits are called correctly (which as far as I can see is the case). I ensure in the tests of ContentMenu that the parameters are passed to Modal correctly.

Mogge:
I do not have a good feeling with using .at(number) to grab the menu items I want to test. This can end up in some painful work when new items are added. Can we find some convention to pass a properly defined class to the items, which has the only purpose to identify the item in the tests? Is there a way to add these classes only if we are testing? Does a component somehow know if it is tested or running in production=

mattwr18:
hey @Mogge can we pair on this? it would be nice to get it merged in, no?

mattwr18:
when can we pair on this @Mogge ... I seem to not be available when you are and vice versa...

Mogge:
As `.filter` returns an `WrapperArray` object, I have to chose the first with `.at(0)` to have an wrapper object to trigger the click. I do not understand how the example in ContributionForm.spec.js can work. How can you trigger a click on an WrapperArray?

webapp/components/ContentMenu/ContentMenu.spec.js
mattwr18: I can see that, and as we discussed today, you are testing it up to one point, and there are other tests that take it further!!

PR #2194, opened 2019-11-11
Add donation status and button

### LINKED ISSUE #2040: 🚀 [Feature] Donation Button Directly On The Social Network
## :rocket: Feature
We want to put a little info graphic showing the monthly donation status along with a donation button on the network to encourage more people to donate.

### Problem
We need to make users aware that we depend on them and their support through donations. We also want to show them how dire the situation can be to create an appropriate sense of urgency.

### Implementation
- We will add a page to the admin interface where admins can enter numbers for the `total required donations` and `incoming donations`.
- We will place a `progress bar`, `the above numbers` and a `donate now button` on top of the index page, next to the sorting dropdown.
- The progress bar should represent the relation between the `total` and the `incoming` numbers. It should not go higher than 100%.
- The button should lead to the `donations` page on https://human-connection.org.
- On mobile we could a) alternate between the button and the progress bar, b) show the button below the progress bar or c) display just the button (whichever is easiest, for now!)

Open question: How will the numbers be stored in the database?

### Design & Layout
Here is a quick mockup:
Screenshot 2019-11-07 at 10 53 31

Different from this mockup we agreed that we will place the button _next to_ the progress bar and we'll of course use the same button design as everywhere else on the network.

### Validation
- more donations coming in! :smile:

### Additional context
In the future we could think about displaying this only to users who are not yet donating monthly because it might be annoying otherwise. On the other hand, it might motivate people to increase their monthly donations if they see how badly we need it. 😬

alina-beck:
## 🍰 Pullrequest
Adds donation status and call to action button on `index`:
Screenshot 2019-11-11 at 15 10 49

Adds donation info input fields to admin pages:
Screenshot 2019-11-11 at 15 09 59

### Issues
- fixes #2040 

### Todo
- [x] store values in database
- [x] use values from database in `DonationInfo` component
- [x] add tests for `DonationInfo` and `ProgressBar` component
- [x] mobile design


mattwr18:
great stuff @alina-beck... I left a comment for you on my latest commit about initializing the values of goal and progress, tell me what you think...
I couldn't finish this unfortunately, but maybe you could let me know what you think and take over?
I created a `Donations` query so that you can load them in `pages/index.js` to pass to `DonationInfo` or query for them in `DonationInfo`, if you want...
The `UpdateDonations` is working from the `pages/donations` as far as I can tell, but there are still some clean up....

webapp/components/DonationInfo/DonationInfo.vue
```diff
@@ -0,0 +1,66 @@
+<template>
+  <div class="donation-info">
+    <progress-bar :title="title" :label="label" :goal="goal" :progress="progress" />
+    <a href="https://human-connection.org/spenden/">
```
roschaefer:
```suggestion
    <a target="_blank" href="https://human-connection.org/spenden/">
```

backend/src/seed/factories/donations.js
```diff
@@ -0,0 +1,18 @@
...
+      const defaults = {
+        id: uuid(),
+        goal: 15000,
+        progress: 0,
+      }
+      args = {
+        ...defaults,
+        ...args,
+      }
+      return neodeInstance.create('Donations', args)
```
roschaefer: Usually I would prefer singular here. Like factory bot: https://github.com/thoughtbot/factory_bot The file is plural but the model is singular.

I think this model is an exception and I'm OK with it here.

roschaefer:
One thing maybe:

Screenshot - 2019-11-12T164104 495

backend/src/seed/factories/donations.js
mattwr18: yeah, I thought of that as well, struggled with it, and just decided it could be `Donations` since it's not a single donation, but an aggregation of donations.

PR #2200, opened 2019-11-11
2187 language does not change on edit

### LINKED ISSUE #2115: 🐛 [Bug] Strange hyphenation in titles of German posts
## :bug: Bugreport

grafik

There are a lot of examples more to find

1. Change to another langauge (I tried it with Spanish). Check some titles of German posts where hyphenation applies. Or check the title of the pinned post and see how the hypenation switches as you change the language.

### Expected behavior
The hyphenation must be related to the language of the post and not to the language selected by the user.

### Version & Environment
 Type: []      
 - OS: []      
 - Browser: [] 
 - Version []  
 - Device: []  

### Additional context
Further investigation showed that the same problem appears in the text of the posts. If you choose a longer German text and switch to another language, the hyphenation changes.

### LINKED ISSUE #2187: 🐛 [Bug] Selected language is changed when editing a post
## :bug: Bugreport
reported by @infinite-dao at the related issue #1467:

> Seeing the same on https://human-connection.social, I just want to add:
> 1. I had posted a news article with English language setting, it is saved as such
> 2. On reediting this article it displays in gray as @ulfgebhardt reported and the very first option is selected regardless that an English language setting was saved before

> So it needs also a selected check, wich language option has to be selected in relation to the previously saved post article language setting

1. Write a new post
2. Select a language _other than your selected locale_
3. Save the post
4. Edit the post
5. See the language setting changed to your _selected locale_

### Expected behavior
When I select a language for a post I expect it **not** to change automatically when editing the post. I still want to be able to manually change the language.

### Version & Environment
originally found on `production`, reproducible on `develop`

### Additional context
The `language select field` needs to be decoupled from the `locale setting`.

Question: Do we want to pre-select a language, based on the current `locale setting`  or should it be empty and the user forced to select a language manually? Can we make sure to _only_ set the language automatically when no language has been selected yet?

possibly related to #2119

Mogge:
## 🍰 Pullrequest
Language property of post is passed to frontend. When editing post in a different locale the language does not change by default. Added language attributes to post and postcard (article element in html) to have the hyphenation corresponding to the language of the post and not to the chosen locale.

### Issues
- fixes #2187
- fixes #2115 

### ToDo
- [ ] Admins should run a batch job to assign `de` as language to all posts where language is null

### DIFF
--- a/webapp/assets/styles/main.scss
+++ b/webapp/assets/styles/main.scss
@@ -171,8 +171,6 @@ hr {
 }
 
 .hyphenate-text {
-  -webkit-hyphens: auto;
-  -moz-hyphens: auto;
   hyphens: auto;
   overflow-wrap: break-word;
   word-wrap: break-word;
--- a/webapp/assets/styles/main.scss
+++ b/webapp/assets/styles/main.scss
@@ -171,6 +171,8 @@ hr {
 }
 
 .hyphenate-text {
+  -webkit-hyphens: auto;
+  -moz-hyphens: auto;
   hyphens: auto;
   overflow-wrap: break-word;
   word-wrap: break-word;
--- a/webapp/components/PostCard/PostCard.vue
+++ b/webapp/components/PostCard/PostCard.vue
@@ -1,6 +1,6 @@
 <template>
   <ds-card
-    :language="language"
+    :lang="language"
     :image="post.image | proxyApiUrl"
     :class="{ 'post-card': true, 'disabled-content': post.disabled, 'post--pinned': isPinned }"
   >
--- a/webapp/components/PostCard/PostCard.vue
+++ b/webapp/components/PostCard/PostCard.vue
@@ -1,5 +1,6 @@
 <template>
   <ds-card
+    :language="language"
     :image="post.image | proxyApiUrl"
     :class="{ 'post-card': true, 'disabled-content': post.disabled, 'post--pinned': isPinned }"
   >
@@ -116,6 +117,10 @@ export default {
         this.post ? this.$filters.truncate(this.post.title, 30) : '',
         this.deletePostCallback,
       )
+    },
+    language() {
+      // if older posts have language not set I assume German as language
+      return this.post.language ? this.post.language : 'de'
     },
     isPinned() {
       return this.post && this.post.pinnedBy
--- a/webapp/components/PostCard/PostCard.vue
+++ b/webapp/components/PostCard/PostCard.vue
@@ -1,6 +1,6 @@
 <template>
   <ds-card
-    :lang="language"
+    :lang="post.language"
     :image="post.image | proxyApiUrl"
     :class="{ 'post-card': true, 'disabled-content': post.disabled, 'post--pinned': isPinned }"
   >
@@ -117,10 +117,6 @@ export default {
         this.post ? this.$filters.truncate(this.post.title, 30) : '',
         this.deletePostCallback,
       )
-    },
-    language() {
-      // if older posts have language not set I assume German as language
-      return this.post.language ? this.post.language : 'de'
     },
     isPinned() {
       return this.post && this.post.pinnedBy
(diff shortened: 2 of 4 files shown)


webapp/assets/styles/main.scss
```diff
@@ -171,6 +171,8 @@ hr {
 }
 
 .hyphenate-text {
+  -webkit-hyphens: auto;
+  -moz-hyphens: auto;
```
alina-beck: we have postcss/autoprefixer doing that for us :wink:

Mogge: I was not sure if it is necessary. So just kick it, or is it better I kick it and push it up again?

Mogge: Ok. I removed it.

mattwr18:
> I like this, great work @Mogge! 
> 
> Do we really have `posts` without a `language` in the database? @mattwr18 @roschaefer do you think this is an ok workaround?

We do have a lot @alina-beck look:
```
match (post:Post) where post.language is null return count(post);
+-------------+
| count(post) |
+-------------+
| 18883       |
+-------------+
```
And it's probably more likely than not that those posts are in `German`... so I guess it's ok to assume that, and leave it up to `Post` creators to change it if it's wrong...

Mogge:
I assumed that there are a lot of posts without language. I saw that when I was working on the language filter. That is why I decided to set the default language to German as the great majority of the posts so far is in German. Unfortunately I have no idea how to work around this problem. We cannot check 18k posts manually for their language.

webapp/components/PostCard/PostCard.vue
```diff
@@ -1,5 +1,6 @@
 <template>
   <ds-card
+    :lang="language"
```
roschaefer:
```suggestion
    :lang="post.language"
```

webapp/components/PostCard/PostCard.vue
```diff
@@ -117,6 +118,10 @@ export default {
         this.deletePostCallback,
       )
     },
+    language() {
+      // if older posts have language not set I assume German as language
+      return this.post.language ? this.post.language : 'de'
+    },
```
roschaefer: I don't feel comfortable to have a default set to German *in the code base*. If, at all, I would suggest to use English as default. But in fact my suggestion is to remove this default so our users have an incentive to fix their old posts.

webapp/pages/post/_id/_slug/index.vue
```diff
@@ -1,6 +1,7 @@
 <template>
   <transition name="fade" appear>
     <ds-card
+      :lang="language"
```
roschaefer:
```suggestion
      :lang="post.language"
```

webapp/graphql/Fragments.js
```diff
@@ -45,6 +45,7 @@ export const postFragment = lang => gql`
     deleted
     slug
     image
+    language
```
roschaefer: How could we have missed that? :thinking:

webapp/pages/post/_id/_slug/index.vue
```diff
@@ -147,6 +148,10 @@ export default {
         this.deletePostCallback,
       )
     },
+    language() {
```
roschaefer: Same as above :point_up:

webapp/components/PostCard/PostCard.vue
Mogge: I can follow your argumentation, and I would prefer your solution if there were not those 18k posts without language. This would lead to something like `lang="null"` and I am not sure is this is what we want. I think it is up to the team to decide which is the better way to handle this problem.

webapp/components/PostCard/PostCard.vue
Mogge: I agree, that it is no good idea to have this in the code base. I did this as a pragmatic solution for the 18k posts without language property, assuming that they are mainly written in German. New posts should have set the language property always, so the default here does not apply.

Mogge:
I am using Mozilla Firefox 70.0.1
German is active
grafik
Spanish is active
grafik
See wir-klich

Mogge:
Now it is as suggested by @roschaefer 
Please try to patch the database asap. The PR is important because of the missing language property in the fragment

PR #2234, opened 2019-11-13
[WIP] Implement cypress tests

### LINKED ISSUE #2232: 🚀 [Feature] Add cypress tests for pinned posts
## :rocket: Feature

Admins are able to pin posts, and it would be nice if we have the pinned posts feature tested with cypress

### User Problem

Add tests for pinned posts

### Implementation

Create feature files and add cypress tests on them

### Validation

run yarn 'cypress:run'

aonomike:
## 🍰 Pullrequest

- Start implementation of cypress tests for pinned posts
### Issues

- fixes #2232 

### Todo

- [X] Add tests for admin
- [X] Add tests for other users


mattwr18:
should we pair on this @aonomike ??

aonomike:
Hi @mattwr18 yeah we can pair on this...yes

mattwr18:
can you please push up the newest code @aonomike ??

webapp/components/PostCard/PostCard.vue
```diff
@@ -132,7 +132,7 @@ export default {
       )
     },
     isPinned() {
-      return this.post && this.post.pinnedBy
```
roschaefer: I added missing setup to `posts` factory. @mattwr18 @aonomike do we want to change implementation here?

mattwr18: I thought the idea was to use `this.post.pinned` instead of pinnedBy... see line `135` 
The reason is that `pinnedBy` is an author with nested attributes, ie more expensive request.. whereas, pinned is an attribute.
The reason it was `pinnedBy` in the first place is that I hadn't full taken into account the performance on production data, then you jumped in and fixed it by adding the `pinned` attribute, then he forgot to refactor this function.

webapp/locales/en.json
```diff
@@ -415,7 +415,7 @@
     "menu": {
       "edit": "Edit Post",
       "delete": "Delete Post",
-      "pin": "Pin post",
+      "pin": "Pin Post",
```
mattwr18:
```suggestion
      "pin": "Pin post",
```
we agreed to capitalize the first word in a sentence and lowercase all others....

backend/src/schema/resolvers/posts.spec.js
```diff
@@ -682,58 +682,62 @@ describe('UpdatePost', () => {
...
-                },
-                {
-                  id: 'i-was-created-after-pinned-post',
-                  pinnedAt: null,
-                },
-              ],
-            },
-            errors: undefined,
-          }
-          variables = { orderBy: ['pinned_desc', 'createdAt_desc'] }
```
roschaefer: This is different from what our frontend does

mattwr18: of the top of my head, it seems to be exactly what it does, oder?

roschaefer: https://github.com/Human-Connection/Human-Connection/blob/master/webapp/pages/index.vue#L157
nope

mattwr18:
quite quick on the trigger @roschaefer 
don't you want a review on the work you did on this `PR`, or do you approve of it? :rofl:

PR #2243, opened 2019-11-14
Explicitly define our schema, improve performance

roschaefer:
## 🍰 Pullrequest
I hard-coded most of the generated graphql schema. Unfortunately it is not possible to completely turn off schema augmentation because `neo4j-graphql-js` replaces some types which represent join models.

I also removed numerous input arguments and orderings which are known to cause performance issues when called.

And last but not least I replaced the default `pinnedBy` filter which will also be beneficial for performance.

### Issues

- None

### Todo

- [X] None


backend/src/middleware/sluggifyMiddleware.js
```diff
@@ -25,9 +25,5 @@ export default {
       args.slug = args.slug || (await uniqueSlug(args.title, isUniqueFor(context, 'Post')))
       return resolve(root, args, context, info)
     },
-    CreateCategory: async (resolve, root, args, context, info) => {
-      args.slug = args.slug || (await uniqueSlug(args.name, isUniqueFor(context, 'Category')))
-      return resolve(root, args, context, info)
-    },
```
mattwr18: good cleanup... we shouldn't need to create categories, and if we do, it would be an admin request later

backend/src/schema/types/type/Badge.gql
```diff
@@ -3,8 +3,6 @@ type Badge {
   type: BadgeType!
   status: BadgeStatus!
   icon: String!
-  #createdAt: DateTime
-  #updatedAt: DateTime
```
mattwr18: I understand why you want to clean this up, but you haven't lost hope on refactoring to use `DateTime`, right?
have you seen https://github.com/neo4j-graphql/neo4j-graphql-js/issues/272 ??

backend/src/schema/types/type/Category.gql
```diff
@@ -1,13 +1,41 @@
...
+  slug_asc
+  slug_desc
+  icon_asc
+  icon_desc
+  createdAt_asc
+  createdAt_desc
+  updatedAt_asc
+  updatedAt_desc
+  postCount_asc
+  postCount_desc
```
mattwr18: any performance issue with this?

backend/src/schema/types/type/Post.gql
```diff
@@ -1,26 +1,101 @@
...
+  emotions_every: _PostEMOTEDFilter
+}
+
 enum _PostOrdering {
   id_asc
   id_desc
-  activityId_asc
-  activityId_desc
-  objectId_asc
-  objectId_desc
```
mattwr18: just temporarily?

roschaefer: Yes, until we might need it for ActivityPub.

backend/src/schema/types/type/Category.gql
roschaefer: Yes, there are. But we need it for the admin page for tags and categories. Removing this would break the cypress tests.

PR #2249, opened 2019-11-15
build: less verbose output on build server

roschaefer:
I'm annoyed to scroll when I'm debugging the build server logs. I think,
if you run `jest` with an environment variable `CI` set to true, it will
not try to output colorized log and also does not output which test it's
currently running.

That should lead to the desired, smaller log on Travis CI.

### Issues

- None

### Todo

- [X] None


roschaefer:
Works :tada:

PR #2250, opened 2019-11-15
Helper script to install all packages

roschaefer:
### Issues

- None

### Todo

- [X] None

### DIFF
--- a/package.json
+++ b/package.json
@@ -8,6 +8,7 @@
     "nonGlobalStepDefinitions": true
   },
   "scripts": {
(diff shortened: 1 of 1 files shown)


roschaefer:
Ignore the cypress test result, seems when I ran cypress locally, that made it believe that I'm the build server.

mattwr18:
it's cool that we have this, but for me it's much faster to install in parallel... can't you add a command to tell it to do that @roschaefer ??

PR #2257, opened 2019-11-16
build(deps): bump metascraper modules

roschaefer:
## 🍰 Pullrequest
This will update all metascraper dependencies.

I also did the following:
* Moved the snapshots of the backend tests to a folder `backend/snapshots` which is tagged as generated code in `.gitattibutes`
* Removed all `<script>` tags from the snapshots

The snapshots are used in the backend tests for the embeds and metascraper. They won't be considered source code (and thus our primary language will become JavaScript, yay). Second, if you do a `git grep whatever` it is unlikely that you get a massively long line in the search results, because of minified javascript code in the `<script>` tags of those websites.

### Issues

- None

### Todo

- [X] None


roschaefer:
@mattwr18 yes this is code creep. But since I was fixing a test where those snapshots have been used, I think it's OK, what do you think?

mattwr18:
> @mattwr18 yes this is code creep. But since I was fixing a test where those snapshots have been used, I think it's OK, what do you think?

yep, I approved it, just thought I'd ask because you might have had a reason... If we are regularly asking ourselves if we are guilty of code creep, we will limit how often we do it, and if I think it's egregious, I'll request changes. ;)

PR #2258, opened 2019-11-16
[WIP] build(deps): update tippy.js

roschaefer:
## 🍰 Pullrequest

I've spent a considerable amount of time to try/error the correct way of integrating tippy.js into tiptap. Here's what I came up with. It's probably not nice to destroy and recreate the tiptap context menu but I didn't know a better way to do it. The new API of tippy.js does not have `instance.popperInstance.scheduleUpdate` method anymore.

There is the sticky plugin which should reposition the popper menu when the content or the reference object changes. I couldn't find a way to get the desired behaviour since it's calling the `virtualNode.getBoundingClientRect()` 4 times, and in the end, it's always at position 0/0 (top left corner) thus invisible. https://atomiks.github.io/tippyjs/plugins/

The original error message was that you cannot pass a normal object to `tippy(object, { ... })` anymore. https://atomiks.github.io/tippyjs/misc/

### Issues

- None

### Todo

- [X] None


webapp/components/Editor/Editor.vue
```diff
@@ -151,21 +151,33 @@ export default {
...
       this.insertMentionOrHashtag = command
     },
-    updateSuggestionList({ items, query, range, virtualNode }) {
+    updateSuggestionList({ items, query, range, decorationNode, virtualNode }) {
       this.query = this.sanitizeQuery(query)
       this.filteredItems = items
       this.suggestionRange = range
       this.navigatedItemIndex = 0
-      this.$refs.contextMenu.displayContextMenu(virtualNode, this.$refs.suggestions.$el)
+      this.$refs.contextMenu.hideContextMenu()
```
roschaefer: This is destroying and re-creating the menu in order to keep the menu attached to the word under cursor and also to move it closer to the word, if the menu shrinks in size (because there are fewer options to select).

@alina-beck you have been the last developer who worked on the code. Can you give me some feedback on this solution?

alina-beck:
Unfortunately I found several issues... 😔

1. When I start typing while the editor is close to the bottom of the screen I see this:
Screenshot 2019-11-19 at 11 15 11

2. The results don't seem to update while I'm typing.
3. It _looks_ like several options are selected at once and like I can't navigate with arrow keys anymore – it is working, though. But invisible. (E.g., when I type `@an` then hit enter I get `@ana-purdy`, when I press the down arrow first I get `@antonia-vandervort`).
4. When scrolling, the whole tooltip jumps to the top of the page with the arrow still in the wrong position:
Screenshot 2019-11-19 at 11 19 07

5. And, a weird one: When I `@mention` someone and _try_ to navigate in the list it breaks the `#hashtags` – i.e. I just don't see the hashtag list anymore:
Screenshot 2019-11-19 at 11 37 25

alina-beck:
works for me now – awesome @mattwr18! ❤️

mattwr18:
now it's breaking the build though @alina-beck with this beauty
```
CypressError: Timed out retrying: cy.click() failed because this element:
<li class="suggestion-list__item is-selected">@matt-r...</li>
has CSS 'pointer-events: none', inherited from this element:
<div class="tippy-popper" x-placement="top-start" style="position: absolute; z-index: 9999; visibility: visible; transition-duration: 0ms; will-change: transform; top: 0px; left: 0px; transform: translate3d(-24px, 15px, 0px);" x-out-of-boundaries="">...</div>
'pointer-events: none' prevents user mouse interaction.
Fix this problem, or use {force: true} to disable error checking.
```

alina-beck:
@mattwr18 I think we need to fix this in the UI rather than force clicking – we now can select users from the list by keyboard but not by mouse click anymore...

mattwr18:
I agree @alina-beck... I hadn't realized this was an actual issue in the browser for real users... we have several examples of it being an issue for cypress only, so I mistakenly thought this was one of those times....
maybe don't merge till we fix the inter-activeness @roschaefer 
at the moment, if we add back `interactive: true` it completely breaks the `CSS` for the `SuggestionList`, inexplicably

mattwr18:
pushed up the latest code, where I upgraded tippy to `5.14`, went back and looked at the docs, issues, and compared the functionality currently deployed.
What I can say is that in v5 they append the `tippy-popper` to `parent` with interactive true, and this causes known issues with `clipping` https://atomiks.github.io/tippyjs/faq/#my-tooltip-appears-cut-off-or-is-not-showing-at-all
The solutions proposed, adding `appendTo: document.body`, fix the styling, but the position get messed up.
Screenshot from 2020-01-15 14-37-00

I originally thought this was cause it was appending to the body, and attempted to append it to `.editor`, but that failed, also checking out the deployed version, it is appended to body there, so that can't be it...

maybe @alina-beck you have some idea how we can append to body, but position it to appear where we want...??
I can also open another issue on `tippy`, but wanted to check here first cause they will probably want a codepen example, and it would take me some time to reproduce our setup simplified there.

alina-beck:
I thought the whole point of `tippy` is that it pops up in the correct position (i.e. next to the element it is connected/appended to)? If that's not working I don't really see the point of using the package at all...

We could, of course, detect the position of the `editor` (_maybe_ even the cursor) with JavaScript, then calculate the popup position based on that and apply it programmatically – but it's not pretty.

But is it really an issue with `tippy` or is it more a matter of combining `tippy` with the `tiptap editor`?

mattwr18:
that's a good question @alina-beck ... I'm not sure, really, it works fine with version 4, so maybe we could explore other options, create an issue on tippy github, or just leave it on version 4 for now...

alina-beck:
I'd say let's leave it on version `4` for now – and close this PR?

mattwr18:
I agree @alina-beck ... I'll close this, and we can always re-open

Issue #1948, opened 2019-10-18
🐛 [Bug] Search delivers non-exact/fuzzy Matches


datenbrei:
## :bug: Bugreport
When searching in the network there are too many results shown, which have nothing to do with the search term. The search algorithm is too fuzzy:

![fuzzy-search](https://user-images.githubusercontent.com/13457963/67070087-61797080-f17f-11e9-9894-f41caf0e7da6.png)

### Steps to reproduce the behavior
1. Search for 'Künast'
2. Get 'Knast'

### Expected behavior
The search should be less fuzzy

### Version & Environment
 Type: [Desktop]      
 - OS: [Linux]      
 - Browser: [Firefox] 
 - Version [69]  
 - Device: [Notebook]

alina-beck:
no, mr bot – keep it open! :grin:

PR #2262, opened 2019-11-17
🍰 Search For Users

### LINKED ISSUE #1463: 🚀 [Feature] Search for Users/Persons
## :rocket: Feature
reported by @ulfgebhardt 

### Problem
I would like to search for persons. At the moment I only get posts as search results:
image

### Implementation
To the current search functionality we add the following:
- we search our user database for the given keyword – in both the `name` and the `slug`
- we display the found users below the found posts
- we display titles to separate posts from users
- when selecting a `user` result we redirect to this user's profile page

### Design
Something similar to this:

### Additional context
This is **Step 1** in our search implementation – see #2146.

Mogge:
Please take a look at the storybook of SearchInput.

grafik

How do we organize the search results? As two different searches or as one search with added properties to identify the type of result.
I was not able to generate headings for the items using ds-select from the style guide. As you can see, the heading is part of the first item.
Does this mean I have to write a totally new component doing this? Or do we make changes in the style guide?
I would really like to discuss the technical approach.

 - fixes #1463 

### ToDos
- [ ] We need to drop the `fulltext_search` index on `Post` as it's been renamed to `post_fulltext_search`
- [ ] we must add an fulltext_search index for users


Mogge:
Store error: the application attempted to write an object with no provided id but the store already contains an id of User:u3 for this object.

This is related to this:
https://github.com/apollographql/apollo-client/issues/2510
They say the solution is, to add an ID to the query, but as I am querying for a union, I have no ID.

Mogge:
I really need help. Now I added an ID to each search, so that the store can identify each search process, but the error remains the same:
```
Store error: the application attempted to write an object with no provided id but the store already contains an id of User:u6 for this object. The selectionSet that was trying to be written is:
{"kind":"Field","name":{"kind":"Name","value":"author"},"arguments":[],"directives":[],"selectionSet":{"kind":"SelectionSet","selections":[{"kind":"Field","name":{"kind":"Name","value":"name"},"arguments":[],"directives":[]},{"kind":"Field","name":{"kind":"Name","value":"__typename"}}]}}"
    ApolloError bundle.esm.js:63
    queryListenerForObserver bundle.esm.js:1247
    broadcastQueries bundle.esm.js:1559
    broadcastQueries bundle.esm.js:1557
    broadcastQueries bundle.esm.js:1555
    networkResult bundle.esm.js:1161
search.js:94
```
These are my XHR parameters:
```
{"operationName":"findResources","variables":{"query":"Jenn","searchId":"96a3fb45-cee1-483d-930a-1232e1fdd001"},"query":"query findResources($query: String!, $searchId: ID) {\n  findResources(query: $query, limit: 5, searchId: $searchId) {\n    id\n    searchResources {\n      __typename\n      ... on Post {\n        id\n        title\n        slug\n        commentsCount\n        shoutedCount\n        createdAt\n        author {\n          name\n          __typename\n        }\n        __typename\n      }\n      ... on User {\n        id\n        name\n        slug\n        avatar\n        __typename\n      }\n    }\n    __typename\n  }\n}\n"}
```

mattwr18:
hey @Mogge , 
the issue I was referencing https://github.com/Human-Connection/Human-Connection/issues/1948
what do people think about implementing this here

backend/src/schema/resolvers/searches.js
```diff
@@ -0,0 +1,69 @@
...
+    ...record.get('resource').properties,
+  }
+}
+export default {
+  Query: {
+    findResources: async (_parent, args, context, _resolveInfo) => {
+      const { query, limit } = args
+      const { id: thisUserId } = context.user
+      // see http://lucene.apache.org/core/8_3_1/queryparser/org/apache/lucene/queryparser/classic/package-summary.html#package.description
+      const myQuery = query + '*'
```
Mogge: I am not sure about this. It is not meant to search full phrases. If you enter more then one word, than only the last word is searched with the wildcard. The other words must be written completely to be found. I am not sure if this is the purpose of a search.

mattwr18: we can add it back @Mogge , but it was kinda broken for me... so if I search for `not blocked` in the cypress test it would not find any posts with the content `blocked`, only if I search for `blocked` would it work.
I think maybe the only reason it started working as I expected is that with adding just the wildcard at the end of the phrase `not blocked` means that it is only looking for `blocked`, but at least it actually finds results.

backend/src/schema/resolvers/searches.js
```diff
@@ -0,0 +1,69 @@
...
+          thisUserId,
+        })
+        return userTransactionResponse
+      })
+      try {
+        userResults = await readUserTxResultPromise
+      } finally {
+        session.close()
+      }
+      let searchResults = [...postResults.records, ...userResults.records]
```
roschaefer: If we're appending users to posts and not sorting them in one array by relevance, I would prefer a graphql query like this:
```graphql
{
  search(query: "Foobar") {
    posts {
      title
    }
    users {
      name
    }
}
```

backend/src/schema/resolvers/searches.js
```diff
@@ -0,0 +1,69 @@
...
+      const readPostTxResultPromise = session.readTransaction(async transaction => {
+        const postTransactionResponse = transaction.run(postCypher, {
+          query: myQuery,
+          limit,
+          thisUserId,
+        })
+        return postTransactionResponse
+      })
+      try {
+        postResults = await readPostTxResultPromise
```
roschaefer: We can improve performance here by waiting on both promises in parallel.

webapp/components/features/SearchResources/SearchResources.vue
```diff
@@ -0,0 +1,51 @@
...
+        })
+        this.searchResults = findResources
+      } catch (error) {
+        this.searchResults = []
+      } finally {
+        this.pending = false
+      }
+    },
+    clear() {
+      this.pending = false
```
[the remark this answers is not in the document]
mattwr18: this pending is for the `styleguide` `ds-select` I guess

webapp/components/generic/SearchPost/SearchPost.vue
```diff
@@ -0,0 +1,66 @@
...
+            <span class="post-shouted">
+              <b>{{ option.shoutedCount }}</b>
+              <base-icon name="bullhorn" />
+            </span>
+          </ds-text>
+        </ds-flex-item>
+        <ds-flex-item>
+          <ds-text size="small" color="softer" align="right">
+            {{ option.author.name | truncate(32) }} -
+            {{ option.createdAt | dateTime('dd.MM.yyyy') }}
```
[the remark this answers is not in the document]
mattwr18: maybe

backend/src/schema/resolvers/searches.js
mattwr18: I tried that and it threw an error saying that you could not start another transaction if there is already one pending...

Mogge:
The `dateTime` filter in `plugins/vue-filters.js` is localized.

webapp/components/generic/SearchableInput/SearchableInput.vue
```diff
@@ -90,16 +90,14 @@ export default {
...
-      }
-    },
   },
   methods: {
+    isFirstOfType(option) {
+      return (
+        this.options.findIndex(o => o === option) ===
+        this.options.findIndex(o => o.__typename === option.__typename)
+      )
+    },
```
mattwr18: why not @roschaefer ?? it's returning true or false and does not modify internal state
I understood this would make it preferable to be a computed property

backend/src/schema/resolvers/searches.js
roschaefer: Actually, I changed my mind. After looking through the code it seems i would break a lot with an API like this :point_up: 

@Mogge would it be ok if we remove `findPosts` and `findUsers` instead? We're not using anywhere, or do we?

webapp/components/generic/SearchableInput/SearchableInput.vue
roschaefer: The computed property is set on the component. But `isFirstOfType` is calculated for each data entry.

backend/src/schema/resolvers/searches.js
roschaefer: Implemented now.

webapp/graphql/Search.js
```diff
@@ -0,0 +1,27 @@
+import gql from 'graphql-tag'
+
+export const findResourcesQuery = gql`
+  query($query: String!) {
+    findResources(query: $query, limit: 5) {
+      __typename
+      ... on Post {
```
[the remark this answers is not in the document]
roschaefer: Done

backend/src/schema/resolvers/searches.js
mattwr18: I have tested this out with `ökologischer Garten` in the Post content and searching for `öko garten` as you suggested in Discord @Mogge... I can see what you are concerned about.
I have added this line of code back :point_up_2:,  but now it is failing the cypress tests when searching for `not blocked` with the content being `cause I'm not blocked` ... what would you suggest?

Mogge:
Please see the comments in the end of `backend/src/schema/resolvers/searches.js`

backend/src/schema/resolvers/searches.js
```diff
@@ -0,0 +1,108 @@
...
+MATCH (u2:User)-[:IS_IN]->(l2:Location)
+WHERE NOT(u2.name = 'Huey')
+WITH point({longitude: l1.lng, latitude: l1.lat}) AS P1,
+point({longitude: l2.lng, latitude: l2.lat}) AS P2,
+l1.name AS Location1, l2.name AS Location2
+WITH distance(P1, P2) AS Distance,
+Location1 AS Location1, Location2 AS Location2
+ORDER BY Distance
+RETURN Location1, Location2, Distance
+*/
```
mattwr18: I :heart: that you have looked into this and added how some cypher statements might look, but I'd ask that you add this to an opened issue, or open a `[WIP] PR` instead of us merging in comments that are not related to this issue/PR

Sorry, if that seems pedantic, but I hope you understand and support our desire to limit code creep/comments that need to be incorporated or removed later. Thanks :smile_cat:

webapp/components/generic/SearchPost/SearchPost.spec.js
```diff
@@ -0,0 +1,62 @@
...
+      expect(wrapper.find('.search-option-label').text()).toMatch('Post Title')
+    })
+    it('renders post commentsCount', () => {
+      expect(
+        wrapper
+          .find('.search-post-meta')
+          .findAll('span')
+          .at(0)
+          .text(),
+      ).toMatch('3')
```
mattwr18: I think we can improve these tests by making them more resilient... if a `span` is added these tests will potentially fail
I think we can filter the spans for the one that has text `3`, then assert that `commentsCount.exists().toBe(true)`.
I think that's an improvement, don't you?
Or did you try that and it didn't work for some reason?

webapp/components/generic/SearchPost/SearchPost.spec.js
```diff
@@ -0,0 +1,62 @@
...
+          .at(1)
+          .text(),
+      ).toMatch('6')
+    })
+    it('renders post author', () => {
+      expect(
+        wrapper
+          .find('.search-post-author')
+          .text()
+          .replace(/\s+-\s+/, ' '),
```
mattwr18: why is this necessary?

webapp/components/generic/SearchableInput/SearchableInput.vue
mattwr18: While that makes some sense on the face of it @roschaefer, I'm not satisfied with that answer. First, is that your opinion, or are you referencing some documentation that you could share with us so that we don't just have to take your word for it?
I have done some searching myself, and come to this video recommended in the VueJS docs, which doesn't seem to support your idea... https://vueschool.io/lessons/vuejs-computed-properties?friend=vuejs
Again, it stresses, `"When you need to change data, you will use methods. And when you need to change the presentation of existing data, you will use computed properties."`
This is my understanding of when to prefer one over the other, and does not say anything about when it is for an entire component, or for each data entry.

mattwr18: Also, I would draw people's attention to this https://vueschool.io/lessons/vuejs-computed-properties?friend=vuejs
where it talks about the benefit of using a computed property over a method, even though in their example, and our use case, they show the same results, is that `"computed properties are cached based on their reactive dependencies. A computed property will only re-evaluate when some of its reactive dependencies have changed. This means as long as message has not changed, multiple access to the reversedMessage computed property will immediately return the previously computed result without having to run the function again."`

webapp/locales/de.json
```diff
@@ -515,7 +515,12 @@
     "search": {
         "placeholder": "Suchen",
         "hint": "Wonach suchst Du?",
-        "failed": "Nichts gefunden"
+        "failed": "Nichts gefunden",
+        "heading": {
+          "Post": "Beiträge",
+          "User": "Benutzer",
+	  "Tag": "Hashtags"
```
mattwr18:
```suggestion
	    "Tag": "Hashtags"
```

mattwr18: the indentation is incorrect here, and we don't need this in this `PR`, aka code creep

webapp/locales/en.json
```diff
@@ -199,7 +199,12 @@
   "search": {
     "placeholder": "Search",
     "hint": "What are you searching for?",
-    "failed": "Nothing found"
+    "failed": "Nothing found",
+    "heading": {
+      "Post": "Posts",
+      "User": "Users",
+      "Tag": "Hashtags"
```
mattwr18: same as :point_up_2: ... let's add this to your other `WIP` PR, please.

webapp/components/features/SearchResources/SearchResources.vue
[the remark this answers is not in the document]
mattwr18: tried this out today with @Mogge and it didn't work, though I'm not sure why since we are using it elsewhere in the codebase... :woman_shrugging:

alina-beck:
I just wanted to give a review but I get no search results – all `POST` requests respond with the following error: `Failed to invoke procedure 'db.index.fulltext.queryNodes': Caused by: java.lang.IllegalArgumentException: There is no such fulltext schema index: post_fulltext_search`

@Mogge I think this is one of the `ToDo`s you added above – wanna ping me when it's done so I can take a closer look? :wink:

mattwr18:
@alina-beck you must drop the `fulltext_search` index in your database and then run the db_setup again.

I don't have the commands in front of me, but I can add them here a bit later

alina-beck:
hehe, @mattwr18 yes please! that would be helpful :blush:

mattwr18:
@alina-beck sure... 
1) run `CALL  db.index.fulltext.drop("full_text_search")` in localhost:7474
2) run `env NEO4J_USERNAME=neo4j NEO4J_PASSWORD=letmein neo4j/db_setup.sh` from the root path of the repo

This should be sufficient. The reason we need to do this, in case it's not clear to anyone reading this, is because we have renamed the `full_text_search`, since we now have 2. See https://github.com/Human-Connection/Human-Connection/pull/2262/files#diff-32b7e4816631c5ff090a8b8576356b98R24

webapp/layouts/default.vue
```diff
@@ -21,15 +21,7 @@
               :class="{ 'hide-mobile-menu': !toggleMobileMenu }"
             >
               <div id="nav-search-box" v-if="isLoggedIn">
```
alina-beck: could we get rid of this additional wrapper? it's a lot of `divs` in `divs` in `divs`... :confused:
the styles could probably be transferred to the `searchable-input` class instead

webapp/components/generic/SearchableInput/SearchableInput.vue
```diff
@@ -0,0 +1,236 @@
...
+    class="searchable-input"
+    aria-label="search"
+    role="search"
+    :class="{
+      'is-active': isActive,
+      'is-open': isOpen,
+    }"
+  >
+    <div class="field">
+      <div class="control">
```
alina-beck: That's a lot of wrappers! One should be enough to position the `ds-button` and `ds-select`, no? :wink:

alina-beck: I would even prefer this to be an `<input>` rather than a `<div>` but since our beloved styleguide adds a bunch of its own wrapper `divs` it would probably mess things up... :laughing:

webapp/components/generic/SearchableInput/SearchableInput.vue
```diff
@@ -0,0 +1,236 @@
...
+          :auto-reset-search="!searchValue"
+          :placeholder="$t('search.placeholder')"
+          @click.capture.native="isOpen = true"
+          @focus.capture.native="onFocus"
+          @input.native="handleInput"
+          @keyup.enter.native="onEnter"
+          @keyup.delete.native="onDelete"
+          @keyup.esc.native="clear"
+          @blur.capture.native="onBlur"
+          @input.exact="onSelect"
```
alina-beck: Whoa, a lot of things going on here – do we really need all these props? I guess the `icon-right` doesn't do anything anymore, since we have the button on top of it now. Anything else that's maybe just left over from before?

I'm also wondering if some of these methods could be combined into one to make it more readable – does `onFocus` really have a different impact than `onClick`? Should different things happen `onBlur` and on `escape`? And isn't `ds-select` handling e.g. `onDelete` by itself already? You'll know better @Mogge, so if you say it all makes sense as it is I trust you! Just saying it's not really easy to follow and I'd be glad if it could be "cleaned up" a bit. 😅

webapp/components/generic/SearchableInput/SearchableInput.vue
```diff
@@ -0,0 +1,236 @@
...
+            >
+              <hc-user :user="option" :showPopover="false" />
+            </span>
+            <span
+              v-if="option.__typename === 'Post'"
+              :class="{ 'extra-space': isFirstOfType(option), 'flex-span': true }"
+            >
+              <search-post :option="option" />
+            </span>
+          </template>
```
alina-beck: Have you thought about adding the `headings` directly in the `options` array as separate elements? Then you wouldn't need to do all the `isFirstOfType` checking and the CSS tweaks to separate the heading from the first element – instead you'd just add an option with `.__typename = 'Heading'` and display the corresponding component.

webapp/components/generic/SearchableInput/SearchableInput.vue
```diff
@@ -0,0 +1,236 @@
...
+    padding: 8px;
+  }
+  .extra-space {
+    margin-top: 8px;
+    padding-top: 4px;
+  }
+  .flex-span {
+    display: flex;
+    flex-wrap: wrap;
+  }
```
alina-beck: These sound like utility-classes which is not the direction we decided to go in – could you come up with more descriptive names, such as `first-option` or `option-with-heading`?

Also, please use design tokens instead of 'magic numbers' anywhere in the CSS! :dizzy:

webapp/components/generic/SearchableInput/SearchableInput.vue
```diff
@@ -0,0 +1,236 @@
...
+  &,
+  .ds-select-dropdown {
```
alina-beck: Is this intentional or a typo? I think like this you are applying the `transition` and `max-height` to both, the `.searchable-input` and the `.ds-select-dropdown`...

mattwr18:
a lot of this is copy, :spaghetti: , and we'd have to play around with your suggestions, but I don't see why we couldn't do that... it shouldn't prevent it from getting merged on `Friday` so that it's inline to get deployed `Monday` ...

Mogge:
@alina-beck At the beginning I played around with the Heading as a separate option. I cannot remember why exactly, but I skipped that idea for some reason.

webapp/components/generic/SearchableInput/SearchableInput.vue
mattwr18: hey @alina-beck @Mogge and I are pairing on this `PR` review change requests and we agree to leave this for now, as 1) this functionality is copy and pasted from what is already in production, and 2) we will refactor this further when we migrate `Select` to our repo.

the difference between `onFocus` and `@click` is just that `onFocus` calls `clearTimeout`... we could discuss whether this is needed, but certainly not something to block it getting merged in, I think you'd agree?

the difference between `onBlur` and `esc` is that `onBlur does not clear the user's search term because they can accidentally click somewhere besides the search field and then be annoyed at losing their search term. Whereas, if a user types `Esc`, I think it's reasonable to assume they want to clear their search. What do you think?

PR #2288, opened 2019-11-19
Migrate styleguide icons

alina-beck:
## 🍰 Pullrequest
Replaces `<ds-icon>` from the styleguide with a new `<base-icon>` directly in our repo.

### Issues
- relates #2074 

### ToDo
- [x] add Storybook stories for `BaseIcon`
- [x] register `BaseIcon` for component tests
- [x] include a list of icons in Storybook (if possible)
- [ ] find/provide a list of all available icons (via icomoon or the styleguide)

### Background
A few things that are good to know:
- I had to adapt the folder structure we discussed earlier and will now put a `_new` folder into the existing folders when adding new stuff because [Nuxt.js forces a particular folder structure on us](https://nuxtjs.org/guide/directory-structure) and [there is currently no way to change it](https://github.com/nuxt/rfcs/issues/30) (this is also reflected in the docs)
- All components in the `generics` folder that are prefixed with `Base` will be globally available and don't have to be imported. But: changes made to these components will only be visible when rebuilding the frontend! (So when you work on `BaseComponents` make sure to import them like regular components temporarily, to see changes reflected immediately.)
- I only copied the icons we are actually using in the app, so instead of more than 600 we now only include 65 icons in the project. To add more icons when you need them just copy-paste the `.svg` file from the styleguide into the `/assets/_new/icons/svgs` folder! (I'll try to find/set up a good way to search all available icons.)

### Files to review
A ton of files changed just because...
- I replaced `<ds-icon>` with `<base-icon>` --> many `.vue` files can be ignored,
- pasted the `svgs` from the styleguide --> all `.svg` files can be safely ignored,
- and changed the `localVue` import in all test files --> all `.spec.js` files can be ignored

Please have a look at:
- [ ] `webapp/README.md`
- [ ] `webapp/assets/_new/icons/index.js`
- [ ] `webapp/components/Category/index.vue`
- [ ] `webapp/components/DropdownFilter/DropdownFilter.vue`
- [ ] `webapp/components/LocaleSwitch/LocaleSwitch.vue`
- [ ] `webapp/components/User/User.vue`
- [ ] `webapp/components/_new/generic/BaseIcon/BaseIcon.story.js`
- [ ] `webapp/components/_new/generic/BaseIcon/BaseIcon.vue`
- [ ] `webapp/package.json`
- [ ] `webapp/plugins/base-components.js`
- [ ] `webapp/storybook/config.js`
- [ ] `webapp/storybook/webpack.config.js`
- [ ] `webapp/test/fileMock.js`
- [ ] `webapp/test/registerContext.js`
- [ ] `webapp/test/testSetup.js`

That should be all the relevant changes. :wink:


alina-beck:
I spent a lot of time today figuring out how to make the global `BaseComponents` play nicely with our jest tests (the problem was that `jest` doesn't use `webpack` and therefore cannot deal with `require.context`)...

During my research I also found out that it is possible to set `localVue` as a global variable in a single `testSetup` file and then use this variable in all test files without the need to configure it over and over again – as we did before.

I thought about implementing this in a separate PR because this one is getting out of hand and will be a nightmare to review – but then I would have had to import `BaseComponents` in all the test files so they would have been changed anyway. Sorry! 🤷‍♀

webapp/README.md
```diff
@@ -92,15 +92,15 @@ You can then visit the Storybook playground on `http://localhost:3002`
...
-The folder structure we are aiming for is based on the [directory setup proposed by Nuxt.js](https://nuxtjs.org/guide/directory-structure):
+The folder structure we are following is [prescribed by Nuxt.js](https://nuxtjs.org/guide/directory-structure):
 
-- **assets** contains icons, images and logos in `svg` format
-- **components** are the generic building blocks of the app – small, reusable and usually not coupled to state
-- **features** are composed of components but tied to a particular function of the app (e.g. `comment` or `post`)
+- **assets** contains icons, images and logos in `svg` format and all shared SCSS files such as `tokens`
+- **components** separated into two sub-folders:
+    - **generics** are the generic building blocks of the app – small, reusable and usually not coupled to state
+    - **features** are composed of components but tied to a particular function of the app (e.g. `comment` or `post`)
```
mattwr18: I was wondering if we could change up the naming convention here? Or if you could talk about your motives for naming it so? I believe what you are trying to achieve here is to break these into `presentational` vs `container` components, or is there some nuance I'm not seeing?
I know I am familiar with this concept from `React`, but have also found some info about in relation to `Vue`
I just feel like it might be easier for other developers to understand the distinction if it's divided by terms they have a higher likelihood to know. What do you think?

webapp/components/_new/generic/BaseIcon/BaseIcon.vue
```diff
@@ -0,0 +1,46 @@
...
+<script>
+import icons, { iconNames } from '~/assets/_new/icons'
+
+export default {
+  props: {
+    name: {
+      type: String,
+      required: true,
+      validator(value) {
+        return iconNames.indexOf(value) !== -1
```
mattwr18:
```suggestion
        return iconNames.includes(value)
```
could we use this? @roschaefer refactored some legacy code I copied and pasted in to use `includes` instead and it's easier to understand... even for non-developers.

webapp/components/_new/generic/BaseIcon/BaseIcon.vue
```diff
@@ -0,0 +1,46 @@
...
+      const icon = icons[this.name]
+      if (!icon) {
+        return false
+      }
+      /*
+      a Vue component needs a render function,
+      so we check if there is a render function directly on the icon –
+      otherwise we know it is wrapped in icon.default
+      */
+      return icon.render ? icon : icon.default
```
mattwr18: really good use of a comment here... don't think I would have understood what was happening here

webapp/storybook/webpack.config.js
```diff
@@ -32,6 +32,30 @@ module.exports = async ({ config, mode }) => {
     include: path.resolve(__dirname, '../'),
   })
 
+  // load svgs with vue-svg-loader instead of file-loader
+  const rule = config.module.rules.find(
+    r =>
+      r.test && r.test.toString().includes('svg') && r.loader && r.loader.includes('file-loader'),
```
mattwr18: haha `&&` `&&` `&&` `&&` :stuck_out_tongue_closed_eyes:

webapp/test/testSetup.js
```diff
@@ -0,0 +1,16 @@
...
+import InfiniteLoading from '~/plugins/vue-infinite-loading'
+
+global.localVue = createLocalVue()
+
+global.localVue.use(Vuex)
+global.localVue.use(VTooltip)
+global.localVue.use(Styleguide)
+global.localVue.use(BaseComponents)
+global.localVue.use(Filters)
+global.localVue.use(InfiniteLoading)
```
mattwr18: really great that you found this!! and I like your explanation for why it was added to this `PR`, and for taking the time to explain, in general!!
Really nice! it makes it easier for your `PR` reviewer when you explain how it is related to the changes you introduce in your `PR`, and also helps the `PR` creator to question whether it's code creep and should be in a separate `PR`.
It's not always cut and dry, but this is a really good habit for all of us to get into!
Thanks ;)

webapp/README.md
alina-beck: I am happy to change the names to whatever we as a team decide makes the most sense! The reason why I went with `generics` and `features` over `presentational` and `container` components is that I've only ever seen the latter ones used in combination.  E.g., a `CommentList` folder would host a `Container.vue` to fetch and transform the needed data and a `View.vue` responsible for how the data is displayed. (I am not a big fan of this pattern, though.)

If you have a different understanding of the two, please share!

(I would still like to merge this PR soon, so if it gets another approval before we make a decision on the names I'll merge it – we can easily rename the folders in a separate PR. 🙂)

webapp/components/_new/generic/BaseIcon/BaseIcon.vue
alina-beck: Yes, great catch! 💚

webapp/components/_new/generic/BaseIcon/BaseIcon.vue
alina-beck: Neither did I when I copied this bit from the styleguide! 😅 So I did some digging and thought it would be good to share what I found with everyone (including future-me).

webapp/README.md
roschaefer: I'm happy with `generics/` and `features/`, or maybe even `generic/` singular,
to avoid name-clash with generics in Java (or pharmaceuticals). I could also
think of `use-cases/`.

`Features` is certainly better than `container`. You could also call it
`context` my all-time favourite meaningless term. :wink:

webapp/components/AvatarMenu/AvatarMenu.vue
```diff
@@ -12,7 +12,7 @@
         @click.prevent="toggleMenu"
       >
         <hc-avatar :user="user" />
-        <ds-icon size="xx-small" name="angle-down" />
```
roschaefer: Now we don't have to explain what `ds` is. :+1:

webapp/components/CategoriesSelect/CategoriesSelect.spec.js
```diff
@@ -1,9 +1,7 @@
-import { mount, createLocalVue } from '@vue/test-utils'
+import { mount } from '@vue/test-utils'
 import CategoriesSelect from './CategoriesSelect'
-import Styleguide from '@human-connection/styleguide'
 
-const localVue = createLocalVue()
-localVue.use(Styleguide)
+const localVue = global.localVue
```
roschaefer: Do we clean `localVue` at some point?

webapp/plugins/base-components.js
```diff
@@ -0,0 +1,11 @@
+import Vue from 'vue'
+
+const componentFiles = require.context('../components/_new/generic', true, /Base[a-zA-Z]+\.vue/)
+
+componentFiles.keys().forEach(fileName => {
+  const component = componentFiles(fileName)
+  const componentConfig = component.default || component
+  const componentName = component.name || fileName.replace(/^.+\//, '').replace('.vue', '')
+
+  Vue.component(componentName, componentConfig)
```
roschaefer: Quite a hack, but I like it.

webapp/components/CategoriesSelect/CategoriesSelect.spec.js
alina-beck: We don't. But the `setupFiles` create a new instance of `localVue` before running each test file and the `global` variable should only be available in the `test` scope, no?

PR #2293, opened 2019-11-19
Fix Typo Of Issue 2230

### LINKED ISSUE #2230: 🐛 [Bug] Misspelled “their” of profile::invites::description
## :bug: Bugreport
In there is a misspelled “thier” in https://github.com/Human-Connection/Human-Connection/blob/a0fe9cb94cb7d814b42dc46c48bdcd432e676e51/webapp/locales/en.json#L174 and also the “their” seems to refer to the very “somebody” in the title 
https://github.com/Human-Connection/Human-Connection/blob/a0fe9cb94cb7d814b42dc46c48bdcd432e676e51/webapp/locales/en.json#L173 
I wonder about the “their” but I learned that “their” referring to “somebody/someone” is just English (see https://forum.wordreference.com/threads/never-sit-on-someone-elses-blanket-without-their-permission.2245981/). In German is a discussion in [lokalise.com](https://lokalise.com/project/556252725c18dd752dd546.13222042/?view=single&reference_lang_id=640&single_lang_id=666) to term it more neutral, e.g. “Enter the e-mail address for invitation” or so. But we can leave it at “their” and just correct the misspelled word.

- Are there multiple e-mails to be inserted or just one?

Mogge:
### Issues
- fixes #2230

### DIFF
--- a/webapp/locales/en.json
+++ b/webapp/locales/en.json
@@ -171,7 +171,7 @@
     },
     "invites": {
       "title": "Invite somebody to Human Connection!",
-      "description": "Enter thier e-mail address for invitation.",
+      "description": "Enter their e-mail address for invitation.",
       "emailPlaceholder": "E-mail to invite"
     }
   },


Mogge:
As lokalise is pulling from the English locale automatically, it is not possible in the current workflow to edit the English text via lokalise.

PR #2295, opened 2019-11-19
If an admin searches for a user by email, don't crash if no user can be found

### LINKED ISSUE #2294: 🐛 [Bug] Filter user by email leads to 500 error if email can't be found
## :bug: Bugreport

I was in the middle of practicing my lecture. When I wanted to explain the effect of `graphql-shield` I ran into a strange error "e.get is not a function".

1. Login as admin user at https://develop.human-connection.org/login
2. Visit https://develop.human-connection.org/admin/users
3. Search for an email address that does not exist, e.g. "lksdjfl@example.org"
4. See this strange backend response:
```json
{
  "errors": [
    {
      "message": "e.get is not a function",
      "locations": [
        {
          "line": 2,
          "column": 3
        }
      ],
      "path": [
        "User"
      ],
      "extensions": {
        "code": "INTERNAL_SERVER_ERROR"
      }
    }
  ],
  "data": {
    "User": null
  }
}
```

### Expected behavior
```json
{
  "data": {
    "User": []
  }
}
```

roschaefer:
Ideally we would not have this inconsistent edge case that we filter
through the user type. Much better would be a graphql query like:

```graphql
{
  User(filter: { primary_email: { email: "something@example.org" } }) {
    id
    name
    slug
  }
}
```

**EDIT**
I tried to implement the above behaviour and I think it's OK to have this edge case. Reason: you could nest `_UserFilter` indefinitely. So it becomes complex and risky to create a `graphql-shield` rule based on `args.filter.primary_email`.

Keeping it simple stupid and having this one inconsistency is OK I would say.

Fix #2294

### DIFF
--- a/backend/src/middleware/permissionsMiddleware.js
+++ b/backend/src/middleware/permissionsMiddleware.js
@@ -100,7 +100,7 @@ const noEmailFilter = rule({
 const publicRegistration = rule()(() => !!CONFIG.PUBLIC_REGISTRATION)
 
 // Permissions
-const permissions = shield(
+export default shield(
   {
     Query: {
       '*': deny,
@@ -176,5 +176,3 @@ const permissions = shield(
     fallbackRule: allow,
   },
 )
-
-export default permissions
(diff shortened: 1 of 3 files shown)


roschaefer:
@mattwr18 is that better?

mattwr18:
great @roschaefer !! I'll give this a review now :smile_cat:

PR #2305, opened 2019-11-20
Lokalise: Translations update

appinteractive:
This pull request was initiated by Lokalise (user Mogge) at 2019-11-20 13:03:46

PR #2310, opened 2019-11-20
Changed translation must change test :(

Mogge:
## 🍰 Pullrequest
As the login-Button text in the German translation changed from "Einloggen" to "Anmelden" (which is better German) I had to change the test parameters for cypress. I do not like the design of the test, assuming something as static which obviously is not.

I hope this will let pass the pull request #2305

### Todo
- [X] Check if #2305 passes now


Mogge:
FEAR: As I changed the parameter of the test for the new locale, the test will fail because it will test against the old locale. Arggghh

PR #2317, opened 2019-11-21
2253 fix scroll layout issue

### LINKED ISSUE #2253: 🐛 [Bug] Scrolling fails after a while
I know that on Windows AND Ubuntu with CHROMNE Browser, the problem appears.
I recorded it, so you can have a own look:
https://www.youtube.com/watch?v=zopX8OyzaIA&feature=youtu.be

Screenshot:
scroll-bug

I have 64GB RAM, 32GB Graphics Card NVIDIA Quadro 4000

As the problem lasts for many weeks, it is not related to a special version of chrome I think, as I have installed Chrome (newest version) not long ago.

mattwr18:
### Issues

- fixes #2253 (hopefully)


mattwr18:
@alina-beck can we work on the layout together? I feel lost when it comes to the `MasonryGrid`, I could study up on it, but I'd love to use it as an excuse to pair with you, and I think it's ready...
I'm sure there is probably code that needs to be deleted that I added to the frontend, but I think I am fetching the data you wanted `imageAspectRatio`, setting it for new images added, and updating all the `Post` that have images in the database- this needs to be tested on `develop` and `nitro-staging` though.

backend/src/seed/factories/posts.js
```diff
@@ -19,6 +19,7 @@ export default function create() {
         visibility: 'public',
         deleted: false,
         categoryIds: [],
+        imageAspectRatio: 1.333,
```
alina-beck: @mattwr18 I set this to `1.333` instead of `1.0` because this is the actual aspect ratio of the `faker` images – and I needed to see if the layout still jumps when it's the correct ratio. (PS: it doesn't! :grin:)

webapp/components/MasonryGrid/MasonryGrid.vue
```diff
@@ -27,7 +26,14 @@ export default {
...
 </script>
 
-<style>
+<style lang="scss">
+/* dirty fix to override broken styleguide inline-styles */
+.ds-grid {
+  grid-template-columns: repeat(auto-fill, minmax(300px, 1fr)) !important;
+  gap: 16px !important;
+  grid-auto-rows: 20px;
+}
```
alina-beck: These values are actually defined in the styleguide but for some reason they get messed up on the way from the styleguide to our repo, so I just copied them in here until we get to migrating the `grid` component.

webapp/components/MasonryGrid/MasonryGridItem.vue
```diff
@@ -5,15 +5,33 @@
...
+const landscapeRatio = 1.3
+const squareRatio = 1
+const portraitRatio = 0.7
+
+const getRowSpan = aspectRatio => {
+  if (aspectRatio >= landscapeRatio) return 13
+  else if (aspectRatio >= squareRatio) return 15
+  else if (aspectRatio >= portraitRatio) return 18
+  else return 25
+}
```
alina-beck: This whole thing is a hack but it works ok for setting up the initial layout. If there's a math whizz around who has an idea for an actual `formula` to map these values I'd appreciate it. :wink:

mattwr18:
Just having a look at this @alina-beck ... at the very end of the scroll, when there are no posts left... I'm seeing this..
Screenshot from 2019-12-02 09-51-27

alina-beck:
@mattwr18 I updated the seed data so that the `imageAspectRatio` corresponds with the actual image size – now it looks good! :wink:

backend/src/seed/seed-db.js
```diff
@@ -350,15 +350,17 @@ const languages = ['de', 'en', 'es', 'fr', 'it', 'pt', 'pl']
         author: peterLustig,
         id: 'p0',
         language: sample(languages),
-        image: faker.image.unsplash.food(),
+        image: faker.image.unsplash.food(300, 169),
         categoryIds: ['cat16'],
+        imageAspectRatio: 1.777,
```
roschaefer:
```suggestion
        imageAspectRatio: 300/169,
```

backend/src/seed/seed-db.js
```diff
@@ -350,15 +350,17 @@ const languages = ['de', 'en', 'es', 'fr', 'it', 'pt', 'pl']
...
       }),
       factory.create('Post', {
         author: bobDerBaumeister,
         id: 'p1',
         language: sample(languages),
-        image: faker.image.unsplash.technology(),
+        image: faker.image.unsplash.technology(300, 1500),
         categoryIds: ['cat1'],
+        imageAspectRatio: 0.2,
```
roschaefer:
```suggestion
        imageAspectRatio: 300/1500,
```

backend/src/seed/seed-db.js
```diff
@@ -382,8 +384,9 @@ const languages = ['de', 'en', 'es', 'fr', 'it', 'pt', 'pl']
         authorId: 'u1',
         id: 'p6',
         language: sample(languages),
-        image: faker.image.unsplash.buildings(),
+        image: faker.image.unsplash.buildings(300, 857),
         categoryIds: ['cat6'],
+        imageAspectRatio: 0.35,
```
roschaefer:
```suggestion
        imageAspectRatio: 300/857,
```

backend/src/seed/seed-db.js
```diff
@@ -400,8 +403,9 @@ const languages = ['de', 'en', 'es', 'fr', 'it', 'pt', 'pl']
         author: louie,
         id: 'p11',
         language: sample(languages),
-        image: faker.image.unsplash.people(),
+        image: faker.image.unsplash.people(300, 901),
         categoryIds: ['cat11'],
+        imageAspectRatio: 0.333,
```
roschaefer:
```suggestion
        imageAspectRatio: 300/901,
```

backend/src/seed/seed-db.js
```diff
@@ -413,8 +417,9 @@ const languages = ['de', 'en', 'es', 'fr', 'it', 'pt', 'pl']
         author: jennyRostock,
         id: 'p14',
         language: sample(languages),
-        image: faker.image.unsplash.objects(),
+        image: faker.image.unsplash.objects(300, 200),
         categoryIds: ['cat14'],
+        imageAspectRatio: 1.5,
```
roschaefer:
```suggestion
        imageAspectRatio: 300/200,
```

deployment/legacy-migration/maintenance-worker/migration/mongo/.env
```diff
@@ -12,6 +12,6 @@
 # On Windows this resolves to C:\Users\dornhoeschen\AppData\Local\Temp\mongo-export (MinGW)
 EXPORT_PATH='/tmp/mongo-export/'
 EXPORT_MONGOEXPORT_BIN='mongoexport'
-MONGO_EXPORT_SPLIT_SIZE=6000
```
roschaefer: Code creep?

deployment/legacy-migration/maintenance-worker/migration/neo4j/import.sh
```diff
@@ -39,6 +39,7 @@ function import_collection () {
       echo "Import $1 ${CHUNK_FILE_NAME} (${chunk})"
       echo "${NEO4J_COMMAND}" | "${IMPORT_CYPHERSHELL_BIN}" > /dev/null
       # add file to array and file
+      echo "$IMPORT_CHUNK_PATH_CQL_FILE"
```
roschaefer: What is that?

webapp/components/MasonryGrid/MasonryGridItem.spec.js
```diff
@@ -10,23 +9,22 @@ describe('MasonryGridItem', () => {
...
   })
 
-  it('emits "calculating-item-height" when starting calculation', async () => {
+  it('parent emits "calculating-item-height" when starting calculation', async () => {
     wrapper.vm.calculateItemHeight()
     await wrapper.vm.$nextTick()
 
-    const firstCallArgument = wrapper.vm.$parent.$emit.mock.calls[0][0]
-    expect(firstCallArgument).toBe('calculating-item-height')
+    const firstEmittedFunction = wrapper.vm.$parent.__emittedByOrder[0]
```
roschaefer: I told my students not to test against `wrapper.vm` unless there is no other way. Couldn't we test the `MasonryGrid` component and use `mount` then click on a `MasonryGridItem` component and check if the `wrapper.emitted()` contains `'calculating-item-height'`?

webapp/components/MasonryGrid/MasonryGridItem.spec.js
```diff
@@ -10,23 +9,22 @@ describe('MasonryGridItem', () => {
...
+    expect(firstEmittedFunction.name).toBe('calculating-item-height')
   })
 
-  it('emits "finished-calculating-item-height" after the calculation', async () => {
+  it('parent emits "finished-calculating-item-height" after the calculation', async () => {
```
roschaefer: This looks more and more as if we are actually testing the unnamed parent component here.

webapp/components/MasonryGrid/MasonryGridItem.vue
roschaefer: I don't know myself, but how about https://github.com/desandro/masonry? It's JavaScript not CSS but it looks quite small: https://github.com/desandro/masonry/blob/master/masonry.js

roschaefer: Does that help maybe: https://w3bits.com/css-grid-masonry/ ?

webapp/components/TeaserImage/TeaserImage.vue
```diff
@@ -112,11 +112,13 @@ export default {
       this.showCropper = false
       const canvas = this.cropper.getCroppedCanvas()
       canvas.toBlob(blob => {
+        const imageAspectRatio = canvas.width / canvas.height
         this.setupPreview(canvas)
         this.removeCropper()
-        const croppedImageFile = new File([blob], this.file.name, { type: 'image/jpeg' })
+        const croppedImageFile = new File([blob], this.file.name, { type: this.file.type })
```
roschaefer: Does that allow us to save SVG files, too? How is it possible that it didn't complain about GIF or PNG files?

backend/Dockerfile
```diff
@@ -25,4 +25,5 @@ ENV NODE_ENV=production
 COPY --from=build-and-test /nitro-backend/dist ./dist
 COPY ./public/img/ ./public/img/
 COPY ./public/providers.json ./public/providers.json
+COPY ./public/add_image_aspect_ratio.sh ./public/add_image_aspect_ratio.sh
```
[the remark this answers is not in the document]
roschaefer: We have `cypher-shell` only in `neo4j/Dockerfile`. Right now, only the `backend` service has access to `uploads/`. Ouch :weary: I think we have to do that manually.

webapp/components/TeaserImage/TeaserImage.vue
mattwr18: I think it would, although there was an unintended removal of `image/jpeg` from line 119
this change would allow us to support `svg`, which is certainly what we want, but would give remove our compression.
I'll add that back and we can work on a solution that compresses in the backend and supports `svg` files

deployment/legacy-migration/maintenance-worker/migration/neo4j/import.sh
mattwr18: sorry, unintended committed change

deployment/legacy-migration/maintenance-worker/migration/mongo/.env
mattwr18: good call, I had seen this and wanted to revert the change, but then forgot... 
thanks for calling me out!! we certainly want to avoid code creep!

alina-beck:
@roschaefer @mattwr18 the masonry grid solution we were using is exactly the one proposed here: https://w3bits.com/css-grid-masonry/ :wink:

The issue with this (as well as the [masonry package](https://masonry.desandro.com/layout.html) – see `imagesLoaded`) is that it first calculates the size of each item in the grid, _then_ the images get loaded, _then_ it needs to recalculate the size which makes the layout jump.

The solution we came up with in this PR is storing an `imageAspectRatio` in the database and using this (instead of waiting for the actual image) to calculate the size of the post cards.

But there's another catch: Before the `PostCards` are finished rendering with their correct size the `MasonryGridItems` are already displayed. And if their `rowSpan` (i.e. height) is set to `0` you see this for a second:
Screenshot 2019-12-04 at 12 31 43

So what I did is pass the `imageAspectRatio` directly to the `MasonryGridItem` to _estimate_ the space needed for the `PostCard` and avoid big jumps in the layout when (a moment later) the `PostCard` has finished rendering – and this is where my (practically non-existent) math skills fail me:

For a width of `300`...
- an aspect ratio of `2` means a height of `150`
- an aspect ratio of `1.5` means a height of `200`
- an aspect ratio of `1` means a height of `300`
- an aspect ratio of `0.5` means a height of `600`
- and with an aspect ratio of `0.2` we land at a height of `1.500`

So it's not a linear relation – I can't just say for each `.5` change add `2` rows to the height. I'm sure there is some kind of formula that would allow us to do this more accurately than having 3 fixed ratios to compare to, like I did above, but it will always be an estimation at this point because we don't yet have the actual `width` to calculate with. So unless one of you has such a formula up their sleeves I'd say let's give this solution a go? :grin:

PS: I'll try to improve the tests before we merge!

PR #2333, opened 2019-11-22
refactor: Remove obsolete code about invitation codes

roschaefer:
Since we decided to open Human Connection for public registration, we
don't need to create artificial shortage of signups. The original
intention here was too:

1. Grow the network slowly and under control
2. Create an incentive for users to invite their friends

Point 2 is now obsolete, everyone can signup herself. Point 1 is based
on the (seemingly) wrong assumption that there will be a "run" on the
network if we open it for public registration. Our growth was fairly
stable so far.

### Issues

- None

### Todo

- [X] None


roschaefer:
@alina-beck Here's the code which we talked about recently. It works, there are backend tests that check the desired behaviour. But the frontend code was missing and so we never implemented that feature before we rushed the network out to the public.

PR #2335, opened 2019-11-22
fix: Re-enable webfinger feature

roschaefer:
Ok, so here is the plan. Let's give both our cucumber features and your
cypress tests a prominent place to live. That would be the root level
folder of our application. Second, let's revive formerly dead code step
by step.

Ie. move code from the former location `backend/features/` to `features/`
when it is ready. All edge cases should be tested with unit tests in
`backend/`, see my `webfinger.spec.js` as an example.

### Issues

- None

### Todo

- [X] None


datenbrei:
"WebFinger is a protocol specified by the Internet Engineering Task Force IETF that allows for discovery of information about people and things identified by a URI.[1] Information about a person might be discovered via an "acct:" URI, for example, which is a URI that looks like an email address."

@roschaefer What data privacy concerns do we have to address here?

.travis.yml
```diff
@@ -8,13 +8,13 @@ addons:
...
 
-before_install:
+install:
   - yarn global add wait-on
   # Install Codecov
   - yarn install
   - cp cypress.env.template.json cypress.env.json
 
-install:
+before_script:
```
mattwr18: how is this and the change :point_up_2: related to this `PR`? 
is this some setup that should be changed anyway, ie code creep, or is this somehow related to running the cucumber tests?

backend/src/activitypub/routes/webfinger.js
```diff
@@ -0,0 +1,59 @@
...
+        type: 'application/activity+json',
+        href: `${CONFIG.CLIENT_URI}/activitypub/users/${name}`,
+      },
+    ],
+  }
+}
+
+export async function handler(req, res) {
+  const { resource = '' } = req.query
+  // eslint-disable-next-line no-unused-vars
```
mattwr18: why would this fail lint? you are using `name` and `domain` on line 26??

backend/src/activitypub/routes/webfinger.spec.js
```diff
@@ -0,0 +1,117 @@
+import { handler } from './webfinger'
+import Factory from '../../seed/factories'
+import { getDriver } from '../../bootstrap/neo4j'
+
+let resource
+let res
+let json
+let status
+let contentType
```
mattwr18: Is there any reason you prefer multi-line `let` statements over one-liners?
Just a question...

backend/src/activitypub/routes/webfinger.spec.js
```diff
@@ -0,0 +1,117 @@
...
+    it('sends HTTP 400', async () => {
+      await request()
+      expect(status).toHaveBeenCalledWith(400)
+      expect(json).toHaveBeenCalledWith({
+        error: 'Query parameter "?resource=acct:<USER>@<DOMAIN>" is missing.',
+      })
+    })
+  })
+
+  describe('?ressource query param', () => {
```
mattwr18:
```suggestion
  describe('?resource query param', () => {
```

backend/src/activitypub/routes/webfinger.spec.js
```diff
@@ -0,0 +1,117 @@
...
+        })
+      })
+    })
+
+    describe('with acct:', () => {
+      beforeEach(() => {
+        resource = 'acct:some-user@domain'
+      })
+
+      it('returns empty json', async () => {
```
mattwr18: seems to return an error still, oder?

backend/src/activitypub/routes/webfinger.spec.js
```diff
@@ -0,0 +1,117 @@
...
+          expect(json).toHaveBeenCalledWith({
+            links: [
+              {
+                href: 'http://localhost:3000/activitypub/users/some-user',
+                rel: 'self',
+                type: 'application/activity+json',
+              },
+            ],
+            subject: 'acct:some-user@localhost:3000',
+          })
```
mattwr18: really exciting stuff!!!

backend/src/activitypub/routes/webfinger.js
roschaefer: It's the `_` that's being unused.

backend/src/activitypub/routes/webfinger.spec.js
roschaefer: No, I think a linter with a previous configuration complained about it.

backend/src/activitypub/routes/webfinger.spec.js
roschaefer: Genau

.travis.yml
roschaefer: Code creep, but a mild one. Starting the docker containers is more a `before_script` whereas installing e.g. `cucumber` is part of the `install` in my eyes.

backend/src/activitypub/routes/webfinger.js
mattwr18: that shouldn't fail linting... it's kind of the purpose to use the `_` in this way

PR #2353, opened 2019-11-25
Add shoutedBy_some to _PostFilter

### LINKED ISSUE #2352: 🐛 [Bug] showShoutsPublicly functionality broken
## :bug: Bugreport

One can go to `privacy` and check a box saying they would like to show their shouts publicly, but a user trying to see their posts cannot

1. go to the `AvatarMenu` dropdown at the top right
2. click the settings link
3. click the privacy link
4. check the box to enable publicly sharing shouts, and save
5. log out, and log in as another user
6. go to the first user's profile
7. click on the shouts tab
8. see the error

### Expected behavior

The workflow should not be broken

### Version & Environment
 Type: []      
 - OS: []      
 - Browser: [] 
 - Version []  
 - Device: []

mattwr18:
- we have a feature that allows users to publicly show their shouts with
their consent. we need this for the filter to work.

### Issues

- fixes #2352 

@roschaefer I believe you removed this from the _PostFilter when you manually added them, but we merged in functionality that requires this filter. please review

### DIFF
--- a/backend/src/schema/types/type/Post.gql
+++ b/backend/src/schema/types/type/Post.gql
@@ -19,6 +19,7 @@ input _PostFilter {
   title_not_starts_with: String
   title_ends_with: String
   title_not_ends_with: String
+  shoutedBy_some: _UserFilter
   slug: String
   slug_not: String
   slug_in: [String!]


mattwr18:
I think it's better to err on the side of removing more than having some filters we don't need @roschaefer ... just glad we caught it before it was deployed

PR #2404, opened 2019-12-02
Get rid of inconsistency with neode setup

roschaefer:
## 🍰 Pullrequest

    Implement getDriver() and getNeode() the same way

    I want to introduce `neode` and `neo4j-graphql-js` to my students and
    saw this.
------
    Fix cypress tests

    Since `dotenv` in the backend now loads configuration from a file
    relative to the current file, we don't have to duplicate the neo4j
    configuration settings here.


backend/src/middleware/hashtags/hashtagsMiddleware.spec.js
```diff
@@ -36,7 +36,7 @@ beforeAll(() => {
     context: () => {
       return {
         user: authenticatedUser,
-        neode: instance,
+        neode: neode,
```
mattwr18:
```suggestion
        neode,
```
can't we use ES6 syntax here?

backend/src/schema/resolvers/users/blockedUsers.spec.js
```diff
@@ -20,7 +20,7 @@ beforeEach(() => {
       return {
         user: authenticatedUser,
         driver,
-        neode: instance,
+        neode: neode,
```
mattwr18:
```suggestion
        neode,
```
same as :point_up_2:

PR #2421, opened 2019-12-04
Remove data-test attriubutes in non-dev env

### LINKED ISSUE #2420: 🚀 [Feature] Remove data-test attributes in production
## :rocket: Feature

We would like to use `data-test` attributes for testing when there is no existing CSS class used for styling.
### User Problem

Sometimes it's challenging to target an element without finding all the elements and then grabbing one of the elements based on its position. This is not a good practice because it makes them less resilient to changes. Any update to the DOM and your likely to waste others time debugging, build time, and asking others to make your tests more resilient.

It is not a good practice to use CSS classes for testing, even if they are named with `test` in their name. CSS classes should be used for styling your application. If they are present, they can certainly be used to find elements, but should not be added solely for the purpose of testing

Enter `data-test` attributes, which are much less likely to be changed/removed by other developers unless in the context of changing tests. See, https://medium.com/@colecodes/test-your-dom-with-data-attributes-44fccc43ed4b and https://kentcdodds.com/blog/making-your-ui-tests-resilient-to-change

The issue is that these `data-test` attributes get built and make their way to production where they are just clutter. 
### Implementation

We can configure `nuxt` with compile options that remove these attributes during our production build. 

### Validation

Dev can create a production build locally, and inspect the DOM

mattwr18:
### Issues

- fixes #2420 
### Todo

- [X] Build production image locally and inspect DOM to ensure `data-test` attributes have been removed.


webapp/nuxt.config.js
```diff
@@ -272,6 +272,26 @@ export default {
...
+        modules: [
+          {
+            preTransformNode(astEl) {
+              if (!ctx.isDev) {
+                const { attrsMap, attrsList } = astEl
+                tagAttributesForTesting.forEach(attribute => {
+                  if (attrsMap[attribute]) {
+                    delete attrsMap[attribute]
+                    const index = attrsList.findIndex(attr => attr.name === attribute)
+                    attrsList.splice(index, 1)
```
alina-beck: This could be written as:
```
attrsList = attrsList.filter(attr => !tagAttributesForTesting.includes(attr.name))
tagAttributesForTesting.forEach(attribute => delete attrsMap[attribute])
```

Instead of finding the `index` of each attribute you could just filter the list once. And `delete` just returns `true` if the attribute doesn't exist, so we don't really need the `if statement`. :wink:

webapp/nuxt.config.js
```diff
@@ -272,6 +272,26 @@ export default {
...
+                  }
+                })
+              }
+              return astEl
```
alina-beck: What does `astEl` stand for? Could we use a more declarative name?

mattwr18: I'm not sure... this was a copy :spaghetti: and once it was working, I thought about how it could be improved, but didn't spend much time cause I need to build the image each time to make sure it works.... 
I learned about `delete` from @roschaefer actually. We have some cases in the backend already, really cool stuff!!
this is what `astEl` looks like 
```
{                                                                                                                                                               13:35:28
  type: 1,
  tag: 'signup',
  attrsList: [
    {
      name: ':invitation',
      value: 'true'
    }
  ],
  attrsMap: {
    ':invitation': 'true'
  },
  rawAttrsMap: {},
  parent: {
    type: 1,
    tag: 'ds-card',
    attrsList: [
      [Object]
    ],
    attrsMap: {
      class: 'signup'
    },
    rawAttrsMap: {},
    parent: {
      type: 1,
      tag: 'ds-section',
      attrsList: [],
      attrsMap: {},
      rawAttrsMap: {},
      parent: undefined,
      children: [Array]
    },
    children: []
  },
  children: []
}
```
Any ideas of names? `El` I guess would be short for `Element`... not sure about `ast`...

mattwr18: Awww, I was able to find out... it stands for `abstractSyntaxTree`... have updated the code to be more verbose... I love it when people ask for changes I could see myself suggesting!! :heart_eyes_cat:

webapp/nuxt.config.js
mattwr18: thanks for approving @alina-beck 
I thought the same thing when I saw the code, filter is great for this...
I tried to get it working now and it didn't though... maybe we can merge and figure out why together. 
one difference in `filter` and `splice` is that the latter changes the array in place whereas the former you need to assign to a new array.
When I tried this :point_up_2: it threw an error in `yarn build` 
```
TypeError: Assignment to constant variable.
```
So, I attempted to make it a let, but it failed to remove the `data-test` attibutes... maybe it would allow to create a new array and then set the `attrsList` array to the new? Not sure.

I tried without the if clause as well, and it built and started, but then threw this error 
```
Cannot read property '_normalized' of undefined
```
So, in the end the only thing I have changed is to make the variable more verbose.
Thoughts?

alina-beck: Interesting... 🙊

You're right, it looks like we're not allowed to assign to `attrsList` which makes `splice` the better (only?) choice. The `_normalized` issue I don't understand – `delete` really shouldn't complain about anything? But well, we have a working solution and it's easy enough to read, so I'm happy to just leave it as is! :smile:

webapp/nuxt.config.js
alina-beck: Thank you for researching! 😍

PR #2423, opened 2019-12-04
Fix issue #2237, comments are 4 times as long before show more

### LINKED ISSUE #2237: 🚀 [Feature] Comments 4-Times As Long Before Show More
## :rocket: Feature

The comments shall be 4-times as long in character amount before the `Show more` button appears.

### User Problem

@DennisHack (our product owner) says that the user to often has to click on `Show more` to see the full comment of someone.

### Implementation

Make the character count four times as high before the `show more` button appears.

### Design & Layout

- none

### Validation

- none

### Additional context

- none

bocajnannerb:
## 🍰 Pullrequest
Comments can now be 1200 characters before being truncated with "show more"

I also wrote some tests since that was specified as a requirement for any fixes, I'm not sure if that was intended for changes this small though, especially since the two tests that I made take a long time to run (>30 seconds on my machine at least)  because they have to type in so much text. I am pretty unfamiliar with the Cypress testing stuff so maybe there is a way to paste text in instead of typing it all out.

commentLengthScreenshot

## Issues
- fixes #2237 

(First time contributing here, if I'm not doing things properly just let me know)


cypress/integration/post/Comment.feature
```diff
@@ -20,3 +20,25 @@ Feature: Post Comment
...
+    """
```
mattwr18: this is really great @bocajnannerb , adding cypress tests is amazing!!
I have a suggestion though that could improve the code a bit... have you heard of `imperative` vs `declarative` with regards to testing/gherkin? Here is a short read about it https://wiki.saucelabs.com/display/DOCS/Best+Practice%3A+Imperative+v.+Declarative+Testing+Scenarios

I try to ask myself in these situations whether the information is necessary for a non-developer who might read this Scenario. Remember that our use of `cucumber` with our cypress means that we are creating executable documentation, where a non-developer/client could have a read of these scenarios and understand what they are meant to do.

So, one question I have is, is it important that the reader of this document know what was typed in the comment? or is it enough to know that it was a comment of longer than 300 characters, but shorter than 1200? or longer than 1200?
If the answer is that it's not important they know exactly what was typed, than we can extract this logic into our `steps` and "hide" it from people who are not developers, thus making it more `declarative`!!

So, you could do something similar to what @ogerly  did here https://github.com/Human-Connection/Human-Connection/blob/master/webapp/components/ContributionForm/ContributionForm.spec.js#L61.

What do you think?

bocajnannerb: Well that explains why it was taking so long to run the integration stuff, I figured I must have done something wrong but didn't know what. I'll have to make a new pull request to do it from a branch, yes? And I will work to get the comment more hidden away, thank you for that recommendation

mattwr18:
closing as superseded by `PR` https://github.com/Human-Connection/Human-Connection/pull/2443 directly to parent repo
thanks @bocajnannerb

PR #2433, opened 2019-12-05
Favor transaction functions

### LINKED ISSUE #2412: 🐛 [Refactor] Favor Transaction Function over Auto-commit transactions
## :zap: Refactor

Related to https://github.com/Human-Connection/Human-Connection/pull/2402

We have been receiving this error message in production `no available threads to serve it at the moment.` on our `neo4j` pod

There is an open issue on `neo4j` GitHub page https://github.com/neo4j/neo4j/issues/12014
where they talk about different causes, one of which is leaky `transactions/statements`.

Currently, we are using mostly `Auto-commit transactions`, which the `Neo4j` docs say, `"It is not recommended to use auto-commit transactions in production environments or when performance or resilience are a primary concern."`. Please see, https://neo4j.com/docs/driver-manual/1.7/sessions-transactions/#driver-transactions-auto-commit.

We should refactor the backend to use `Transaction functions`. See, https://neo4j.com/docs/driver-manual/1.7/sessions-transactions/#driver-transactions-transaction-functions

mattwr18:
### Issues

- fixes #2412 

### Middleware
- [x] permissionsMiddleware.js
- [x] sluggifyMiddleware.js
- [x] validationMiddleware.js

### Resolvers
- [x] locations.js
- [x] comments.js
- [x] createPasswordReset.js
- [x] existingEmailAddress.js
- [x] notifications.js
- [x] passwordReset.js
- [x] posts.js
- [x] rewards.js 
- [x] shout.js
- [x] statistics.js
- [x] user_management.js
- [x] users.js
### Tests
- [x] Add tests for middleware/nodes/locations.js
- [x] passwordReset.spec.js
### Factories
- [x] index.js


mattwr18:
I have to go, but I would love to get a review on this... I think the `validateUpdateUser` is not what we want because it means that we force all users to have a name. At the moment, we have it to where users signing up need to add their name, but in the old alpha it was possible to have no name.

So, we are in the position where we want to validate that users cannot change their names to null or anything less than 3 characters, maybe we check if `params.name === null`? cause if we do `!params.name` then we need to pass in a `name` to every `UpdateUser`, as I've done, but what happens to the `43` users who have null as their user.name?? 

@roschaefer what do you think??

roschaefer:
Fix users with missing data and give them a placeholder. Like: "Unknown"

mattwr18:
> Fix users with missing data and give them a placeholder. Like: "Unknown"

Then, there name would be displayed as `Unknown`... I'm proposing we do nothing about it... if a user from the web UI tries to update their name to an empty string, or anything less than three characters, they will get an error.

The cannot set it to `null` or `undefined` from the webapp, if they know a way to bypass our webapp and send messages directly to the backend, we have more serious issues. @roschaefer

PR #2466, opened 2019-12-09
Paginate moderations page without losing filtering

### LINKED ISSUE #2442: 🐛 [Bug] reports query is too expensive and times out often
## :bug: Bugreport

The reports query fetches all reports, which have nested `FiledReports`, nested `reviews`,  nested `resources`, which have nested attributes as well. 
This makes it an extremely expensive request. Leading to time outs.

1. log in as admin
2. go to `/moderation`
3. see if the reports load
4. ...
5. Profit

### Expected behavior

We'd expect that `reports` would be shown

### Additional context

@roschaefer suggested we could do `eager` loading of certain resources.
We could also try pagination

mattwr18:
Co-authored-by: Tirokk <wolle.huss@pjannto.com>

### Issues

- fixes #2442

### DIFF
--- a/webapp/components/User/User.vue
+++ b/webapp/components/User/User.vue
@@ -8,7 +8,7 @@
   <dropdown v-else :class="{ 'disabled-content': user.disabled }" placement="top-start" offset="0">
     <template slot="default" slot-scope="{ openMenu, closeMenu, isOpen }">
       <nuxt-link :to="userLink" :class="['user', isOpen && 'active']">
-        <div @mouseover="openMenu(true)" @mouseleave="closeMenu(true)">
+        <div @mouseover="openInfoMenu" @mouseleave="closeMenu(true)">
           <hc-avatar v-if="showAvatar" class="avatar" :user="user" />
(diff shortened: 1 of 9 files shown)


webapp/graphql/Moderation.js
```diff
@@ -82,9 +82,6 @@ export const reportsListQuery = () => {
             name
             disabled
             deleted
-            followedByCount
-            contributionsCount
-            commentedCount
```
roschaefer: Awww yeah. That's what I wanted to do in my next PR. I'm going to rebase my code on that.

webapp/components/User/User.vue
```diff
@@ -26,7 +26,7 @@
         </div>
       </nuxt-link>
     </template>
-    <template slot="popover">
+    <template slot="popover" v-if="showCounts">
```
roschaefer: I would say `showPopover` no?

mattwr18: yes, that's better

PR #2470, opened 2019-12-10
Improve notification query performance by reducing db calls

roschaefer:
## 🍰 Pullrequest
We had plenty of database calls for our notification query. That could very well be the source of our recent performance issues in the backend.

### DIFF
--- a/backend/src/schema/resolvers/reports.js
+++ b/backend/src/schema/resolvers/reports.js
@@ -1,3 +1,5 @@
+import log from './helpers/databaseLogger'
+
 const transformReturnType = record => {
   return {
     ...record.get('report').properties,
@@ -16,8 +18,7 @@ export default {
       const { driver, user } = context
       const session = driver.session()
       const reportWriteTxResultPromise = session.writeTransaction(async txc => {
-        const reportTransactionResponse = await txc.run(
-          `
+        const cypher = `
             MATCH (submitter:User {id: $submitterId})
             MATCH (resource {id: $resourceId})
             WHERE resource:User OR resource:Post OR resource:Comment
@@ -27,15 +28,16 @@ export default {
             CREATE (report)<-[filed:FILED {createdAt: $createdAt, reasonCategory: $reasonCategory, reasonDescription: $reasonDescription}]-(submitter)
 
             RETURN report, resource, labels(resource)[0] AS type
-          `,
-          {
-            resourceId,
-            submitterId: user.id,
-            createdAt: new Date().toISOString(),
-            reasonCategory,
-            reasonDescription,
-          },
-        )
+          `
+        const params = {
+          resourceId,
+          submitterId: user.id,
+          createdAt: new Date().toISOString(),
+          reasonCategory,
+          reasonDescription,
+        }
+        const reportTransactionResponse = await txc.run(cypher, params)
+        log(reportTransactionResponse)
         return reportTransactionResponse.records.map(transformReturnType)
       })
       try {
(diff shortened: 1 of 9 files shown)


backend/src/schema/resolvers/reports.js
```diff
@@ -153,14 +155,14 @@ export default {
...
-          { id },
-        )
+        const cypher = `
+          MATCH (resource)<-[:BELONGS_TO]-(report:Report {id: $id})<-[review:REVIEWED]-(moderator:User)
+          RETURN moderator, review
+          ORDER BY report.updatedAt DESC, review.updatedAt DESC
+        `
+        const params = { id }
+        const allReportsTransactionResponse = await tx.run(cypher, params)
+        log(allReportsTransactionResponse)
```
mattwr18: same question as your other `PR` @roschaefer 
why so much refactoring just to add `log`??

roschaefer: I want to see the database calls being made when I set `DEBUG=human-connection:neo4j:*`. So I implemented this helper function which outputs the cypher statement and/or the database call statistics, the counters. We should always call the `log` method from now on when we implement database calls.

mattwr18: that part, I get... I don't understand how why that needed so much refactoring
I have refactored it and pushed it up.
Would this somehow not work with the code you added... 
It would be nice to know how to use as well
If I run `DEBUG=human-connection:neo4j:* yarn dev` it throws this error
```
No matches for wildcard “DEBUG=human-connection:neo4j:*”. See `help expand`.
env DEBUG=human-connection:neo4j:* yarn dev
```
so I'm assuming the wildcard needs to be replaced by something, but what? Ahh 
it takes either `human-connection:neo4j:cypher` or `human-connection:neo4j:stats`??

roschaefer: On fish shell:
```fish
env DEBUG="human-connection:neo4j:*" yarn run dev
```
if you don't escape the `*` your shell believes it's a file glob pattern.

backend/src/schema/resolvers/reports.js
```diff
@@ -83,35 +82,35 @@ export default {
...
+        )
         log(allReportsTransactionResponse)
         return allReportsTransactionResponse.records.map(record => record.get('report'))
       })
       try {
-        const txResult = await reportReadTxPromise
-        if (!txResult[0]) return null
-        reports = txResult
+        const reports = await reportReadTxPromise
+        if (!reports) return []
```
roschaefer: How could `reports` ever be not an array?

roschaefer: If we run into an error I would prefer HTTP 500

backend/src/schema/resolvers/reports.js
```diff
@@ -120,23 +119,24 @@ export default {
...
           submitter: record.get('submitter').properties,
           filed: record.get('filed').properties,
         }))
       })
       try {
-        const txResult = await readTxPromise
-        if (!txResult[0]) return null
-        filed = txResult.map(reportedRecord => {
+        const filedReports = await readTxPromise
+        if (!filedReports) return []
```
roschaefer: Same, the promise is returning an array why this check?

backend/src/schema/resolvers/reports.js
mattwr18: So, revert it to return null?

backend/src/schema/resolvers/reports.js
mattwr18: copy :spaghetti:

backend/src/schema/resolvers/reports.js
mattwr18: great thanks @roschaefer 
then run a query and it will give me some crazy feedback?

mattwr18: ahhh, ok here are the result of running a `reports` query
```
2019-12-10T16:04:20.346Z human-connection:neo4j:cypher 
            MATCH (report:Report)-[:BELONGS_TO]->(resource)
            WHERE (resource:User OR resource:Post OR resource:Comment)
            
            WITH report, resource,
            [(submitter:User)-[filed:FILED]->(report) |  filed {.*, submitter: properties(submitter)} ] as filed,
            [(moderator:User)-[reviewed:REVIEWED]->(report) |  reviewed {.*, moderator: properties(moderator)} ] as reviewed,
            [(resource)<-[:WROTE]-(author:User) | author {.*} ] as optionalAuthors,
            [(resource)-[:COMMENTS]->(post:Post) | post {.*} ] as optionalCommentedPosts,
            resource {.*, __typename: labels(resource)[0] } as resourceWithType
            WITH report, optionalAuthors, optionalCommentedPosts, reviewed, filed,
            resourceWithType {.*, post: optionalCommentedPosts[0], author: optionalAuthors[0] } as finalResource
            RETURN report {.*, resource: finalResource, filed: filed, reviewed: reviewed }
            ORDER BY report.createdAt DESC
             
          
2019-12-10T16:04:20.348Z human-connection:neo4j:cypher {}
2019-12-10T16:04:20.349Z human-connection:neo4j:stats StatementStatistics { _stats: { nodesCreated: 0, nodesDeleted: 0, relationshipsCreated: 0, relationshipsDeleted: 0, propertiesSet: 0, labelsAdded: 0, labelsRemoved: 0, indexesAdded: 0, indexesRemoved: 0, constraintsAdded: 0, constraintsRemoved: 0 } }
2019-12-10T16:04:20.349Z human-connection:neo4j:stats { resultConsumedAfter: 2, resultAvailableAfter: 56 }
```

PR #2487, opened 2019-12-10
Fix broken scroll behaviour on index and profile page

roschaefer:
I don't know where the bug originates. But it can only be that either
`previousResult` or `fetchMore` result is sometimes undefined. This
should make the function bullet-proof for these situations.

### Issues

- None

### Todo

- [X] None

### DIFF
--- a/webapp/components/utils/UpdateQuery.js
+++ b/webapp/components/utils/UpdateQuery.js
@@ -1,0 +1,17 @@
+import unionBy from 'lodash/unionBy'
+
+export default function UpdateQuery(component, { $state, pageKey }) {
+  if (!pageKey) throw new Error('No key given for the graphql query { data } object')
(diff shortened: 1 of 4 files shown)


mattwr18:
maybe, we should test it out in staging? it certainly doesn't seem to hurt, but maybe you want to have another stab at it?

roschaefer:
@mattwr18 I didn't see your comment. This problem is **not** being solved with this PR. Instead, on my browser, sometimes the scroll would hang. That is the case when you have runtime errors during `onUpdate`.

roschaefer:
The problem of your screenshot is not that scrolling simply stops. Instead, new posts are getting added all the time and since they don't have a height, the scrolling gets triggered all the time. Like many times in short succession.

PR #2513, opened 2019-12-14
fix: editor not visible in server-side-rendering

roschaefer:
This is fixing a bug where sometimes the editor would not get displayed
until you click into the title text fields. This commit also removes
some obscure optimizations.

I'm really annoyed by the cruft that we still carry around until this
very day.

Every single line of untested code, which you left because you thought:
"Well sure this is going to improve performance!" is going to bite you.

### DIFF
--- a/webapp/components/Editor/Editor.vue
+++ b/webapp/components/Editor/Editor.vue
@@ -24,7 +24,6 @@
 import { Editor, EditorContent } from 'tiptap'
 import { History } from 'tiptap-extensions'
 import linkify from 'linkify-it'
-import stringHash from 'string-hash'
 import { replace, build } from 'xregexp/xregexp-all.js'
 
 import * as key from '../../constants/keycodes'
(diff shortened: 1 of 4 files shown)


webapp/components/Editor/Editor.vue
```diff
@@ -107,29 +106,7 @@ export default {
...
-      immediate: true,
-      handler: function(val) {
-        if (!val || !this.editor) {
-          return
-        }
-        this.editor.extensions.options.placeholder.emptyNodeText = val
-      },
-    },
-  },
-  created() {
```
mattwr18: code creep? @roschaefer ... why were these changes made in this context? 
this has the undesired side effect that makes the placeholder text not reactive, but also now instead of just approving and merging, we need to discuss what these changes have anything to do with fixing the server side-rendering bug

mattwr18: sorry, I reverted your changes @roschaefer because I know that you are busy today with your talk, and I'd like to get this bug fix in today's deploy. Normally, I would like to discuss this before making the decision, but under the circumstances, I felt it more important to get the bug fix in....

PR #2539, opened 2019-12-17
Update neode

roschaefer:
## 🍰 Pullrequest

    Update neode

    This fixes some API changes with the update of neode.

See https://github.com/Human-Connection/Human-Connection/pull/1913

This also fixes a bug along the way (wrong model name). And we have to format the validation to send them back to the client.


backend/src/schema/resolvers/registration.js
```diff
@@ -40,7 +40,7 @@ export default {
       `,
         { nonce, email },
       )
-      const emailAddress = await neode.hydrateFirst(result, 'email', neode.model('Email'))
+      const emailAddress = await neode.hydrateFirst(result, 'email', neode.model('EmailAddress'))
       if (!emailAddress) throw new UserInputError('Invalid email or nonce')
```
mattwr18: what would we have to do to remove `neode` from production code?
This is one of the only remaining resolvers which uses it...

roschaefer: The fact that @johnymontana considers using neode in neo4j-graphql-js is making me unsure about if we should try to refactor neode out of our production code.

PR #2545, opened 2019-12-17
Lokalise: Translations update

appinteractive:
This pull request was initiated by Lokalise (user Ewald Arnold) at 2019-12-17 17:50:26


ewanold:
Hello @roschaefer  it may look good, but it isn't :-/ for some reason the path on lokalise changed from **webapp/locale/ru.json** to **locale/ru.json** I only noticed because I could not see my changes where I expected them to be.

PR #2546, opened 2019-12-17
Update neo4j-driver

roschaefer:
Fix API changes. Also close the session in resolver only and refactor
userMiddleware to become a part of the resolver.

The fix for the test was to move `session.close` into the finally block at the end of the resolver. It was complaining about using a session in `userMiddleware.js` which was already closed.


backend/src/middleware/user/userMiddleware.js
```diff
@@ -1,16 +0,0 @@
...
-      await createOrUpdateLocations(result.id, args.locationName, context.driver)
-      return result
-    },
-    UpdateUser: async (resolve, root, args, context, info) => {
-      const result = await resolve(root, args, context, info)
-      await createOrUpdateLocations(args.id, args.locationName, context.driver)
-      return result
-    },
-  },
-}
```
mattwr18: why are these changes in this `PR` @roschaefer 
code creep?

backend/src/schema/resolvers/registration.js
```diff
@@ -51,11 +54,14 @@ export default {
...
           emailAddress.update({ verifiedAt: new Date().toISOString() }),
         ])
+        await createOrUpdateLocations(args.id, args.locationName, session)
         return user.toJson()
       } catch (e) {
         if (e.code === 'Neo.ClientError.Schema.ConstraintValidationFailed')
           throw new UserInputError('User with this slug already exists!')
         throw new UserInputError(e.message)
+      } finally {
+        session.close()
```
mattwr18: maybe these make sense here, if so, please explain... 
if not, please open an issue to refactor, then add them there where our future selves can find them more easily

backend/src/schema/resolvers/users/location.js
```diff
@@ -110,44 +105,36 @@ const createOrUpdateLocations = async (userId, locationName, driver) => {
...
+          MATCH (location:Location {id: $locationId})
+          MERGE (user)-[:IS_IN]->(location)
           RETURN location.id, user.id
         `,
-        { userId: userId, locationId: data.id },
-      )
-    })
-  } finally {
-    session.close()
-  }
```
mattwr18: why are you no longer closing the session here in this function?
is it closed later?

backend/src/schema/resolvers/registration.js
roschaefer: The signature of `createOrUpdateLocations` has changed. It is now expecting session and not driver. When I updated I got complaints that a session was already closed. So I re-use the same session and close it at the end.

backend/src/schema/resolvers/users/location.js
roschaefer: See above :point_up:

PR #2565, opened 2019-12-18
Update it.json

### LINKED ISSUE #2356: 🐛 [Bug] Donations slider broken when missing translations
## :bug: Bugreport

If you are using a language that has untranslated phrases/words for the donations section, it doesn't show the amount or any fallback language like English

1. log into your account
2. switch to Italian
3. see that the Donations slider is broken
4. ...
5. Profit

### Expected behavior

It is quite important that this information is always displayed, even if it's in a fallback language

Screenshot from 2019-11-25 19-24-08

erhardtstiefel:
## 🍰 Pullrequest

The missing italian text added by translation from English test

### Issues

Expand the definition of   "moderation": { "name": "Moderation", "reports": { ....

### DIFF
--- a/webapp/locales/it.json
+++ b/webapp/locales/it.json
@@ -7,13 +7,13 @@
         "password": "La tua password",
         "moreInfo": "Che cosa è Human Connection?",
         "hello": "Ciao",
-        "forgotPassword": "",
-        "no-account": "",
-        "register": "",
-        "moreInfoURL": "",
-        "moreInfoHint": "",
-        "success": "",
-        "failure": ""
+        "forgotPassword": "Password dimenticata?",
+        "no-account": "Non hai un account?",
+        "register": "Iscriviti",
+        "moreInfoURL": "https://human-connection.org/en/",
+        "moreInfoHint": "alla pagina di presentazione",
+        "success": "Hai effettuato il login!",
+        "failure": "Indirizzo e-mail o password errati."
     },
     "profile": {
         "name": "Il mio profilo",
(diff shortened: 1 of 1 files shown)


erhardtstiefel:
I would like to have contact to someone responsible of text translations to schedule the next steps

Mogge:
@erhardtstiefel I have the same wish as you. I fear that so far there is nobody responsible for the translations. So we have to create the responsibility.

erhardtstiefel:
Ok, lets meet in Discord. 
 
 
 
 
-------Originalmeldung-------
 
Von: Mogge
Datum: 01/07/20 09:35:37
An: Human-Connection/Human-Connection
Cc: Erhardt Stiefel;  Mention
Betreff: Re: [Human-Connection/Human-Connection] Update it.json (#2565)
 
@erhardtstiefel I have the same wish as you. I fear that so far there is
nobody responsible for the translations. So we have to create the
responsibility.

alina-beck:
@erhardtstiefel @Mogge I just had a quick glance at the translations – I don't really speak Italian but understand some and from what I see some of the machine translations would indeed be helpful while others are plain nonsense. (One example: "following" was translated with "it is the following:" instead of "(he/she is) following" as a verb.)

So, I am not sure if it is better to have no Italian translations or a mixture of helpful and confusing translations... 🤷‍♀ What do you think?

Also, @Mogge you said the `json` is corrupt and would need some cleaning up – I'd definitely say don't put too much work into it! If it's easy to fix I'm fine with either decision.

erhardtstiefel:
I had assumed that my PR would be checked by an Italian speaking developer. If you want to delete my PR, I don't mind. I just wanted to help.

alina-beck:
That's great @erhardtstiefel, thank you! Unfortunately we don't have any Italian speakers on the team right now... :blush:

Mogge:
I agree with closing this PR

PR #2582, opened 2019-12-19
refactor(styleguide): improve emotion buttons and header responsiveness

alina-beck:
## 🍰 Pullrequest
During the migration of the styleguide buttons I found several HTML/CSS issues I just had to fix – couldn’t stop myself… 😊

This PR sums them up:
- replaces the last few instances of `ds-button` with `base-button`
- makes sure the `delete data` button is only enabled while the user name matches the text input (so far it was just enabled once but not disabled again when the user added/removed characters)
- moves `embed` related CSS classes to the `EmbedComponent` file
- fixes `EmotionsButtons` markup (before, the `ds-flex-items` were not having any effect because they were wrapped in additional `divs`)
- removes the `LoadMore` button because it was no longer displayed
- moves the `dropdown-arrow` class to `main.scss` to make sure it loads in the correct order
- improves markup and design of `Paginate` component
- slightly improves header responsiveness by overriding styleguide-inline-styles (cleaning this up properly is on my todo list for next year…)
- refactors the `SearchableInput` component to use less wrapper `divs` and less CSS

### ToDo
- [x] merge PR #2562
- [x] rebase this PR to master
- [ ] approve and merge this one! :wink:


webapp/components/DeleteData/DeleteData.vue
```diff
@@ -55,22 +55,19 @@
           <ds-space margin-bottom="xx-small" />
           <ds-flex :gutter="{ base: 'xx-small', md: 'small', lg: 'large' }">
             <ds-flex-item :width="{ base: '100%', sm: '100%', md: '100%', lg: 1.75 }">
-              <ds-input
-                v-model="enableDeletionValue"
-                @input="enableDeletion"
-                class="enable-deletion-input"
-              />
+              <ds-input v-model="enableDeletionValue" class="enable-deletion-input" />
```
mattwr18: nice change, I think I added this, but having a `v-model` means there is no need for an `@input` since their values are bound

webapp/components/EmotionButton/EmotionButton.vue
```diff
@@ -0,0 +1,56 @@
+<template>
+  <div class="emotion-button">
+    <base-button :id="emotion" circle ghost @click="$emit('toggleEmotion', emotion)">
+      <img :src="emojiPath" />
```
mattwr18: the pics are not showing up...
Screenshot from 2020-01-15 15-21-54

I played around with it, and adding a max width to the `img`, they look good to me 
Screenshot from 2020-01-15 15-43-02
Screenshot from 2020-01-15 15-43-30

I'll push up the code... you can change, if you have a better solution...

webapp/components/Paginate/Paginate.spec.js
```diff
@@ -5,35 +5,31 @@ import Paginate from './Paginate'
...
       it('is disabled by default', () => {
         propsData = {}
         wrapper = Wrapper()
-        nextButton = wrapper.findAll('.ds-button').at(0)
+        nextButton = wrapper.find('[data-test="next-button"]')
         expect(nextButton.attributes().disabled).toEqual('disabled')
       })
 
-      it('is not disabled if hasNext is true', () => {
+      it('is enabled if hasNext is true', () => {
```
mattwr18: better antonym for `disabled` :stuck_out_tongue:

webapp/components/Paginate/Paginate.vue
```diff
@@ -1,26 +1,42 @@
...
+      data-test="previous-button"
+    />
+    <base-button
+      @click="$emit('next')"
+      :disabled="!hasNext"
+      icon="arrow-right"
+      circle
+      data-test="next-button"
+    />
+  </div>
```
mattwr18: I like the new design of the buttons in the `Paginate` component...
I guess you considered renaming it to multi-word following Vue guidelines, but decided not to in this `PR`??
Also, I think the way it is now, it wasn't clear to me that the buttons were disabled until I looked at the design in `admin/users` to see what they look like when there is a `hasNext` or `hasPrevious`... maybe we could just not show them if both those conditions are false? I would be happy to get other opinions/take care of this in a separate `PR`

Screenshot from 2020-01-15 17-01-59

webapp/components/generic/SearchableInput/SearchableInput.spec.js
```diff
@@ -32,81 +32,79 @@ describe('SearchableInput.vue', () => {
...
+      select.trigger('focus')
+      select.element.value = 'abcd'
+    })
 
-      it('opens the select dropdown when focused on', () => {
-        expect(wrapper.find('.is-open').exists()).toBe(true)
-      })
+    it('opens the dropdown when focused', () => {
+      expect(wrapper.find('.ds-select-dropdown').isVisible()).toBe(true)
+    })
```
mattwr18: this is an improvement, for sure in my mind...

webapp/components/generic/SearchableInput/SearchableInput.spec.js
```diff
@@ -32,81 +32,79 @@ describe('SearchableInput.vue', () => {
...
 
-      it('opens the select dropdown and blurs after focused on', () => {
-        select.trigger('blur')
-        expect(wrapper.find('.is-open').exists()).toBe(false)
-      })
+    it('closes the dropdown when blurred', () => {
+      select.trigger('blur')
+      expect(wrapper.find('.ds-select-is-open').exists()).toBe(false)
+    })
```
mattwr18: why couldn't the opposite of :point_up_2: be done?

webapp/components/generic/SearchableInput/SearchableInput.spec.js
```diff
@@ -32,81 +32,79 @@ describe('SearchableInput.vue', () => {
...
 
-      it('changes the unprocessedSearchInput as the value changes', () => {
-        select.trigger('input')
-        expect(select.element.value).toBe('abcd')
-      })
+    it('changes the unprocessedSearchInput as the value changes', () => {
+      select.trigger('input')
+      expect(select.element.value).toBe('abcd')
+    })
```
mattwr18: I know you didn't add this, but this is not testing anything in my mind, and can be simply removed...originally this was testing `wrapper.vm.unprocessedSearchInput`, but @roschaefer said to avoid this cause it can cause false negatives.
maybe we don't need this `unprocessedSearchInput` at all... but we can certainly keep refactoring in the future, especially when we migrate the `Select` component...

webapp/components/generic/SearchableInput/SearchableInput.vue
```diff
@@ -1,60 +1,46 @@
...
-          @click.capture.native="isOpen = true"
-          @focus.capture.native="onFocus"
-          @input.native="handleInput"
-          @keyup.enter.native="onEnter"
-          @keyup.delete.native="onDelete"
-          @keyup.esc.native="clear"
-          @blur.capture.native="onBlur"
-          @input.exact="onSelect"
+  <div class="searchable-input" aria-label="search" role="search">
+    <ds-select
```
mattwr18: no `div` inside a `div` inside another `div`!??

webapp/components/generic/SearchableInput/SearchableInput.vue
```diff
@@ -170,58 +149,32 @@ export default {
...
-    }
-  }
-  .field {
-    width: 100%;
-    display: flex;
-    align-items: center;
-  }
-  .control {
-    width: 100%;
+    right: $space-xx-small;
```
mattwr18: `CSS` magic! down from `52` to `25` lines!!! impressive!

webapp/layouts/default.vue
```diff
@@ -4,30 +4,30 @@
...
               class="mobile-hamburger-menu"
             >
-              <ds-button icon="bars" @click="toggleMobileMenuView" right />
+              <base-button icon="bars" @click="toggleMobileMenuView" circle />
             </ds-flex-item>
             <ds-flex-item
-              :width="{ base: '85%', sm: '85%', md: '50%', lg: '50%' }"
+              :width="{ base: '45%', sm: '45%', md: '45%', lg: '50%' }"
               :class="{ 'hide-mobile-menu': !toggleMobileMenu }"
+              style="flex-shrink: 0; flex-grow: 1;"
```
mattwr18: why did you use inline styling in these cases? something to do with the styleguide?
is this to avoid targeting the class and applying `!important`... just guessing...

webapp/components/EmotionButton/EmotionButton.vue
alina-beck: maybe I should start testing things in Chrome... 😝

webapp/components/Paginate/Paginate.spec.js
```diff
@@ -5,35 +5,31 @@ import Paginate from './Paginate'
...
 describe('Paginate.vue', () => {
-  let propsData, wrapper, nextButton, backButton
-
-  beforeEach(() => {
-    propsData = {}
-  })
+  let propsData = {}
+  let wrapper
+  let nextButton
+  let backButton
```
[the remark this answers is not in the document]
alina-beck: You're right! I didn't think about it – just thought I could make it shorter by initializing the value. The extra lines were then added by `prettier`, I didn't even notice...

webapp/components/Paginate/Paginate.vue
alina-beck: I like the idea of hiding the buttons when there is neither `next` nor `previous`! And yes, I didn't want to take this PR too far. But I'm glad you decided to change the name in your commit. :wink:

webapp/components/generic/SearchableInput/SearchableInput.spec.js
alina-beck: Very good question. I tried but didn't get it to work. To be honest, there are so many nested elements from the styleguide that I didn't even know which one to select and `blur`. I thought the native `input` should do the trick – since the event should then bubble up through all the other elements – but it just broke everything... 🤷‍♀

webapp/components/generic/SearchableInput/SearchableInput.spec.js
alina-beck: Yeah, I was wondering about this and didn't really get what it was supposed to test. Let's keep refactoring! :grin:

webapp/layouts/default.vue
alina-beck: It's to override inline styles applied by the styleguide – and I thought it's harder to miss this way than with applying a class and `!important` styles. Needs to be refactored either way!

PR #2589, opened 2019-12-20
fix: swap lat and lng

### LINKED ISSUE #2587: 🐛 [Bug] Geolocation. lat and lng are swapped
When selecting a location in the user profile page, the latitude and the longitude are exchanged when saved in the database.

1. Go to your profile
2. Select Configuration
3. Type in some City or Region
4. Save
5. Check the lat and lng properties in the location node
grafik

### Expected behavior
The value of lat should be in lng and reverse.

Mogge:
### Issues
- fixes #2587

### Todo
- [x] Find a solution how to repair the wrong data in production. Best would be to have script that calls the data by id from the geomap api for each location and merges the received data correctly in the lng and lat fields.


mattwr18:
before deploying this, we need to do a database manipulation where we swap the existing locations, as @Mogge said in our pairing session.

This script works... 
```
MATCH (location:Location) with location.lng as longitude, location.lat as latitude, location 
SET location.lng = latitude, location.lat = longitude 
RETURN location;
```

Mogge:
```cypher
MATCH (l:Location) WHERE NOT(l.lat IS NULL)
WITH  l.lng AS longitude, l.lat AS latitude, l AS Location
SET Location.lat = longitude, Location.lng = latitude
```

mattwr18:
Why do we have `Location` with no `lat` property @Mogge ??

backend/src/db/migrations/20200206190233-swap_latitude_with_longitude.js
```diff
@@ -0,0 +1,55 @@
...
+        SET Location.lat = longitude, Location.lng = latitude
+    `)
+    await transaction.commit()
+    next()
+  } catch (error) {
+    // eslint-disable-next-line no-console
+    console.log(error)
+    await transaction.rollback()
+    // eslint-disable-next-line no-console
+    console.log('rolled back')
```
roschaefer:
```suggestion
    console.log('rolled back')
    throw new Error(error)
```

backend/src/db/migrations/20200206190233-swap_latitude_with_longitude.js
```diff
@@ -0,0 +1,55 @@
+import { getDriver } from '../../db/neo4j'
+
+export const description =
+  'This migration swaps the value stored in Location.lat with the value of Location.lng.'
+
+export async function up(next) {
```
roschaefer:
```suggestion
const swap = (next) => {
```

backend/src/db/migrations/20200206190233-swap_latitude_with_longitude.js
```diff
@@ -0,0 +1,55 @@
...
+    console.log(error)
+    await transaction.rollback()
+    // eslint-disable-next-line no-console
+    console.log('rolled back')
+  } finally {
+    session.close()
+  }
+}
+
+export async function down(next) {
```
roschaefer: Instead:
```js
export async function up(next) {
  swap(next)
}
export async function down(next) {
  swap(next)
}
```

backend/src/db/migrations/20200206190233-swap_latitude_with_longitude.js
Mogge: Done

backend/src/db/migrations/20200206190233-swap_latitude_with_longitude.js
Mogge: Done

backend/src/db/migrations/20200206190233-swap_latitude_with_longitude.js
Mogge: Done

backend/src/db/migrations/20200206190233-swap_latitude_with_longitude.js
```diff
@@ -0,0 +1,42 @@
...
+values lat=10.0 and lng=53.55, which is close to the horn of Africa,
+but it is lat=53.55 and lng=10.0
+`
+
+const swap = async function(next) {
+  const driver = getDriver()
+  const session = driver.session()
+  const transaction = session.beginTransaction()
+  try {
+    // Implement your migration here.
```
mattwr18:
```suggestion
    
```
I would remove this line, but I won't request changes...

Issue #1922, opened 2019-10-16
🚀 [Feature] Compress uploaded images automatically


alina-beck:
## :rocket: Feature
Post authors should be able to upload images without too much hassle (so the file size should not be too heavily restricted).

Users should be able to view posts without having to download huge images (eating up bandwidth and taking a long time).

So the desired solution is to automatically compress all images that are uploaded to our network by our users. From my experience I would propose a maximum size of roughly 500-800KB – though this should be tested to make sure the images still have an acceptable quality.

### Additional context
This came up when we realized that images touched by `cropperjs` would have file sizes around 10MB! More thoughts here: #1933

mattwr18:
I have been playing around with this in https://github.com/Human-Connection/Human-Connection/pull/1935 and learned that the `canvas.toBlob()` that we use automatically uses `loosy compression`, which I had to look up what means.

Point, being, after I specified type `image/jpeg` as @vardyb suggested, and created a new File from the blob to restore filename, the cropped images saved are anywhere between around `175k-950k`, kinda in line with what you suggest, no @alina-beck 

Could you have a look at the `PR`?

alina-beck:
@mattwr18 I approved your PR! Great that you figured this out! :smile:

I'm wondering if we should still leave this issue open... File size obviously isn't a problem anymore as long as we force users to open the cropping tool, which we currently do but were thinking of changing to an opt-in feature soon. When that is implemented cropped images will have the desired file size but others not. So, what do you think?

mattwr18:
I think that makes sense

alina-beck:
@HelixX23 had an interesting comment about image validation and hidden content in images – we could consider this when working on the above issue: https://github.com/Human-Connection/Human-Connection/issues/2119#issuecomment-554490773

alina-beck:
still relevant – and probably related to #2255

HelixX23:
Maybe imagemagick could do the trick? 
https://www.imagemagick.org/script/index.php

mattwr18:
So, we have added this to the `EPIC` #2265 and therefore it's been put into the current Sprint, but we haven't had much time to plan it out, and share ideas... 
I opened a `PR` for an Image microservice, which leverages `sharp` 
https://github.com/lovell/sharp
https://github.com/Human-Connection/Human-Connection/pull/2602/files , but this seems to be more about resizing images, then compressing them.

[...]

PR #2602, opened 2019-12-20
[WIP] Image MicroService

mattwr18:
- Maybe this isn't helpful at all, but since no one else has opened up a
WIP PR, I though I would try to learn a bit before the holidays. I hope
we can use this as a starting point to start a discussion about this.

- Following tutorial https://malcoded.com/posts/nodejs-image-resize-express-sharp/


image-micro-service/src/server.js
```diff
@@ -0,0 +1,19 @@
+const express = require('express')
+const resize = require('./resize')
+const server = express()
+
+server.get('/', (req, res) => {
```
roschaefer: This looks great! Maybe we can merge this into the backend in order to quickly merge the responsive image feature?

mattwr18:
should I just close this then @Human-Connection/hc-dev-team ??

Tirokk:
We should discuss it in the team with all other possible solutions together and make a team decision at all. Or? @mattwr18 @roschaefer

image-micro-service/src/server.js
mattwr18: how would this work with what we have in the backend? 
I'm looking at this as a potential solution for https://github.com/Human-Connection/Human-Connection/issues/1922
but I don't think we would use `server.get('/')`
I couldn't easily figure out what it should be... I think maybe `server.get('/api/')`, oder? @roschaefer

mattwr18:
we are about to be in a position to re-evaluate this `PR` and see if we want to go in this direction(resize on the fly), or load multiple images to `spaces`

mattwr18:
closing in favor of [imgproxy](https://github.com/imgproxy/imgproxy)

PR #2629, opened 2019-12-29
chore: 🍰 [WIP] Deploy Backend And Neo4j Resources Requests And Limits

### LINKED ISSUE #2628: 🐛 [Bug] Backend Exceeds Heap Size Limit and Restarts
## :bug: Bugreport

Before Christmas the backend restarted every 5 minutes.

Matt and I have found that the heap size of the backend by default should be 1.5Gi.

It seems that we have to define:

```yaml
    spec:
      containers:
        resources:
          requests:
            memory: XXX
            cpu: XXX
          limits:
            memory: XXX
            cpu: XXX
```

And that should be done for the Neo4j database as well where only `requests. memory` was already defined.

See:
- https://kubernetes.io/docs/concepts/configuration/manage-compute-resources-container/
- https://cloud.google.com/blog/products/gcp/kubernetes-best-practices-resource-requests-and-limits

Restarted 24 time since 21.12.2019 …
But does not since 25.12.2019 …

### Expected behavior

Should not restart by error.

### Additional context

Suggestions in VSCode are:

Bildschirmfoto 2019-12-23 um 10 25 52

Bildschirmfoto 2019-12-23 um 10 25 28

May we have to set `requests` and `limits` for web app as well!?

Tirokk:
## 🍰 Pullrequest

Hopefully fixes the heap size error of the backend and went into better organisation of Kubernetes resources.

- [ ] May we have to set requests and limits for the webapps as well!?

This all has to be discussed.

### Issues

- fixes #2628

### Todo

- [ ] ? (see above)
- [ ] we have merged a potential fix into master and will deploy to production on 13/01/2020, then we can revisit this, agree on sensible limits, and test them before merging(?)

### DIFF
--- a/backend/yarn.lock
+++ b/backend/yarn.lock
@@ -1035,7 +1035,7 @@
     url-regex "~4.1.1"
     video-extensions "~1.1.0"
 
-"@metascraper/helpers@^5.8.10", "@metascraper/helpers@^5.8.12":
+"@metascraper/helpers@^5.8.12":
   version "5.8.12"
   resolved "https://registry.yarnpkg.com/@metascraper/helpers/-/helpers-5.8.12.tgz#c4c1375a90ee9b674f8fb4d5a65cce6f5c6ce30d"
   integrity sha512-hmaIRXWcLGFWAXFKBHECHhf3VhHrbz/iV6spPtTeYyxCVO1TX62qYigqbizZwHk4dGeU1cTtbT2YN8/RCr1RiQ==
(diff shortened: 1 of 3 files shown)


mattwr18:
can this be closed? or can you link to the docs you found describing why this is good to set @Tirokk ??

mattwr18:
should we discuss this, close it @Tirokk

Tirokk:
The reason is to discuss is in the related and fixed issue #2629 (was already closed and now reopened).

See:
- https://kubernetes.io/docs/concepts/configuration/manage-compute-resources-container/
- https://cloud.google.com/blog/products/gcp/kubernetes-best-practices-resource-requests-and-limits

Tirokk:
This was successful at the end.
So I close this one …

But there is still a problem see #3648 .

PR #2677, opened 2020-01-06
[WIP] Migrate Avatar component to Monorepo

### LINKED ISSUE #2675: 🚀 [Refactor] Migrate Avatar component to code base
## :zap: Refactor

As part of our migration, we want to move the `Avatar` component from the styleguide repo to our monorepo
Also, we need to add `srcset` as part of our `Responsive` images `Issue` https://github.com/Human-Connection/Human-Connection/issues/2255
### User Problem

See responsive images issue
### Implementation

migrate

mattwr18:
- I did not refactor any of the styleguide's code yet... just wanted to
get it working without the styleguide. Let's refactor?
### Issues

- fixes #2675

### DIFF
--- a/webapp/components/Avatar/Avatar.spec.js
+++ b/webapp/components/Avatar/Avatar.spec.js
@@ -31,9 +31,7 @@ describe('Avatar.vue', () => {
     describe('with a relative avatar url', () => {
       beforeEach(() => {
         propsData = {
-          user: {
-            avatar: '/avatar.jpg',
-          },
+          image: '/avatar.jpg',
         }
       })
 
@@ -49,9 +47,7 @@ describe('Avatar.vue', () => {
     describe('with an absolute avatar url', () => {
       beforeEach(() => {
         propsData = {
-          user: {
-            avatar: 'https://s3.amazonaws.com/uifaces/faces/twitter/sawalazar/128.jpg',
-          },
+          image: 'https://s3.amazonaws.com/uifaces/faces/twitter/sawalazar/128.jpg',
         }
       })
 
--- a/webapp/components/Avatar/Avatar.spec.js
+++ b/webapp/components/Avatar/Avatar.spec.js
@@ -1,8 +1,10 @@
 import { mount } from '@vue/test-utils'
 import Avatar from './Avatar.vue'
+import Vue from 'vue'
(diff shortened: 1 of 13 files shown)


alina-beck:
Hey @mattwr18, sorry to spoil the fun but this PR doesn't seem very helpful to me... 😕

When I migrate things from the styleguide I try to look at how it's implemented, check which of the properties are actually used in our project and then transfer only the bits that we use and need. I think it is much harder to do it the other way around – to first copy everything and then get rid of the stuff we don't need.

But maybe that's just me! 🙂 If you want to go ahead in this direction here's a few things to think about:

My impression is that we could replace all the `token` mappings and helpers with actual design tokens and CSS classes, now that everything lives in the same repo – it would make the code much more readable and save us a few lines (more than 1.000 I think? :wink:).

We should also rename the `Avatar` component (see the [Vue Styleguide recommendations](https://vuejs.org/v2/style-guide/#Multi-word-component-names-essential)) and maybe think about splitting it into even smaller components (e.g. a `RoundedImage` could be helpful?).

I am also not happy about introducing `Vuetify` components at this point because we are essentially replacing one dependency with another. Of course `Vuetify` is much better maintained and powerful than our styleguide, so it's definitely an improvement. But as soon as we want to do something that `Vuetify` didn't account for we run into the exact same problems we have now, except that adding features to `Vuetify` might be even more of a hassle. I would prefer to build our own components, based on our own needs, as long as it isn't overly complicated (I think an `Image` component should be doable whereas building our own `Editor` would be pure madness :grin:).

But I know that other people on the team have different opinions on this (@roschaefer :wave:) and I guess using the `Vuetify` image component might be the quickest solution right now – so both approaches have their advantages and disadvantages. 🤷‍♀

mattwr18:
I have gone ahead and went with your suggestion, and opened a new `PR` where we can take it from a minimalist standpoint. See https://github.com/Human-Connection/Human-Connection/pull/2700
I agree there are advantages and disadvantages, I added it to see if we could get it working and then we could discuss them. I :green_heart: having control over the components so they are as customizable as possible, but I also understand when @roschaefer talks about how `Vuetify` has probably thought about a lot of issues that we have not even considered.

PR #2686, opened 2020-01-07
feat: Convert block/unblock to mute/unmute

### LINKED ISSUE #279: 🚀 [Feature] Blacklist Users
## :rocket: Feature

~~~ gherkin
As a network User 
I'd like to filter out content of other users 
in order to better concentrate on the content I really search for.
~~~
This is deciding, **what I'd like to see**.
This is different from blocking users: Deciding, what the other user can't do anymore.
This is part of the blacklist (filter users, categories and for terms).
This is agreed with Dennis on 2019-09-23.

**Details of blacklisting Users**
- [ ] I don't see his articles anymore
- [ ] I see his comments on articles blurred
- [ ] I don't see his contributions in my search results

- [ ] Optionally ask, if a user should also be blocked

### Design & Layout

blacklist

### Additional context

Later there should be multiple blacklists which can be shared. 

Also there will be a "family friendly"/ "Ponyhof" blacklist which is trimmed for new users so the don't get offended when entering the network

mattwr18:
- at the moment, we have implemented blocked like we want the
blacklist/whitelistUserContent to be, with the exception that is should
not be both ways. If I blacklist a user's content, they still see my
content in their news feed.

### Issues

- fixes #279


cypress/integration/user_profile/mute-users/Mute.feature
```diff
@@ -1,43 +1,53 @@
...
     And I search for "Spam"
     Then the search has no results
-    But I search for "not blocked"
+    But I search for "not muted"
     Then I should see the following posts in the select dropdown:
       | title                    |
       | Post that should be seen |
+
+  Scenario: Muted users can still see my posts
+    Given I previously created a post
```
roschaefer: Improve your narration by saying
```
Given sb. created a post
And muted me
```
because it's quite unlikely that you would for real set up two accounts yourself and let them mute each other.

I'm not going to request changes for this though.

package.json
```diff
@@ -14,7 +14,7 @@
     "cypress:backend": "cd backend && yarn run dev",
     "cypress:webapp": "cd webapp && yarn run dev",
     "cypress:setup": "run-p cypress:backend cypress:webapp",
-    "cypress:run": "cross-env cypress run --browser chromium",
+    "cypress:run": "cross-env cypress run",
```
roschaefer: Oh really? We can use electron again? We should remove the snap `chromium` from our build pipeline then.

webapp/locales/en.json
```diff
@@ -313,26 +313,26 @@
...
-      "empty": "So far, you have not blocked anybody.",
-      "how-to": "You can block other users on their profile page via the content menu.",
-      "block": "Block user",
-      "unblock": "Unblock user",
-      "unblocked": "{name} is unblocked again"
+      "empty": "So far, you have not muted anybody.",
+      "how-to": "You can mute other users on their profile page via the content menu.",
+      "mute": "Mute user",
+      "unmute": "Unmute user",
+      "unmuted": "{name} is unmuted again"
```
roschaefer: Are there translations in other languages for this?

mattwr18: the translations still need to be fixed
I changed the keys, but the actual translations still say block/unblock

PR #2687, opened 2020-01-07
Enable apoc imports

mattwr18:
## 🍰 Pullrequest

This is helpful, no @roschaefer ??


mattwr18:
closing... we can discuss whether we actually want this... I think for production, we want this configurable from `configMaps`, no @roschaefer ??
also, the commit history is all messed up

PR #2688, opened 2020-01-07
🍰 2138 Slug Of A User Must Be Unique

### LINKED ISSUE #2138: 🐛 [Bug] user slug
## :bug: Bugreport
the user slug should either be restricted or best locked for modification. 

FireShot Capture 168 - Test Benutzernamenlänge - Human Connection - human-connection social

 

### Expected behavior
give the slug length a maximum length OR

if we use the slug as a link it should be unique and remain so. it then replaces the userid to reach a user or organization with link redirection. 

otherwise we have to save the old slugs to a user when we change them. to be able to assign a user to a changed slug. 

i would suggest to leave it unambiguous. 
composed of first+surname(++int) with double name assignment int upcount. 

a change of the plug must then be requested. in order to be able to adapt this in the system. which means however further support.

ogerly:
## 🍰 Pullrequest
A user can no longer change a slug in his settings. 

### Issues
- fixes #2138 
 
FireShot Capture 278 - Human Connection - Human Connection - localhost

### TODO 
- Userslug Request change - Moderation level feature : https://github.com/Human-Connection/Human-Connection/issues/2699
- Select and check userslug at "New Account" form: https://github.com/Human-Connection/Human-Connection/issues/2701

### DIFF
--- a/webapp/locales/de.json
+++ b/webapp/locales/de.json
@@ -42,11 +42,11 @@
     "settings": {
         "name": "Einstellungen",
         "data": {
-            "name": "Deine Daten",
-            "labelName": "Dein Name",
-            "labelCity": "Deine Stadt oder Region",
-            "labelBio": "Über Dich",
-            "labelSlug": "Dein eindeutiger Benutzername",
+            "name": "Meine Daten",
+            "labelName": "Mein Name",
+            "labelCity": "Meine Stadt oder Region",
+            "labelBio": "Über Mich",
+            "labelSlug": "Mein eindeutiger Benutzername",
             "namePlaceholder": "Petra Lustig",
             "success": "Deine Daten wurden erfolgreich aktualisiert!"
         },
--- a/webapp/locales/en.json
+++ b/webapp/locales/en.json
@@ -204,11 +204,11 @@
   "settings": {
     "name": "Settings",
     "data": {
-      "name": "Your data",
-      "labelName": "Your Name",
-      "labelSlug": "Your unique user name",
+      "name": "My data",
+      "labelName": "My Name",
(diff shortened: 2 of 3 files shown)


alina-beck:
I think a user _should_ be able to change their slug! But it should be restricted
- to a maximum length (maybe somewhere between 20 and 30 characters?)
- to a minimum length (maybe 4 characters?)
- to be unique (@roschaefer suggested to treat `_` and `-` as one character when comparing slugs to avoid too close similarities)

ogerly:
> Ich denke, ein Benutzer _sollte_ in der Lage sein, seinen Slug zu ändern! Aber es sollte eingeschränkt werden
> 
> * auf eine maximale Länge (vielleicht zwischen 20 und 30 Zeichen?)
> * auf eine Mindestlänge (vielleicht 4 Zeichen?)
> * um einzigartig zu sein ( @roschaefer schlug vor, `_`und `-`beim Vergleich von Schnecken als ein Zeichen zu behandeln , um zu enge Ähnlichkeiten zu vermeiden)

NO
i think it is too high a risk that a user can simply change his slug. this should only be done on request.  
The slug is a unique human readable ID. 
- The slug may only be given once. 
- if you change a slug you have to save it.

a readable URL has no userid, so we can only use one slug. 

i create an issues to do this extra. that a user can request a change of the slugs. maximum every 6 months. 

the username can be changed at any time. 

Translated with www.DeepL.com/Translator (free version)

alina-beck:
Ok, I understand. 🙂 Then I would suggest to:
- add these restrictions to the registration process when a user initially chooses their slug
- communicate clearly during signup that the slug cannot be changed easily

PR #2700, opened 2020-01-08
refactor(styleguide): Migrate Avatar component to monorepo

### LINKED ISSUE #2675: 🚀 [Refactor] Migrate Avatar component to code base

mattwr18:
- follow @alina-beck PR review
- KISS
### Issues

- fixes #2675 

## ToDos
- [ ] Add `RoundedImage` component that supports image responsiveness... as a minimum `srcset`


webapp/components/_new/generic/UserAvatar/UserAvatar.vue
```diff
@@ -0,0 +1,106 @@
...
+    width: $size-avatar-x-large;
+    height: $size-avatar-x-large;
+  }
+  .no-image {
+    height: 100%;
+    display: flex;
+    flex-wrap: wrap;
+    border-radius: 50%;
+    background-color: $background-color-secondary;
+    color: $text-color-primary-inverse;
```
mattwr18: I don't think we want to obviously leave it like this, but I'm not sure how we should do it, if we want to avoid the helpers backgroundColorsArray @alina-beck

mattwr18:
hey @alina-beck... is this a better start?
I will leave it for now so I can get some feedback before continuing 
What I like about this is I have tried to minimize it to what we are actually using, as you suggested.

I also like that I was able to refactor out all styleguide components from the newly renamed `UserAvatar` component, which means I was able to add it to our `_new/generic` directory.

In refactoring out `<ds-flex> and <ds-flex-item>` I'm sure there is better CSS than adding `margin: 0 auto, display: table`, but I guess I'll wait for your feedback.

webapp/components/_new/generic/UserAvatar/UserAvatar.vue
alina-beck: I still need to take a closer look at everything that's going on in the `helper` and `tokenMap` files. There's nothing speaking against simply adding the `avatar-tokens` to the main design token file and I don't really see why we would need different background colors? And when that's working there's not much more to add here. 🤷‍♀

webapp/components/_new/generic/UserAvatar/UserAvatar.vue
```diff
@@ -0,0 +1,106 @@
...
+  }
+
+  .no-image .flex-item {
+    box-sizing: border-box;
+    padding: 0;
+    margin: 0 auto;
+    align-self: center;
+    display: table;
```
alina-beck: I would avoid `display: table` unless it's for an actual table – with rows and columns and headings and stuff. Again, I don't have the right solution up my sleeve! I'd just open the browser console and play around a bit. :smile:

webapp/pages/profile/_id/_slug.vue
```diff
@@ -11,9 +11,9 @@
           style="position: relative; height: auto;"
         >
           <hc-upload v-if="myProfile" :user="user">
-            <hc-avatar :user="user" class="profile-avatar" size="x-large"></hc-avatar>
+            <user-avatar :user="user" class="profile-avatar" size="x-large"></user-avatar>
```
alina-beck: same as above with the `avatar` class!

webapp/components/_new/generic/UserAvatar/UserAvatar.vue
```diff
@@ -0,0 +1,106 @@
...
+
+<script>
+export default {
+  name: 'UserAvatar',
+  props: {
+    size: {
+      type: String,
+      default: 'base',
+      validator: value => {
+        return value.match(/(small|base|large|x-large)/)
```
alina-beck: I'd just keep the sizes we're actually using in the project! :wink:

alina-beck:
@mattwr18 I found a solution for the `text-overflow` – but it's not perfect because it's the container that's overflowing, so we can only have one styling for two types of text:
Screenshot 2020-01-21 at 11 22 36

Do you think it would be better to make the `...` small and grey instead of big and green? I guess it would be the more common use case.

The perfectionist in me is also saying "Ooh, I still don't like the separate template for `Anonymous` users" and "We could surely display the username below the slug dynamically when there's no `date-time`". But I'm trying to ignore her for now. :smile:

PR #2714, opened 2020-01-09
feat: Blocked users cannot comment on posts

### LINKED ISSUE #1724: 🚀 [Feature] Block User 2.0
## :rocket: Feature

~~~ gherkin
As a network User 
I'd like to block one or more other users
to not get confrontated with them any longer.
~~~
This is deciding what **an other user can't do anymore with me**.
This is different from blacklisting users: **what I'd like to see**.
This is agreed with Dennis on 2019-09-23.

**Changes to current implementation**
- [ ] The blocked user can always see my articles, these are not hidden
- [ ] Blocked users always get notices if they are mentioned by me (no shadow banning)

Shadow banning: (the other user does not know, that he is blacklisted)

**Behaviour as current implementation**
- [ ] But the blocked user can't comment under my articles
- [ ] Blocked users can't @mention me

- [ ] Optionally ask, if a user should also be blacklisted #279 

### Design & Layout

block-user

### Additional context

- [ ] Later: Blocked users can't have friendship with me
- [ ] Later: Blocked users can't contact me via chat

mattwr18:
### Issues

- fixes #1724 

### Todo

- [X] None


webapp/locales/en.json
```diff
@@ -317,7 +317,9 @@
         "their-perspective": "Vice versa: The blocked person will also no longer see your posts in their news feed.",
         "search": "Posts of blocked people disappear from your search results.",
         "notifications": "Blocked users will no longer receive notifications if they are mentioned in your posts.",
-        "closing": "This should be sufficient for now so that blocked users can no longer bother you."
+        "closing": "This should be sufficient for now so that blocked users can no longer bother you.",
+        "commenting-disabled": "Commenting is not possible at this time on this post.",
+        "commenting-explanation": "This can happen for several reasons, please see "
```
roschaefer: That's the thing. It can happen only for one reason at the moment: You got blocked. If this is indeed the case, I would be transparent about this. Exposing this information is exactly what we discussed with @DennisHack and it didn't seem to a problem.

backend/src/schema/resolvers/users.js
```diff
@@ -48,6 +48,29 @@ export default {
         throw new UserInputError(e.message)
       }
     },
+    blockedByPostAuthor: async (_parent, params, context, _resolveInfo) => {
```
roschaefer: Why is it necessary that Alice can ask Bob if he blocked Mallory? Only Bob should be allowed to ask Mallory if she blocked him.

webapp/pages/post/_id/_slug/index.vue
```diff
@@ -89,7 +89,19 @@
           @toggleNewCommentForm="toggleNewCommentForm"
         />
         <ds-space margin-bottom="large" />
-        <hc-comment-form v-if="showNewCommentForm" :post="post" @createComment="createComment" />
+        <comment-form
+          v-if="showNewCommentForm && !this.blocked"
```
roschaefer:
```suggestion
          v-if="showNewCommentForm && !post.author.isBlocked"
```

I think that's enough and you can remove `blockedByPostAuthor`.

webapp/pages/post/_id/_slug/index.vue
```diff
@@ -215,6 +223,26 @@ export default {
           id: this.$route.params.id,
         }
       },
+      update({ Post }) {
```
roschaefer: I like this refactoring :+1:

backend/src/schema/types/type/User.gql
```diff
@@ -68,10 +68,11 @@ type User {
   RETURN COUNT(u) >= 1
   """
   )
+  
   isBlocked: Boolean! @cypher(
   statement: """
-  MATCH (this)<-[: BLOCKED]-(u: User { id: $cypherParams.currentUserId})
-  RETURN COUNT(u) >= 1
+  MATCH (this)<-[:BLOCKED]-(user:User {id: $cypherParams.currentUserId})
```
roschaefer:
```suggestion
  MATCH (this)-[:BLOCKED]-(user:User {id: $cypherParams.currentUserId})
```
I would argue `BLOCKED` relationship is vice versa and then we can refactor the code below :point_down:

backend/src/schema/resolvers/users.js
mattwr18: We need to know if the `currentUser` has been blocked by the post author

backend/src/schema/types/type/User.gql
mattwr18: I was working under the understanding that blocking someone means only that a user can't interact with my content... Therefore not reciprocal

webapp/locales/en.json
mattwr18: We have talked about following some guidelines and maybe using Mastodon as a starting point.
This post explains that blocked users are not notified in any way they've been blocked. https://amp.reddit.com/r/Mastodon/comments/a1d5sx/mute_vs_block/

Also, since we have issues open to be able to turn off commenting on Posts, it makes sense to me to either decide we do or don't want to support them in the future. Then, we can either close those issues, or leave a generic message like it is in this PR. Then, we aren't divulging that comments are turned off because the user has been blocked by the other.

backend/src/schema/resolvers/users.js
mattwr18: I don't understand... The mutation checks if the `currentUser` represented by `context.user.id` has been blocked by the post author, oder?

backend/src/schema/resolvers/searches.js
```diff
@@ -40,8 +39,7 @@ export default {
       YIELD node as resource, score
       MATCH (resource)
       WHERE score >= 0.5
-      AND NOT (resource.deleted = true OR resource.disabled = true
-      OR (:User { id: $thisUserId })-[:BLOCKED]-(resource))
+      AND NOT (resource.deleted = true OR resource.disabled = true)
```
roschaefer: I think this introduced a bug: Your search results would contain posts by muted users.

backend/src/schema/resolvers/users.js
```diff
@@ -23,6 +23,21 @@ export const getMutedUsers = async context => {
...
+  const userModel = neode.model('User')
+  let blockedUsers = neode
+    .query()
+    .match('user', userModel)
+    .where('user.id', context.user.id)
+    .relationship(userModel.relationships().get('blocked'))
+    .to('blocked', userModel)
+    .return('blocked')
+  blockedUsers = await blockedUsers.execute()
+  blockedUsers = blockedUsers.records.map(r => r.get('blocked').properties)
```
roschaefer: Note that this will be imperformant on nested objects as we're not using `neo4j-graphql-js` to construct the query.

cypress/integration/common/steps.js
```diff
@@ -550,3 +570,11 @@ Then("I see only one post with the title {string}", title => {
     .should("have.length", 1);
   cy.get(".main-container").contains(".post-link", title);
 });
+
+Then("they should not see the comment from", () => {
+  cy.get(".ds-card-footer").children().should('not.have.class', 'comment-form')
+})
+
+Then("they should see a text explaining commenting is not possible", () => {
+  cy.get('.ds-placeholder').should('contain', "Commenting is not possible at this time on this post.")
```
roschaefer: That's quite important to avoid false negatives :+1: I told my students yesterday that they should look for something existing in the test before looking for something that does not exist, like in the step definition above :point_up:

cypress/integration/user_profile/BlockUser.feature
```diff
@@ -0,0 +1,48 @@
...
+    And I previously created a post
+    And the blocked user visits my post
+    Then they should not see the comment from
+    And they should see a text explaining commenting is not possible
+
+  Scenario: Block a previously followed user
+    Given I follow the user "Harassing User"
+    When I visit the profile page of the annoying user
+    And I click on "Block user" from the content menu in the user info box
+    And nobody is following the user profile anymore
```
roschaefer:
```suggestion
    Then I get removed from his follower collection
```

cypress/integration/user_profile/BlockUser.feature
```diff
@@ -0,0 +1,48 @@
...
+    When I click on "Block user" from the content menu in the user info box
+    And I navigate to my "Blocked users" settings page
+    Then I can see the following table:
+      | Avatar | Name           |
+      |        | Harassing User |
+
+  Scenario: Blocked user cannot interact with my contributions
+    Given I block the user "Harassing User"
+    And I previously created a post
+    And the blocked user visits my post
```
roschaefer: Sum up this and the two step definitions before:
```
  Given a blocked user visits the post page of one of my authored posts
```

cypress/integration/user_profile/BlockUser.feature
```diff
@@ -0,0 +1,48 @@
...
+    When I search for "see"
+    Then I should see the following posts in the select dropdown:
+      | title                      |
+      | You can still see my posts |
+
+  Scenario: Blocked users can still see my posts
+    Given I previously created a post
+    And I block the user "Harassing User"
+    Given I log out
+    And I am logged in as the "blocked" user
```
roschaefer: This sounds strange as those two people wouldn't be the same person in reality.

Better write something like:
```
  When the blocked user searches for "previously created"
```
And this logs out the current user, logs in the blocked user and types something into the search bar.

webapp/locales/de.json
```diff
@@ -143,27 +148,43 @@
...
+          "mute": "Stumm schalten",
+          "unmute": "Stummschaltung aufheben",
+          "unmuted": "{name} ist nicht mehr stummgeschaltet"
+        },
+        "blocked-users": {
+          "name": "Blocked users",
+          "explanation": {
+            "intro": "Wenn ein anderer Benutzer von dir blockiert wurde, dann passiert folgendes:",
+            "your-perspective": "In deiner Beitragsübersicht tauchen keine Beiträge der blockierten Person mehr auf.",
+            "their-perspective": "Umgekehrt das gleiche: Die blockierte Person sieht deine Beiträge auch nicht mehr in ihrer Übersicht.",
```
roschaefer:
```suggestion
            "their-perspective": "Die blockierte Person kann deine Beiträge nicht mehr kommentieren",
```
Translates to "The blocked person cannot comment on your posts anymore".

webapp/locales/de.json
```diff
@@ -143,27 +148,43 @@
...
+            "search": "Die Beiträge von blockierten Personen verschwinden aus deinen Suchergebnissen.",
```
roschaefer:
```suggestion
```

webapp/locales/de.json
```diff
@@ -143,27 +148,43 @@
...
+            "notifications": "Von dir blockierte Personen erhalten keine Benachrichtigungen mehr, wenn sie in deinen Beiträgen erwähnt werden.",
```
roschaefer:
```suggestion
```

webapp/locales/de.json
```diff
@@ -143,27 +148,43 @@
...
+          },
+          "columns": {
+            "name": "Name",
+            "slug": "Alias",
+            "unblock": "Entsperren"
+          },
+          "empty": "Bislang hast du niemanden blockiert.",
+          "how-to": "Du kannst andere Benutzer auf deren Profilseite über das Inhaltsmenü blockieren.",
+          "block": "Nutzer blockieren",
+          "unblock": "Nutzer entblocken",
```
roschaefer:
```suggestion
          "unblock": "Nutzer entsperren",
```

webapp/locales/de.json
```diff
@@ -143,27 +148,43 @@
...
+          "how-to": "Du kannst andere Benutzer auf deren Profilseite über das Inhaltsmenü stummschalten.",
+          "mute": "Stumm schalten",
+          "unmute": "Stummschaltung aufheben",
+          "unmuted": "{name} ist nicht mehr stummgeschaltet"
+        },
+        "blocked-users": {
+          "name": "Blocked users",
+          "explanation": {
+            "intro": "Wenn ein anderer Benutzer von dir blockiert wurde, dann passiert folgendes:",
+            "your-perspective": "In deiner Beitragsübersicht tauchen keine Beiträge der blockierten Person mehr auf.",
```
roschaefer:
```suggestion
            "your-perspective": "Du kannst keine Beiträge der blockierten Person mehr kommentieren.",
```
Translates to: "You cannot comment on posts of the blocked person"

webapp/locales/en.json
```diff
@@ -329,6 +329,28 @@
...
       "unmute": "Unmute user",
       "unmuted": "{name} is unmuted again"
+    },
+    "blocked-users": {
+      "name": "Blocked users",
+      "explanation": {
+        "intro": "If another user has been blocked by you, this is what happens:",
+        "your-perspective": "You will no longer be able to interact with their contributions.",
+        "their-perspective": "Vice versa: The blocked person will also no longer be able to interact with your contributions.",
+        "notifications": "Blocked users will no longer receive notifications if they mention each other.",
```
roschaefer: Is that actually true? I suggested to remove it from the German translations, see above :point_up:

webapp/locales/en.json
```diff
@@ -329,6 +329,28 @@
...
+        "closing": "This should be sufficient for now so that blocked users can no longer bother you.",
+        "commenting-disabled": "Commenting is not possible at this time on this post.",
```
roschaefer: That translates to:
"Du kannst den Beitrag derzeit nicht kommentieren."

webapp/locales/en.json
```diff
@@ -329,6 +329,28 @@
...
+        "commenting-explanation": "This can happen for several reasons, please see our "
```
roschaefer: Translates to: "Dafür kann es mehrere Gründe geben, bitte schau in unsere "

backend/src/schema/resolvers/searches.js
```diff
@@ -20,7 +20,6 @@ export default {
       AND NOT (
         author.deleted = true OR author.disabled = true
         OR resource.deleted = true OR resource.disabled = true
-        OR (:User { id: $thisUserId })-[:BLOCKED]-(author)
```
mattwr18: as discussed in our pairing session @roschaefer ... this was unintended after merging in changes from `master`
I found out why it was still passing the cypress tests and will push up the code...
turns out we had some issues with copy :spaghetti: in the `Search User` PR where we didn't properly rename variables

webapp/components/ContentMenu/ContentMenu.vue
```diff
@@ -172,6 +172,23 @@ export default {
               icon: 'user-times',
             })
           }
+          if (this.resource.isBlocked) {
+            routes.push({
+              label: this.$t(`settings.blocked-users.unblock`),
+              callback: () => {
+                this.$emit('unblock', this.resource)
+              },
+              icon: 'user-plus',
```
mattwr18: questionable if we want to use the same icons as the `mute/unmute` or if we should have similar, but slightly different ones

webapp/pages/profile/_id/_slug.vue
```diff
@@ -67,7 +69,7 @@
           <ds-space margin="small">
             <template v-if="!myProfile">
               <hc-follow-button
-                v-if="!user.isMuted"
+                v-if="!user.isMuted || !user.isBlocked"
```
roschaefer: This is equivalent to:
```js
!(user.isMuted && user.isBlocked)
```

Are you sure? This whould show the button e.g. if you muted the user but did not block the user. :confused: 

Either:
```js
!(user.isMuted || user.isBlocked)
```

Or just leave it as it was:
```js
!user.isMuted
```
in case you want to allow that you can block **and** follow a user at the same time.

webapp/pages/profile/_id/_slug.vue
```diff
@@ -417,6 +420,24 @@ export default {
         this.$apollo.queries.profilePagePosts.refetch()
       }
     },
+    async blockUser(user) {
+      try {
+        await this.$apollo.mutate({ mutation: blockUser(), variables: { id: user.id } })
```
roschaefer:
```suggestion
        await this.$apollo.mutate({ mutation: blockUser(), variables: user })
```
Would be OK, too, I guess. I'm OK to only pass the `id` call though. Choose whatever you like more. Maybe the explicit `id` is more, well, explicit.

webapp/pages/profile/_id/_slug.vue
mattwr18: this is questionable what we want to do here cause it was the case that when a user `BLOCKED` another, that it would remove the `Follow` button and replace it with a `Unblock user` button
what should happen now? If a user has `MUTED` another user, it should appear `Unmute user`... and if they have `BLOCKED` as well?
should there be two buttons? I guess that might be noisy...

webapp/pages/settings/blocked-users.vue
```diff
@@ -0,0 +1,118 @@
...
+          <ds-list-item>
+            {{ $t('settings.blocked-users.explanation.notifications') }}
+          </ds-list-item>
+        </ds-list>
+      </ds-card>
+    </ds-space>
+    <ds-card v-if="blockedUsers && blockedUsers.length">
+      <ds-table :data="blockedUsers" :fields="fields" condensed>
+        <template #avatar="scope">
+          <nuxt-link
```
roschaefer: Another moment to hate our styleguide. It is not possible to create a `blocked-user-table-row` component from it :angry:

backend/src/schema/resolvers/searches.js
mattwr18: this is part of the `UserQuery`, so it would so users who you have blocked... this was specifically asked for.
I have updated the `PostQuery` to remove posts by those you have `Muted`... and found out why it was till passing cypress...

backend/src/schema/resolvers/users.js
mattwr18: it's copied from what we have in production, but noted...

cypress/integration/user_profile/BlockUser.feature
mattwr18: hahaha... it was you who wrote this, but I'll update it

webapp/pages/profile/_id/_slug.vue
mattwr18: see my comment :point_up_2:

webapp/pages/settings/blocked-users.vue
mattwr18: we'll refactor it out... `ReportRow` is a pretty good example, I think...

webapp/locales/de.json
mattwr18: you don't receive notifications if there exists a `BLOCKED` relationship between two users..
this is following `mastodon`'s specifications of blocking

webapp/locales/en.json
mattwr18: yes, see my response :point_up_2:

webapp/pages/profile/_id/_slug.vue
mattwr18: I copied what you added when you introduced the block feature... and I guess it's good enough ;)

webapp/pages/profile/_id/_slug.vue
```diff
@@ -68,16 +68,23 @@
...
+              <base-button
+                v-if="user.isBlocked"
+                @click="unblockUser(user)"
+                class="unblock-user-button"
+              >
+                {{ $t('settings.blocked-users.unblock') }}
+              </base-button>
+              <base-button v-if="user.isMuted" @click="unmuteUser(user)" class="unmute-user-button">
+                {{ $t('settings.muted-users.unmute') }}
+              </base-button>
```
roschaefer: It's possible that you could see both unblock and unmute buttons, right? I wonder how that would look like :thinking:

mattwr18: I uploaded a screen shot, but somehow I added it as a review, and cancelled it, without knowing it would remove it.

Screenshot from 2020-01-23 19-23-50

webapp/pages/post/_id/_slug/index.vue
```diff
@@ -90,7 +90,17 @@
           @toggleNewCommentForm="toggleNewCommentForm"
         />
         <ds-space margin-bottom="large" />
-        <hc-comment-form v-if="showNewCommentForm" :post="post" @createComment="createComment" />
+        <comment-form
+          v-if="showNewCommentForm && !post.author.isBlocked"
```
alina-beck: I'm just looking at the frontend right now, so not sure where this is coming from – but this reads like "when I have blocked the post author" when it should be "when the post author has blocked me". Or am I wrong?

webapp/pages/post/_id/_slug/index.vue
```diff
@@ -90,7 +90,17 @@
...
+          :post="post"
+          @createComment="createComment"
+        />
+        <ds-placeholder v-else>
+          {{ $t('settings.blocked-users.explanation.commenting-disabled') }}
+          <br />
+          {{ $t('settings.blocked-users.explanation.commenting-explanation') }}
+          <a href="https://support.human-connection.org/kb/">FAQ</a>
+        </ds-placeholder>
```
alina-beck: @mattwr18, you've been asking about the placeholder a few times – of course I'm not extremely happy about it since we're trying to get rid of styleguide components when we can. But in this case I think a) it looks fine, b) I don't know a better (and reusable!) solution off the top of my head and c) we'll have to migrate the `ds-placeholder` anyway, so we can leave it for future-us. :grin:

webapp/pages/post/_id/_slug/index.vue
```diff
@@ -122,7 +132,7 @@ export default {
     UserTeaser,
     HcShoutButton,
     ContentMenu,
-    HcCommentForm,
+    CommentForm,
```
alina-beck: :+1: I like this!

webapp/pages/profile/_id/_slug.vue
```diff
@@ -66,16 +68,23 @@
           </ds-flex>
           <ds-space margin="small">
             <template v-if="!myProfile">
+              <base-button
+                v-if="user.isBlocked"
+                @click="unblockUser(user)"
+                class="unblock-user-button"
```
alina-beck: It's probably hard to target the `.base-button` class directly since we don't have a direct (or close) parent with a unique class name? I'd prefer something like `.profile > .base-button` but that would be a bigger refactoring...

webapp/pages/profile/_id/_slug.vue
```diff
@@ -66,16 +68,23 @@
...
+              >
+                {{ $t('settings.blocked-users.unblock') }}
+              </base-button>
+              <base-button v-if="user.isMuted" @click="unmuteUser(user)" class="unmute-user-button">
```
alina-beck: same as ☝️

webapp/pages/profile/_id/_slug.vue
```diff
@@ -66,16 +68,23 @@
           </ds-flex>
           <ds-space margin="small">
```
alina-beck: I know this is not your code but here's an idea:
- put the `v-if` from the line below on the `ds-space`
- maybe make it a `div` (unless you come up with something more meaningful! :smile:)
- give it a class name and add the `margin` with CSS
- then you can target the buttons with `.this-class > .base-button` :tada:

webapp/pages/settings/blocked-users.vue
alina-beck: sooo many styleguide elements! :scream:

but I guess we'll get rid of them one by one within the next few months... :grin:

webapp/locales/en.json
alina-beck: I would be more direct – this sounds like commenting is disabled for everyone. I'd rather say "You cannot comment on this post at this time." and then have the second line as an explanation. (Also corresponds exactly with @roschaefer translation above. :wink:)

webapp/pages/post/_id/_slug/index.vue
mattwr18: it is reciprocal @alina-beck ... we could think about renaming it to make it clear... 
maybe `blockedExists` or `hasBlocked`
`post.authord.hasBlocked` works well, but `user.hasBlocked`, maybe not so much??

webapp/pages/post/_id/_slug/index.vue
mattwr18: ok, I looked at in the styleguide and it's quite simple, but we can look at it, and design it later

webapp/pages/post/_id/_slug/index.vue
mattwr18: yeah, I opened a ticket to rename all these in bulk, but I've been sneaking some of these in cause I can't help myself

webapp/pages/profile/_id/_slug.vue
mattwr18: I like this idea, I'll have a look at it tomorrow!

mattwr18:
hey @alina-beck not being specific is on purpose, following mastodon's specs that it's not open that you have been blocked.

webapp/pages/post/_id/_slug/index.vue
mattwr18: I just renamed it to `blocked`, which could stand for `isOrWasBlocked` ... didn't really love any of the names I could come up with to describe it..

webapp/pages/profile/_id/_slug.vue
mattwr18: I looked at alternatives to `div`, with the goal of reducing the amount we use them, but didn't find anything I thought fit, would love your feedback.
I looked at https://developer.mozilla.org/en-US/docs/Web/HTML/Element and tried to do some reading, for example https://dev.to/kenbellows/stop-using-so-many-divs-an-intro-to-semantic-html-3i9i 
I thought for a minute maybe `<section>`, but then read this 
```
The section element is not a generic container element. When an element is needed only for styling purposes or as a convenience for scripting, authors are encouraged to use the div element instead. A general rule is that the section element is appropriate only if the element's contents would be listed explicitly in the document's outline
```
and decided against it... 

I took a look as well at `<template>`, since the docs say it's `is a mechanism for holding HTML that is not to be rendered immediately when a page is loaded but may be instantiated subsequently during runtime using JavaScript.` , but it messed up the styling, meaning I would need to add more styling to get it to work. Feedback?

PR #2728, opened 2020-01-12
🍰 feat(webapp): Display deployed version in footer

### LINKED ISSUE #1831: 🚀 [Feature] Used Release Version Number
## :rocket: Feature
I would like to see the version/release number of the hc software.

### Design & Layout
e.g. in the footer :
HC Version 1.03

roschaefer:
## 🍰 Pullrequest

See #1831 this should help all our testers to see and compare the different deployed versions.

### Issues
- close #1831


alina-beck:
How about we replace `Changes` with the version number and link to the Github releases from there? :grin:

datenbrei:
@alina-beck  For me it sound like a good idea.

roschaefer:
@mattwr18 I didn't disable the auto-commit in `yarn run release` because it would also disable the auto-tagging. I think we should avoid human errors here, so I left it as it was.

PR #2777, opened 2020-01-14
feat: Set up Prometheus monitoring for Kubernetes

### LINKED ISSUE #2411: 🚀 [Feature] Collect performance analysis data
## :rocket: Feature

It's always difficult to say "improve performance on X" because we never know when this issue is finished. There is always something to improve on performance.

That's why I would like to set the goal of this issue to: "Collect performance analysis data" on our kubernetes cluster, because we just don't know how long a request takes time in our webapp, backend and database pods respectively.

### User Problem
We have non-deterministic performance problem on the network right now. It's difficult to know what causes the performance issues. Certainly it's even more difficult if we don't have any statistical data on performance.

### Implementation
Hard to tell.

### Validation
A goal could be to get sth. like a [flame graph](http://www.brendangregg.com/flamegraphs.html) on our kubernetes cluster.

mattwr18:
### Issues

- fixes #2411 
maybe?

### Todo

- [X] None

### DIFF
--- a/deployment/human-connection/monitoring/cluster-role.yaml
+++ b/deployment/human-connection/monitoring/cluster-role.yaml
@@ -1,0 +1,33 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: prometheus
+rules:
+- apiGroups: [""]
+  resources:
+  - nodes
+  - nodes/proxy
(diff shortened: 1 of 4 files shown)


mattwr18:
close in favor of https://github.com/Human-Connection/Human-Connection/pull/2825

PR #2824, opened 2020-01-17
feat: Translations Update

### LINKED ISSUE #2528: 🐛 [Bug] spelling mistake: "Kommentar Gesendet"
## :bug: Bugreport
If you publish your comment, you get the message "Kommentar Gesendet".

Correct is "Kommentar gesendet".

image

appinteractive:
This pull request was initiated by Lokalise (user Mogge) at 2020-01-17 13:36:47

 * fixes #2528


Mogge:
I checked the German file and it looks to me that the only changes are the spelling of 'Gesendet' to 'gesendet' and the obsolete key of 'reporter', which now is 'submitter'.
The Russian file should be checked by someone who controls the language.

webapp/locales/pt.json
```diff
@@ -321,7 +321,63 @@
             "disabledBy": "desativado por",
             "reasonCategory": "Categoria",
             "reasonDescription": "Descrição",
-            "createdAt": "Data"
+            "status": "",
```
roschaefer: @Mogge we must make sure that we don't add empty strings to our locales. Either leave it out or add `"status": null` to the translation files. Otherwise we disable the English fallback and the user will see a blank page.

roschaefer:
@Mogge can we also unlink the user @appinteractive from lokalise? It's the user account that keeps being added as PR author. I would prefer @human-connection-release-bot  or a new dedicated user account like `@human-connection-translation-bot`.

webapp/locales/pt.json
Mogge: I check this. I tried to reconfigure the @appinteractive , just to match the new semantic requirements but failed so far

Mogge:
@appinteractive It would also be nice if we could change the title of the generated pull requests.

appinteractive:
What do you need @mogge ? I don’t know if and how I can help you @roschaefer.

PR #2825, opened 2020-01-17
docs(deployment): Explain how to setup metrics

### LINKED ISSUE #2411: 🚀 [Feature] Collect performance analysis data

roschaefer:
## 🍰 Pullrequest
@mattwr18 I would suggest that we use helm and only update our docs. We have a dashboard on `staging` now. I'm OK to close the original issue now until we have additional demands for logging and metrics.

### Issues

   - close #2411
   - close #2777

### Todo

- [X] None


roschaefer:
Nice, I can log in to grafana, but you seem to have deployed another dashboard. Looks good anyways. Great teamwork :raised_hands:

PR #2826, opened 2020-01-17
fix: Error pages can be translated

### LINKED ISSUE #2409: 🐛 [Bug] User-not-found-page not translated
## :bug: Bugreport
The page "User not found" is not yet translated.

user

1. Call a user profile which had been deleted

### Expected behavior
The Page should
1. Be translated
2. Explain, why this page comes up. For example: "The user deleted his own account".

### Version & Environment
any

Mogge:
## 🍰 Pullrequest
grafik

### Issues
- fixes #2409


Mogge:
@alina-beck do you have suggestions for a better layout. May be a nice image ...

alina-beck:
yeah, an image would be nice – you could check the `/static/img/` folder to see if you can find one that fits?

Mogge:
Now it looks like this:
image

I took the `exclamation-circle` icon.
That do you think @alina-beck ?

Mogge:
One question. There are some 403 from webapp middleware. Should they be localised too?
There seems to be a generated 404 which so far I could not find in the Codebase. It appears when there is an URL which is not mapped at all. Where could I find this message?

webapp/layouts/error.vue
```diff
@@ -0,0 +1,34 @@
+<template>
+  <div class="error-container">
+    <base-icon class="error-icon" name="exclamation-circle" />
+    <br />
+    <span v-if="error.message === 'This page could not be found'" class="error-message">
```
Mogge: I was not able to locate the place where the default 404 error message comes from. So I catch it here. Any ideas?

roschaefer: You could dispatch the error code:

```js
computed: {
  message() {
    const mapping = {
       403: 'error-pages.not-authorized',
       404: // ...
       500: // ...
    }
    const key = mapping(this.error.code) || 'error.default'
    return this.$(key)
  }
}

File: webapp/middleware/isAdmin.js [hunk 2]
```diff
@@ -1,5 +1,5 @@
 export default ({ store, error }) => {
   if (!store.getters['auth/isAdmin']) {
-    return error({ statusCode: 403 })
+    return error({ statusCode: 403, message: 'error-pages.not-authorized' })
```
roschaefer: I'm having difficulties with `message` here. It's not the message, it's the translation key.

So, on the consuming side, I would check if there is a translation key for `error.message` (check if it's a key or if the error message is coming from NuxtJS). Just to make sure that we're not trying to translate a default Nuxt error message here.

webapp/layouts/error.vue
```diff
@@ -0,0 +1,50 @@
...
+      const mapping = {
+        403: 'error-pages.403-default',
+        404: 'error-pages.404-default',
+        500: 'error-pages.500-default',
+        503: 'error-pages.503-default',
+      }
+      return this.error.key || mapping[this.error.statusCode] || 'error-pages.default'
+    },
+    image() {
+      return `/img/svg/errors/error${this.error.statusCode ? this.error.statusCode : '500'}.svg`
```
roschaefer:
```suggestion
      return `/img/svg/errors/error${this.error.statusCode || '500'}.svg`
```

webapp/locales/de.json
```diff
@@ -798,5 +798,16 @@
         "donations-for": "Spenden für",
         "donate-now": "Jetzt spenden",
         "amount-of-total": "{amount} von {total} € erreicht"
+    },
+    "error-pages" : {
+	"profile-not-found": "Dieses Profil konnte nicht gefunden werden",
+	"back-to-index": "Zurück zur Startseite",
+	"post-not-found": "Dieser Beitrag konnte nicht gefunden werden",
+        "cannot-edit-post": "Dieser Beitrag kann nicht editiert werden",
```
roschaefer: Fix indentation

webapp/layouts/error.vue
Mogge: @roschaefer I thought you wanted an image :)

PR #2828, opened 2020-01-19
feat(db): Setup neo4j data migrations

### LINKED ISSUE #2381: 🐛 [Bug] Merge duplicate user accounts and email addresses
## :bug: Bugreport

So we just changed 940 *unnormalized* email addresses (those gmail email addresses which had dots) into their normalized version. We had to skip 40 email addresses where the users already registered again with their email addresses. That created another user account with the normalized email.

So in the long term should merge those unnormalized and normalized email addresses into one node and also merge the user accounts that are connected to these email addresses. That would allow the 40 users to get the former posts back.

In the short term we should maybe notify those 40 users and tell them about this issue?

roschaefer:
## 🍰 Pullrequest
I use #2381 as a use case to add consistent and reliable data migrations to our backend.

### Issues

- fix #2381 

### Todo

- [x] Configure `migrate` in docker image (`production` build stage)
- [x] Implement Neo4J data store for `migrate`
- [x] Refactor existing code for data tasks (seed, cleaning) into one folder `db/`
- [ ] Run migration on production
- [ ] Finish #1613 so we can run `yarn run prod:migrate init` on `post-install` and `yarn run prod:migrate up` on `post-upgrade`

I guess we have to do the latter manually (during a downtime)


backend/README.md
```diff
@@ -53,6 +53,27 @@ can issue GraphQL requests or access GraphQL Playground in the browser.
...
 
+### Database Indices and Constraints
+
+Database indices and constraints need to be created when the database and the
+backend is running:
+
+{% tabs %}
+{% tab title="Docker" %}
+```bash
+docker-compose exec backend yarn run db:migrate init
```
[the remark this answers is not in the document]
mattwr18: looking into this further @roschaefer ... this has nothing, it seems, to do with `Docker` vs. local setup, but production vs seed data.
It appears there is an error thrown when setting up the unique constraints on `Location` and then the rest of the 4 :point_up_2: don't get run. I ran them individually in the `neo4j` browser and got this error for `Location`, the other 4 were successful.

```
Unable to create CONSTRAINT ON ( location:Location ) ASSERT location.id IS UNIQUE:
Both Node(48981) and Node(186812) have the label `Location` and property `id` = 'place.4463223470618060'
```
Looking this `Location` up by it's id results in 4 identical `Locations` in production data, that look like
```
{
  "lng": 48.06028,
  "nameES": "Villingen-Schwenningen",
  "nameFR": "Villingen-Schwenningen",
  "nameIT": "Villingen-Schwenningen",
  "nameEN": "Villingen-Schwenningen",
  "type": "place",
  "namePT": "Villingen-Schwenningen",
  "nameDE": "Villingen-Schwenningen",
  "nameNL": "Villingen-Schwenningen",
  "name": "Villingen-Schwenningen",
  "namePL": "Villingen-Schwenningen",
  "id": "place.4463223470618060",
  "lat": 8.45861
}
```

I guess we don't really need to run `init` in production, right? should we delete these duplicate records anyways? 
Looking at our db schema in production confirms that we do no have any `CONSTRAINT` `on `Location`, so I guess we want to set that up?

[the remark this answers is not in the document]
roschaefer: @mattwr18 yes, we want to set that up. I used `neode` to setup unique constraints and if that fails on production data, then we had one constraint missing on `production`. Maybe we can run a migration before we setup the constraints. Sounds odd but is the best solution that comes to my mind right now.

backend/src/db/migrations/1579387929122-merge_duplicate_user_accounts.js
```diff
@@ -0,0 +1,84 @@
...
+    )
+    .subscribe({
+      next: ({ user, email, oldUser, oldEmail }) =>
+        // eslint-disable-next-line no-console
+        console.log(`
+          Merged:
+          =============================
+          userId: ${user.id}
+          email: ${oldEmail} => ${email.email}
+          =============================
```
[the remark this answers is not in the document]
roschaefer: Maybe my memory is wrong - or the migration. We should double check it.

backend/README.md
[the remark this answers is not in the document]
mattwr18: I've added a migration that runs first... but we would need to run `yarn db:migrate up` before running `yarn db:migrate init`... otherwise the latter will fail.
Also, I changed the timestamp of the `merge_duplicate_location_nodes` so that it run before the `merge_duplicate_user_nodes`, but actually that's not important.
We have talked about adding the migrations we have already run on production data to this folder so that we have a comprehensive list of every migration we have run... should those timestamps be altered as well to appear at the top of the list and be run first?
Also, when I added the unique constraints on `Migration` node, I realized it failed as it was creating multiple nodes for the same migration.
I've updated it to `MERGE` by title, then `ON CREATE SET` the properties and changed the timestamp to `createdAt` so that we have a record of when the migration was run in production, and it's human readable.

backend/src/db/migrations/1579387929122-merge_duplicate_user_accounts.js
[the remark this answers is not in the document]
mattwr18: I thought maybe the users could have deleted the old accounts, but I guess not, if they couldn't log in...

backend/src/db/migrate/store.js
```diff
@@ -35,7 +35,7 @@ class Store {
     const session = driver.session()
     const readTxResultPromise = session.readTransaction(async txc => {
       const result = await txc.run(
-        'MATCH (migration:Migration) RETURN migration {.*} ORDER BY migration.timestamp DESC',
+        'MATCH (migration:Migration) RETURN migration {.*} ORDER BY migration.createdAt DESC',
```
roschaefer: One could argue we should save two attributes: `createdAt` and `migratedAt`

backend/src/db/migrations/1579387929111-merge_duplicate_location_nodes.js
```diff
@@ -0,0 +1,77 @@
...
+              `
+                    MATCH (location:Location)
+                    RETURN location {.id}
+                  `,
+            )
+            .records()
+            .pipe(
+              map(record => {
+                const { id: locationIds } = record.get('location')
+                return { locationIds }
```
roschaefer: Why is this plural?

backend/src/db/migrations/1579387929111-merge_duplicate_location_nodes.js
```diff
@@ -0,0 +1,77 @@
...
+              }),
+              mergeMap(({ locationIds }) => {
+                return transaction
+                  .run(
+                    ` 
+                    MATCH(location:Location {id: $locationIds}), (location2:Location {id: $locationIds})
+                    WHERE location.id = location2.id AND id(location) < id(location2)
+                    CALL apoc.refactor.mergeNodes([location, location2], { properties: 'combine', mergeRels: true }) YIELD node as updatedLocation
```
roschaefer: This would create duplicate relationships between the merged node and the nodes which were connected to the previous nodes.

backend/src/models/Migration.js
```diff
@@ -1,5 +1,5 @@
 export default {
   title: { type: 'string', primary: true, token: true },
   description: { type: 'string' },
-  timestamp: { type: 'number', unique: true },
+  createdAt: { type: 'string', isoDate: true, default: () => new Date().toISOString() },
```
roschaefer: Yeah, so this is certainly `migratedAt`.

backend/src/db/migrate/store.js
mattwr18: maybe, but I guess those will be the same cause the `Migration` is only created in the database when it is run for the first time... after that, since we use merge, it is not updated.
I thought about calling it `migratedAt`, and wouldn't be opposed to it, but thought that `createdAt` could also be ok...

backend/src/db/migrations/1579387929111-merge_duplicate_location_nodes.js
mattwr18: good question... it actually should be, really

backend/src/db/migrations/1579387929111-merge_duplicate_location_nodes.js
mattwr18: would it?
so it should be false?

PR #2871, opened 2020-01-22
fix(webapp): 🐛 Adjust Avatar Sizes and 'z-index'

alina-beck:
## 🍰 Pullrequest
fixes some small issues @mattwr18, @Tirokk and I found with the new `UserAvatar` and `UserTeaser` that we had simply overlooked in the `migrate avatar PR` (#2700).


alina-beck:
@mattwr18 I found that using `$event` instead of just `event` in the vue template fixes the issue you pointed out on discord earlier – please check if it works for you, too! 🙂

mattwr18:
works for me too! thanks :)

Issue #2786, opened 2020-01-15
🚀 [Feature] Nano IDs For Human Readability


Tirokk:
## :rocket: Feature

Always if the UUIDs getting displayed to a user we need a much shorter nano (micro) ID for optimal human readability.

- Our UUIDs now like (36 chars long): 2812731d-6329-46ca-93b7-b9c7694246ec
- A nano-ID like (XXX): XXX

### User Problem

Readability and it’s just very long.

sushidave:
How about a short ID like [the one of Threema](https://threema.ch/en/faq/threema_id), the messenger app? It's 8 characters long and consists of uppercase letters and numbers only, e.g.: FN5YA4S. This looks user friendly to me.

roschaefer:
@sushidave problem is collision of IDs. We rely on the IDs never be the same. Nasty things happen in case of collisions.

sushidave:
@roschaefer Thanks. So the solution is to avoid identical IDs. Can be done by a check?

sushidave:
Oh, now I understand. Your comment was related to the "Nano ID" solution. If it is too risky then why not use short alphanumeric IDs (like Threema)?

Tirokk:
Is not stale!

Tirokk:
Is not stale …

Issue #2784, opened 2020-01-15
🚀 [Feature] On A Post Show User ID Button In User Hover Menu


Tirokk:
## :rocket: Feature

As a user I like to see the user profile ID of a user commented on a post or somewhere else.
It should show up if I hover over the user slug of a post or a comment and click on a button named `Show all user IDs`.

1. After the click all the **authors** IDs of **the post** and of **all comments** in the list on the post page shall appear. They shall be visible until I refresh the page.
2. The same shall happen if I hover over the post author slug and click in the same button on the menu which appears there. (is decided with @DennisHack 15.01.2020)

Suggestion:

<img width="1813" alt="Bildschirmfoto 2020-01-15 um 11 28 15" src="https://user-images.githubusercontent.com/25344101/72439888-108e0e00-37a8-11ea-9ccb-a949d8a5ff40.png">

### User Problem

If some users change their slugs or names I as a user like to have the ability to see a unique identifier to identify a user safely.

### Implementation

See above. ☝🏼

### Design & Layout

See above. ☝🏼

### Validation

None

### Additional context

None

Tirokk:
My dear @alina-beck !
How you would design this nicely?
Please contact me personally ;-)

alina-beck:
This is not so easy!

First off, I don't like the idea of showing all `user IDs` on a page by clicking a button hidden in the hover menu of _one_ user. And then hiding them again by reloading the page – this is not at all transparent and I would strongly advise against it from a UX perspective.

Also, why would I be able to see all `user IDs` on the `Post page` but nowhere else? Maybe I want to see the IDs of all post authors on the `index page` or check the IDs of users following another user, on the `profile page`. Why is that less important? And how do we explain to the users where this feature exists and where it doesn't? (It's also more difficult to implement – we'd have to check for the route to display or hide the button accordingly.)

Another way to implement this would be to have a `show user IDs` button somewhere globally available. I'm not sure where, though. In the `Avatar Menu`, maybe? Or in the `footer`? Either way, it should have the same effect on _all_ pages. It could also be a `setting`, similar to allowing `Embeds` or not.

And there's another issue: displaying the `user IDs` next to (or close to) the user slug will definitely mess up the design – on any page but the `Post page` where there's usually enough space. Except on mobile, of course.

One way to solve this would be to _replace_ the `username` or `date-time` below the slug with the `user ID` when the setting is active – when it's a temporary change I think this solution is preferable to a messed up layout.

[...]

Issue #209, opened 2019-02-25
Slugs - Permanent, unbreakable, readable and text changeable slugs for users, posts etc.


Tirokk:
**Is your feature request related to a problem? Please describe.**
1. An URL – a slug – should potentially last forever and do never break. *(We do program for eternity!)*
2. The individual and human readable part of the slug should be changeable. Because real names can change and headlines of posts as well. And since the human readable part of the slug is important for SEO it should be changeable, because you may have typed wrong or your realise that another headline of a post will do much better or even violates rules etc.

**Describe the solution you'd like**
A solution can be a combination of a unique ID for the user, post etc. followed by a human readable part, that can be changed by the user by changing their names or headlines etc.

As an example stackexchange.com uses them:
[https://webmasters.stackexchange.com/questions/113717/will-having-product-id-in-urls-hurt-seo-compared-to-a-url-with-only-keywords](https://webmasters.stackexchange.com/questions/113717/will-having-product-id-in-urls-hurt-seo-compared-to-a-url-with-only-keywords)

So for HC it will look like:
`https://human-connection/profile/85739/peter-lustig`
`https://human-connection/post/6976396/distinctio-amet-voluptate/`

*But there have to be some things considered closely !!!*
See later …

**Describe alternatives you've considered**
An alternative could be to put the ID at the very end of the slug.
`https://human-connection/profile/peter-lustig/85739`
But I read that this has the danger to be cut of by algorithms that **truncate the URL after 80** chars as an example. And we would have to be aware to limit the headlines of posts somehow.

**Additional context**
I have investigate this today more closely after my discussion about this with @roschaefer on Saturday.

If we do so …
1. … we have do redirect 301 every URL that hits the right ID but does not match the human readable text part. Otherwise the SEO will decrease, because of duplicate URLs to the same content and loose the SEO of the first version.
2. Every page has to have it’s unique URL as „rel=canonical“ meta tag. Here the readable text should be included, because otherwise this text will be excluded from the SEO by Google.

The ID may have an impact on the SEO as well. But with this the people are not really sure.
1. On one hand the ID part counts as an additional „word“ of the URL which may lower the wait of the words in the readable text part.
2. But on the other hand Google itself forces the news companies to put unique numbers at the articles URLs of their news feeds to avoid duplicated URLs to different content.

Tirokk:
We should have micro-UUIDs or short-UUIDs in our URL as standard.
The long versions should be possible for compatibility.

[...]

PR #2884, opened 2020-01-23
feat: 🍰 Display User ID On User Profile

### LINKED ISSUE #252: Show user ID in the profile, small and grey somewhere underneath
At the moment one has problems to identify a user safely.
If you have fake users or trouble with one, it is difficult to see if this is still the same profile.

Show user ID in the profile somewhere underneath, small and grey. Will say not really prominent. If you get trouble you can make a screenshot of the profile.

**Describe alternatives you've considered**
You might kind of see the change of a users name if you remember a comment or so somewhere which  seems to change it’s name. But this is really unsafe.

**Additional context**
A user ID showing up in profiles we use on PlanetRomeo and this is kind of effective.

Tirokk:
## 🍰 Pullrequest

Show user ID on the user profile somewhere underneath small and grey.

### Issues

- fixes #252

### Todo

- [x] implement testing
- [x] implement several other tests for the user profile page as well
- [ ] change the title of the page to contain the user name and slug

### Look actually like

Bildschirmfoto 2020-01-23 um 09 41 47

### DIFF
--- a/webapp/pages/profile/_id/_slug.spec.js
+++ b/webapp/pages/profile/_id/_slug.spec.js
@@ -21,12 +21,15 @@ describe('ProfileSlug', () => {
         id: 'p23',
         name: 'It is a post',
       },
-      $t: jest.fn(),
+      $t: jest.fn(a => a),
+      $filters: {
+        date: jest.fn(a => a),
+      },
       // If you're mocking router, then don't use VueRouter with localVue: https://vue-test-utils.vuejs.org/guides/using-with-vue-router.html
       $route: {
         params: {
-          id: '4711',
-          slug: 'john-doe',
+          id: 'f26bc193-605d-4465-994e-c2f328d9db01',
+          slug: 'bob-the-builder',
         },
       },
       $router: {
@@ -63,7 +66,7 @@ describe('ProfileSlug', () => {
           getters: {
             'auth/isModerator': () => false,
             'auth/user': {
-              id: 'u23',
+              id: 'f26bc193-605d-4465-994e-c2f328d9db01',
             },
           },
         }
@@ -75,8 +78,11 @@ describe('ProfileSlug', () => {
           wrapper.setData({
             User: [
               {
-                id: 'u3',
+                id: 'f26bc193-605d-4465-994e-c2f328d9db01',
                 name: 'Bob the builder',
+                slug: 'bob-the-builder',
+                createdAt: '2020-01-23T06:38:58.813Z',
+                about: 'Oh! I am so excited about me and myself …',
                 contributionsCount: 6,
                 shoutedCount: 7,
                 commentedCount: 8,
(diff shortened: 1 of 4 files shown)


webapp/pages/profile/_id/_slug.vue
```diff
@@ -85,6 +85,10 @@
             </ds-space>
           </template>
         </ds-card>
+        <!-- User ID -->
+        <ds-text color="softer" size="small" class="hyphenate-text" style="text-align: center;">
+          <small>{{ $t('profile.userId') }}: {{ user.id }}</small>
```
roschaefer: Can we move it into the user card? I think it belongs to the user details.

webapp/pages/profile/_id/_slug.spec.js
```diff
@@ -21,12 +21,15 @@ describe('ProfileSlug', () => {
         id: 'p23',
         name: 'It is a post',
       },
-      $t: jest.fn(),
+      $t: jest.fn(a => a),
```
roschaefer: :+1: that way you can test translations :point_down: in `wrapper.text()`

webapp/pages/profile/_id/_slug.vue
Tirokk: I understand your thinking @roschaefer !
But for me it is ugly additional technical information and therefore I kind of hide it.

What does the other @Human-Connection/hc-team and especially @alina-beck say?

webapp/pages/profile/_id/_slug.spec.js
Tirokk: Yeahhhh ;-)

webapp/pages/profile/_id/_slug.vue
mattwr18: as I posted in zoom chat
```
I guess since I didn't see any planning on this issue, I would question if there isn't another way of dealing with the user issue

Like would having it somewhere display on hover be sufficient

is there a way that we can shorten the ids, and still have them be unique, as you have pointed out? then it might look better than these super long ids that can easily break our design...
```

alina-beck: I'm with @mattwr18, I think we should have a closer look at _what exactly_ we are trying to solve with displaying the `user IDs`, _where_ we need them and _how_ users should be able to interact with them (e.g. toggle them on `click` or `hover`, in the `settings`, have them visible all the time...) and _then_ make some design decisions.

There's also this other issue dealing with `user IDs` that is closely related: #2784. And then I remember that @Mogge pointed out that the `IDs` we're using right now are not really humanly readable and thus not useful for anyone, so maybe #2786 should take priority over other `user ID` issues (also because it might make the design easier).

roschaefer: The problem is that people make screenshots and send those as email as some sort of evidence for guideline violations to our moderation team. The IDs are in the URL and you cannot see them on the screenshots. So, `hover` would not make a lot of sense here.

alina-beck: So we're defining a feature for users who refuse to use our built-in reporting system? :grin:

Tirokk: It’s more than that:
On PlanetRomeo I had several times the experience in my live that I felt this user was already somewhen contacting me under a different name. So I made a screenshot which included their IDs at that time. Later I was able to prove for me and others that I was right.

The hover does not work for several screen shot tools I was told by @DennisHack and others …

To shorten the ID(s) there are already old issues I made long ago:

- Slugs - Permanent, unbreakable, readable and text changeable slugs for users, posts etc. #209
- 🚀 [Feature] Show User ID On User Profile Page #252

Which I have extracted into:

- 🚀 [Feature] Nano IDs For Human Readability
#2786

Tirokk: I would love to have a general discussion about it @Human-Connection/hc-dev-team !

mattwr18:
can this be closed @Tirokk ??

mattwr18:
I'd say this could be closed, and reopened if we want to use it, or start from a fresh branch off `master`
@Tirokk ??

mattwr18:
closing as this `PR` is stale, and can always be reopened, if we want to continue on it's work

PR #2898, opened 2020-01-26
feat(editor): Underline markup for posts+comments

roschaefer:
## 🍰 Pullrequest
We had tiptap's underline and never used it. A user has asked for it, so
why not implement it right away?

Suggested by `@raimundo@human-connection.social`:
https://human-connection.social/post/612bf465-91ce-4be8-906e-3bde63617346/hcsocial-dein-netzwerk-news-mit-dennis-and-robert-1#commentId-b864d04f-1e8c-48c4-9ed0-98aefaed1c02


roschaefer:
@mattwr18 I think the costs of having longer build times are higher than potential benefits of the test.

PR #2921, opened 2020-01-28
refactor(factories): Refactor test factories with rosie.js

### LINKED ISSUE #2942: Refactor test factories
## :construction_worker_man: Refactor test factories

Not knowing there is a [better implementation](https://github.com/rosiejs/rosie) for NodeJS out there we rolled our own fragile implementation of test factories. Our lack of experience in the field has lead to inconsistent APIs how to use those factories and this makes it in turn really difficult to refactor the database model.

The one example which makes it obvious is #2605. I try to move all the image urls into a dedicated node with label `:Image` and it's just so hard to tell if all the tests are still passing *after* the refactoring. I hope if we use `rosie` which defines a strict way how to implement factories it will help us with refactoring in the future.

roschaefer:
## 🍰 Pullrequest

fix #2942 

we can do it

When I wrote the factories the first time it was a pain to add optional arguments and/or other factories. Rosie, https://github.com/rosiejs/rosie, has a smarter way of adding optional parameters.

This is something I need for our refactoring of images.


backend/src/db/factories.js
```diff
@@ -0,0 +1,227 @@
...
+  .after((buildObject, options) => {
+    return neode.create('Badge', buildObject)
+  })
+
+Factory.define('userWithoutEmailAddress')
+  .option('password', '1234')
+  .attrs({
+    id: uuid,
+    name: faker.name.findName,
+    email: faker.internet.email,
```
mattwr18:
```suggestion
    
```

backend/src/db/factories.js
```diff
@@ -0,0 +1,227 @@
...
+
+Factory.define('socialMedia')
+  .attrs({
+    url: 'https://mastodon.social/@Gargron',
+  })
+  .after((buildObject, options) => {
+    return neode.create('SocialMedia', buildObject)
+  })
+
+export default Factory
```
mattwr18: out of curiosity, why did  you decide to collate this all in one file instead of having separate files for each like before?

backend/src/middleware/validation/validationMiddleware.spec.js
```diff
@@ -119,20 +118,30 @@ beforeEach(async () => {
...
-      id: 'post-4-commenting',
-      authorId: 'commenting-user',
-    }),
+    Factory.build(
+      'post',
+      {
+        id: 'offensive-post',
+      },
+      {
+        authorId: 'moderating-user',
```
mattwr18: I know this isn't your code, but it doesn't really make sense, oder?
the person who wrote an offensive post is the moderating user? what they are moderating themselves?

backend/src/db/factories.js
```diff
@@ -0,0 +1,227 @@
...
+  })
+
+Factory.define('badge')
+  .attr('type', 'crowdfunding')
+  .attr('status', 'permanent')
+  .after((buildObject, options) => {
+    return neode.create('Badge', buildObject)
+  })
+
+Factory.define('userWithoutEmailAddress')
```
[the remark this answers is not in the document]
roschaefer: There is one test which needs a user without email address node. So I created two factories for the user, the `user` factory extending the `userWithoutEmailAddress` base factory.

backend/src/db/factories.js
roschaefer: I'm open for changes here. It's because this one file exports `Factory` which needs to go through all the factory definitions.

roschaefer:
@mattwr18 rosiejs is not compatible with `async/await` but seems to be easily fixed. That's the reason why seeds are not behaving deterministically.

backend/src/db/factories.js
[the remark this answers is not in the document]
mattwr18: I saw that, but you have `email` as a property, for some reason. This is confusing and should be removed, unless I haven't seen something.

PR #2930, opened 2020-01-30
fix: deploy script with new naming convention

### LINKED ISSUE #2929: 🐛 [Bug] Master build broken in deploy step
## :bug: Bugreport

We tore down and deployed the develop server with a new infrastructure using helm. All of the resources were renamed and have `deploy` prepended.
This has broken our travis build since the automated deploy script expects different names.

mattwr18:
### Issues

- fixes #2929


mattwr18:
we have those prefixes @roschaefer because you added them with the `Helm` chart
`{{ .Release.Name }}`, I chose `develop` for the develop server, and will choose `staging` for `staging` and `production` for the `production` server.

I really like this cause it means that someone can't accidentally delete a resource because of generic naming, ie `kubectl delete deployment neo4j -n human-connection` when using one context and thinking you are using another.

PR #2964, opened 2020-02-03
[WIP] feat: List view for posts

### LINKED ISSUE #2968: 🚀 [Feature] List view for posts
## :rocket: Feature
Users should be able to switch between the `grid view` we currently have on the index page and a new `list view`.

### User Problem
Not sure – there are different opinions on this:

@ogerly said that the `list view` is supposed to be a lightweight alternative to the current index page – with no images, so users can choose to save data traffic, e.g. when they're on mobile.

@DennisHack and Viktor suggested to add thumbnail images. In that case the list view would maybe just cater to a different taste and offer users more freedom? Could you clarify your intentions in a bit more detail?

### Implementation
This is already half-implemented in this PR: #2964 – it adds a button to toggle between the two views on the index page. (It doesn't play nicely with `loading more` posts yet.)

### Design & Layout
Current status of the design:
Screenshot 2020-02-03 at 12 45 21

### Validation
?

### Open questions
- [ ] Which problem are we trying to solve with this feature?
- [ ] Do we want to save the preferred view for each user?
- [ ] Should we automatically detect if a user is on mobile and activate the `list view` if that's the case?
- [ ] Do we keep the `grid view` as the `default`?
- [ ] Should we add thumbnails to the `list view` or keep them text-only?
- [ ] Is this setting only valid on the `index` page or also for the posts on the `user profile` pages and the `related posts` section of the `post page`?

ogerly:
## 🍰 Pullrequest
Adds a list view for posts on the index page – and a button to toggle between `list view` and `grid view`.

### Issues
- fixes #2968

### DIFF
--- a/webapp/components/PostCard/PostCard.vue
+++ b/webapp/components/PostCard/PostCard.vue
@@ -94,6 +94,14 @@ export default {
     HcRibbon,
     ContentMenu,
   },
+  data(){
+    return {
+      views: {
+        grid:   true,
+        list:   false,
+      }
+    }
+  },
   props: {
     post: {
       type: Object,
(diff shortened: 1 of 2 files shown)


mattwr18:
can this be closed @ogerly ??

ogerly:
close

PR #3003, opened 2020-02-05
feat: Normalize locales/json files

### LINKED ISSUE #3001: 🚀 [Feature] Nornalize the locale json files
## :rocket: Feature
We need to normalize the locale json files in order to improve the integrations pf translations coming in from lokalise.

### Implementation
 - Keys must be in alphabetical order
 - Whitespace is two blanks
 - Keys of German and English locale must be the same
 - The test must pass to get a PR merged
 - a `--fix` option to fix problems

Mogge:
### To Do

 - [x] Check if keys in en.json and de.json are the same
 - [x] Check dependencies (jq must be installed)
 - [x] Where to place the file?
 - [x] Add test to travis? How?

### Issues
- fixes #3001


roschaefer:
@Mogge see #3052 and if you're adding a `package.json` script in 3fdac60d6cfc5a55265874c61e6607b86c675d9c then do so for `missing-keys`, too.

Mogge:
In folder `webapp` you can run `yarn run locales`
This checks:
 1. The keys in `locales/de.json` are the same as in `locales/en.json`. If not, an error is signaled, showing the differing keys.
 2. If the keys are identical, it tests for each json-file in folder `locales`, that the keys are sorted alphabetically. If not, an error is signaled.

If the keys are identically in the German and English locale file, you can run `yarn locales --fix` to sort the keys.

Mogge:
The jq-script to sort the keys is necessary, because lokalise exports the keys sorted case insensitive, whereas jq by default sorts case sensitive.

scripts/translations/sort_filter.jq
```diff
@@ -0,0 +1,13 @@
+def walk(f):
```
roschaefer: Whatever that file is doing :laughing:

Mogge: Wonders

PR #3017, opened 2020-02-06
fix: Search for Hyphens

### LINKED ISSUE #2798: 🐛 [Bug] Searches with a hyphen return no results
## :bug: Bugreport

A search that contains a hyphen returns no results, neither posts nor users.

Example 1:

hc-bug-search-doesnt-return-dash-example-1a

Example 2:
hc-bug-search-doesnt-return-dash-example-1b

Example 3:
hc-bug-search-doesnt-return-dash-example-2

1. Type in a string containing a hyphen in the search field.
2. The search results show "Nothing found".

### Expected behavior

The query returns correct results.

### Version & Environment
 Type: [any]      
 - OS: [any]      
 - Browser: [any] 
 - Version [any]  
 - Device: [any]

Mogge:
## 🍰 Pullrequest
In the search query, the hyphen is replaced by a space.

### Issues
- fixes #2798

### DIFF
--- a/backend/src/schema/resolvers/searches.js
+++ b/backend/src/schema/resolvers/searches.js
@@ -8,7 +8,8 @@ export default {
       // see http://lucene.apache.org/core/8_3_1/queryparser/org/apache/lucene/queryparser/classic/package-summary.html#package.description
       const myQuery = query
         .replace(/\s+/g, ' ')
-        .replace(/[[@#:*~\\$|^\]?/"'(){}+?!,.-;]/g, '')
+        .replace(/[[@#:*~\\$|^\]?/"'(){}+?!,.;]/g, '')
+        .replace(/-/g, ' ')
         .split(' ')
         .map(s => (s.toLowerCase().match(/^(not|and|or)$/) ? '"' + s + '"' : s + '*'))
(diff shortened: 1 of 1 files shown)


backend/src/schema/resolvers/searches.js
```diff
@@ -8,7 +8,8 @@ export default {
       // see http://lucene.apache.org/core/8_3_1/queryparser/org/apache/lucene/queryparser/classic/package-summary.html#package.description
       const myQuery = query
         .replace(/\s+/g, ' ')
-        .replace(/[[@#:*~\\$|^\]?/"'(){}+?!,.-;]/g, '')
+        .replace(/[[@#:*~\\$|^\]?/"'(){}+?!,.;]/g, '')
+        .replace(/-/g, ' ')
```
roschaefer: This, as well, is the job of the tokenizer. Maybe you could search for "JavaScript tokenizer" to find a less hacky solution. Or look for a solution that is related to elastic search: "elastic search normalize query". 

You could add tests for this behaviour as well, at least for documentation purposes.

Mogge:
I found a better solution.

PR #3018, opened 2020-02-06
fix: Search For Words With Less Than Three Chars

### LINKED ISSUE #2799: 🐛 [Bug] Search input with a space ignores minimal character length
## :bug: Bugreport

If the search input contains at least two words separated by a space the search results contain matches that ignore the minimal character length.

Example 1:
hc-bug-seach-with-space-example-1

Example 2:
hc-bug-seach-with-space-example-2

1. In the search field, type in at least two words with a space in between, see the examples.
2. The search results show also records that only match the single character after the space.

### Expected behavior

Search results should be more accurate: "martina s" should return results that contain _martina AND s_ instead of _martina OR s_.

### Version & Environment
 Type: [any]      
 - OS: [any]      
 - Browser: [any] 
 - Version [any]  
 - Device: [any]

Mogge:
## 🍰 Pullrequest

Hyphens are replaced by one space, which also fixes the hyphen search, this time in the frontend.
If the search string contains spaces, only words with at least three chars are passed to the backend.

### Issues
- fixes #2799

### DIFF
--- a/webapp/components/generic/SearchableInput/SearchableInput.vue
+++ b/webapp/components/generic/SearchableInput/SearchableInput.vue
@@ -89,9 +89,18 @@ export default {
     },
     handleInput(event) {
       clearTimeout(this.searchProcess)
-      this.value = event.target ? event.target.value.replace(/\s+/g, ' ').trim() : ''
+      this.value = event.target ? event.target.value.replace(/[-\s]+/g, ' ').trim() : ''
       this.unprocessedSearchInput = this.value
-      if (isEmpty(this.value) || this.value.replace(/\s+/g, '').length < 3) {
+      if (isEmpty(this.value)) {
+        return
+      }
+      if (this.value.includes(' ')) {
+        this.value = this.value
+          .split(' ')
+          .filter(word => word.length > 2)
+          .join(' ')
+      }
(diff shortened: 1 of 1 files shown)


webapp/components/generic/SearchableInput/SearchableInput.vue
```diff
@@ -89,9 +89,18 @@ export default {
...
+      if (isEmpty(this.value)) {
+        return
+      }
+      if (this.value.includes(' ')) {
+        this.value = this.value
+          .split(' ')
+          .filter(word => word.length > 2)
+          .join(' ')
+      }
+      if (this.value.length < 3) {
```
roschaefer: This is a quite hacky solution. I'm going to approve this but I feel bad about it. Do you know "stopwords"? It's a feature of search indices and I think that would have been the proper solution.

@Mogge if you feel like it, you could add an integration test for it to document this behaviour at least.

roschaefer:
Not gonna merge this to give @Mogge time to add tests if he feels like it.

Lulalaby:
Seems like he don't want it.
Merging?

Mogge:
Better solution available

PR #3034, opened 2020-02-10
fix: Display unblock feature only for blocking user

### LINKED ISSUE #3033: 🐛 [Bug]Blocked users can unblock themselves
## :bug: Bugreport

**HIGH PRIORITY BUG**
Blocked users should not see a button or link to unblock themselves since 
a) they should not be able to find out who has blocked them
b) they should not be able to remove this. 

1. log in as a user(user1)
2. block another user(user2)
3. login as user2
4. visit profile of user1
5. see that you have a button and a link to unblock the user
6. unblock yourself.
7. :rofl: 

### Expected behavior

It should be solely the control of the blocking user to decide this.

### Additional context

This was introduced with Block users 2.0 where we wanted to make commenting on another user's posts only possible if there is no `BLOCKED` relationship in either direction. We still want this functionality, but we need a second query `hasBlocked` to look for the direction of the blocking relationship.

mattwr18:
- add isBlocked query to check if userA initiated the blocking and, if
  true, only display link in content menu and button to unblock.
- userB should not have an indication they have been blocked, or a button/link that tries to remove a relationship that originated from userA.
- add cypress tests to make sure we don't break this in the future.

### Issues

- fixes #3033


mattwr18:
> I'm going to approve this but I don't understand how users can unblock themselves? You did not change any permissions so I don't know how that prevents from users unblocking them for other users?

yeah, I actually overstated what the user could do... they currently would have a button to unblock the other user, but if clicked, it doesn't unblock themselves, it removes a blocked relationship if it exists between the current user and the owner of the profile. 
This is still misleading, but not as dangerous as I originally thought. I have updated the title

PR #3041, opened 2020-02-11
build(deps): Node v13 compatbility

roschaefer:
## 🍰 Pullrequest
Force update of `fs-capacitor`, see https://github.com/jaydenseric/graphql-upload/issues/170#issue-527336432

### Issues
I am testing file uploads on my machine. My local `node` has version `v13.7.0`. This resolution fix does the trick for me.


mattwr18:
I tried this, and it didn't seem to work for me, but I can test it again @roschaefer

PR #3042, opened 2020-02-11
refactor(cypress): Speed up builds, avoid login through UI

### LINKED ISSUE #2887: 🚀 [Feature] Refactor login step definitions to reduce build server time
## :rocket: Feature
Our build server takes ~30mins until we get a feedback. There is potential for optimization.

### User Problem
Look at how we login a user during fullstack testing: https://github.com/Human-Connection/Human-Connection/blob/master/cypress/support/commands.js#L51

It is bad practice to login a user through the UI. We should do that once (when we test login). But if we need it during the setup, we can speed up build server time by generating a valid JWT and set browser cookies with it. Our browser should be immediately logged in and we save time on each cucumber test which requires authentication.

### Implementation
I would suggest that you try to use [setCookie](https://docs.cypress.io/api/commands/setcookie.html#Syntax#article) and maybe you can re-use [encode](https://github.com/Human-Connection/Human-Connection/blob/master/backend/src/jwt/encode.js) to generate the token. Don't forget to add `CONFIG.JWT_SECRET` to cypress [here](https://github.com/Human-Connection/Human-Connection/blob/master/cypress/plugins/index.js#L26).

### Design & Layout
:woman_shrugging: 

### Validation
We could compare build server times before and after we merge this PR into `master`.

### Additional context
https://github.com/heartcombo/devise/wiki/How-To%3A-Test-with-Capybara

mattwr18:
### Issues

- fixes #2887 
### Todo

- [X] None

### DIFF
--- a/cypress/integration/common/steps.js
+++ b/cypress/integration/common/steps.js
@@ -58,8 +58,15 @@ Given("I log in as {string}", name => {
 
 Given("the {string} user searches for {string}", (_, postTitle) => {
   cy.logout()
-    .login({ email: annoyingParams.email, password: '1234' })
-    .get(".searchable-input .ds-select-search")
+  cy.neode()
+    .first("User", {
+      id: "annoying-user"
+    })
+    .then(async user => {
+      const userJson = await user.toJson()
+      cy.login(userJson)
+    })
+  cy.get(".searchable-input .ds-select-search")
     .focus()
     .type(postTitle);
 });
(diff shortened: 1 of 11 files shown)


cypress/integration/common/steps.js
```diff
@@ -35,13 +35,38 @@ const annoyingParams = {
 };
 
 Given("I am logged in", () => {
-  cy.login(loginCredentials);
+  cy.neode()
+    .first("User", {
+      name: narratorParams.name
+    })
+    .then(async user => {
```
roschaefer: Doesn't that lead to warning of cypress complaining about using promises?

package.json
```diff
@@ -52,5 +52,8 @@
   },
   "resolutions": {
     "set-value": "^2.0.1"
+  },
+  "dependencies": {
```
roschaefer: Development dependencies!

cypress/integration/common/steps.js
mattwr18: maybe, but honestly we'd have to refactor the whole cypress setup since we have these in many of the factories

PR #3043, opened 2020-02-11
feat: Introduce graphql image type

roschaefer:
## 🍰 Pullrequest
This is a first step to implement responsive images. We only refactor the schema here. Nothing needs to be uploaded, no files need to be moved.

### Todo
- [ ] Implement and check user avatar file upload
- [ ] Double-check and fix image blurring
- [ ] Refactor code
- [ ] Check how many `user.coverImg` have survived our migration from alpha


backend/src/db/migrations/20200125010142-refactor_all_images_to_separate_type.js
```diff
@@ -0,0 +1,101 @@
...
+    `,
+        `
+      MATCH(user)-[:AVATAR_IMAGE]->(avatar:Image)
+      SET user.avatar = avatar.url
+      DETACH DELETE avatar
+    `,
+        `
+      MATCH(user)-[:COVER_IMAGE]->(coverImage:Image)
+      SET user.coverImg = coverImage.url
+      DETACH DELETE coverImage
```
roschaefer: Maybe just `REMOVE user.coverImg`.

mattwr18:
this is still `[WIP]`, oder @roschaefer ?

webapp/components/ContributionForm/ContributionForm.vue
```diff
@@ -212,11 +207,10 @@ export default {
         title,
         content,
         image,
-        teaserImage,
-        imageAspectRatio,
         categoryIds,
         blurImage,
       } = this.form
+      if (image) image.blurred = blurImage
```
alina-beck: not sure, but `this.blurImage`?

webapp/components/Upload/index.vue
```diff
@@ -41,6 +41,12 @@ export default {
       hover: false,
     }
   },
+  computed: {
+    avatarUrl() {
+      const { avatar } = this.user
+      return avatar && avatar.url
```
alina-beck: anybody else feel like using the [optional chaining operator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining)? :sweat_smile:

it's still experimental technology but I'd like to see if there's a webpack plugin for it already...

webapp/plugins/vue-filters.js
```diff
@@ -82,6 +82,7 @@ export default ({ app = {} }) => {
       return contentExcerpt
     },
     proxyApiUrl: url => {
+      url = url && (url.url || url)
```
alina-beck: very unclear to me what's going on here... could we call the input... `input`? and set it as `url` when there is no `input.url`? or am I reading this wrong?

backend/src/models/Post.js
```diff
@@ -4,6 +4,12 @@ export default {
   id: { type: 'string', primary: true, default: uuid },
   activityId: { type: 'string', allow: [null] },
   objectId: { type: 'string', allow: [null] },
+  image: {
+    type: 'relationship',
+    relationship: 'TEASER_IMAGE',
```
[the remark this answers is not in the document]
mattwr18: It is called `TeaserImage` because that is what it was called in the old `alpha` system, and I copied the naming convention without thinking about whether it should be renamed. 
It does however kinda make sense to me cause the image teases what the article/post is about, at least that is the way some use the image. 
Having said that, I have no issue renaming it, and `HeroImage` would be a good choice, in my mind.

webapp/components/Upload/index.vue
mattwr18: you want more fun configuring webpack? I got no problem with using it, although the `experimental` part makes me a bit nervous.

backend/src/db/migrations/20200125010142-refactor_all_images_to_separate_type.js
```diff
@@ -0,0 +1,101 @@
...
+export async function up() {
+  const driver = getDriver()
+  const session = driver.session()
+  const writeTxResultPromise = session.writeTransaction(async txc => {
+    const runs = await Promise.all(
+      [
+        `
+      MATCH (post:Post)
+      WHERE post.image IS NOT NULL
+      CREATE (post)-[:TEASER_IMAGE]->(image:Image)
```
mattwr18:
```suggestion
      CREATE (post)-[:HERO_IMAGE]->(image:Image)
```
I think we have started to rename this, so let's keep it consistent?

backend/src/db/migrations/20200125010142-refactor_all_images_to_separate_type.js
```diff
@@ -0,0 +1,101 @@
...
+      REMOVE user.coverImg
+    `,
+      ].map(s => txc.run(s)),
+    )
+    return runs.map(({ summary }) => summary)
+  })
+
+  try {
+    const stats = await writeTxResultPromise
+    console.log('Splitted all images from users and posts.')
```
mattwr18:
```suggestion
    console.log('Create image nodes from all user avatars and post images.')
```

backend/src/db/migrations/20200125010142-refactor_all_images_to_separate_type.js
```diff
@@ -0,0 +1,101 @@
...
+}
+
+export async function down() {
+  const driver = getDriver()
+  const session = driver.session()
+  const writeTxResultPromise = session.writeTransaction(async txc => {
+    const runs = await Promise.all(
+      [
+        `
+      MATCH (post)-[:TEASER_IMAGE]->(image:Image)
```
mattwr18:
```suggestion
      MATCH (post)-[:HERO_IMAGE]->(image:Image)
```

backend/src/db/migrations/20200125010142-refactor_all_images_to_separate_type.js
mattwr18: good question... we can discuss

backend/src/db/migrations/20200213230248-add_unique_index_to_image_url.js
```diff
@@ -0,0 +1,53 @@
...
+  } catch (error) {
+    // eslint-disable-next-line no-console
+    console.log(error)
+    await transaction.rollback()
+    // eslint-disable-next-line no-console
+    console.log('rolled back')
+  } finally {
+    session.close()
+  }
+}
```
mattwr18: why do we need a migration for this? can't we just run `prod:migrate init` after having added a `model` for it?

backend/src/db/seed.js
```diff
@@ -472,24 +478,28 @@ const languages = ['de', 'en', 'es', 'fr', 'it', 'pt', 'pl']
         'post',
         {
           id: 'p10',
-          imageBlurred: true,
         },
         {
           categoryIds: ['cat10'],
           author: dewey,
+          image: Factory.build('image', {
+            blurred: true,
```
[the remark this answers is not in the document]
mattwr18: I don't think so!?

backend/src/jwt/decode.spec.js
```diff
@@ -69,24 +69,25 @@ describe('decode', () => {
...
-      it('returns user object except email', async () => {
+      it('returns user object without email', async () => {
         await expect(decode(driver, authorizationHeader)).resolves.toMatchObject({
           role: 'user',
           name: 'Jenny Rostock',
-          avatar: 'https://s3.amazonaws.com/uifaces/faces/twitter/sasha_shestakov/128.jpg',
           id: 'u3',
-          email: null,
           slug: 'jenny-rostock',
+          // email: undefined,
```
mattwr18: ?

backend/src/models/Image.js
```diff
@@ -0,0 +1,7 @@
+export default {
+  url: { primary: true, type: 'string', uri: { allowRelative: true } },
```
mattwr18:
```suggestion
  url: { primary: true, type: 'string', unique: 'true', uri: { allowRelative: true } },
```

mattwr18: this in my mind would make the extra `migration` unnecessary, oder?

backend/src/models/Image.js
```diff
@@ -0,0 +1,7 @@
...
+  alt: { type: 'string' },
```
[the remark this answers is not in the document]
mattwr18: that is unfortunate, why's that? doesn't that mean our site takes a huge hit for accessibility?

backend/src/schema/resolvers/posts.js
```diff
@@ -77,14 +77,16 @@ export default {
...
-      params = await fileUpload(params, { file: 'imageUpload', url: 'image' })
+      delete params.image
       params.id = params.id || uuid()
       const session = context.driver.session()
       const writeTxResultPromise = session.writeTransaction(async transaction => {
         const createPostTransactionResponse = await transaction.run(
           `
-            CREATE (post:Post {params})
+            CREATE (post:Post)
+            SET post += $params
```
mattwr18: why should this be `+=` ?? 
it's creating a post, so it should never have any `params` to add to, oder?

backend/src/schema/resolvers/posts.js
```diff
@@ -94,14 +96,26 @@ export default {
...
         )
-        return createPostTransactionResponse.records.map(record => record.get('post').properties)
+        const [post] = createPostTransactionResponse.records.map(record => record.get('post'))
+        if (imageInput) {
+          const image = await createImage({ imageInput, transaction })
+          await transaction.run(
+            `
+            MATCH (post:Post {id: $post.id})
+            MATCH (image:Image {url: $image.url})
+            MERGE (post)-[:TEASER_IMAGE]->(image)
```
mattwr18:
```suggestion
            MERGE (post)-[:HERO_IMAGE]->(image)
```

backend/src/schema/resolvers/posts.js
```diff
@@ -142,17 +157,49 @@ export default {
...
             updatePostCypher,
             updatePostVariables,
           )
-          return updatePostTransactionResponse.records.map(record => record.get('post').properties)
+          const [post] = updatePostTransactionResponse.records.map(record => record.get('post'))
+          if (imageInput) {
+            if (imageInput.upload) {
+              const previousImageResponse = await transaction.run(
+                `
+                MATCH (post:Post {id: $post.id})-[:TEASER_IMAGE]->(image:Image)
```
mattwr18:
```suggestion
                MATCH (post:Post {id: $post.id})-[:HERO_IMAGE]->(image:Image)
```

backend/src/models/Image.js
roschaefer: Primary keys are always unique. The extra migration is an training for us to create migrations when we setup new indices. So, during a helm upgrade the migrations will run but no `init`.

backend/src/db/migrations/20200213230248-add_unique_index_to_image_url.js
roschaefer: I thought we would only run `db:migrate up` after a helm upgrade? This is a training for us to create migrations when we add new constraints.

backend/src/db/seed.js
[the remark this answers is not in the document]
roschaefer: `Factory.build('image', ...) ` makes sure that there are default values when you don't explicitly override them when you call the factory.

backend/src/jwt/decode.spec.js
roschaefer: The `authorizationHeader` does not contain the email anymore. It's a small code creep when I removed `.avatar` from the list of attributes passed to `encode`.

backend/src/models/Image.js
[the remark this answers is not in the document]
roschaefer: I considered to remove it from the PR but changed my mind. I have a lot of backend tests where I check whether I can update metadata while re-using the existing image node. The `alt` attribute comes in handy and has quite an obvious benefit for later. It's one of the rare occasions where I'm willing add code anticipating future requirements. Usually I avoid code which does not have an immediate use case.

backend/src/schema/resolvers/posts.js
roschaefer: You're right. But it does not hurt either, or does it? :thinking:

backend/src/schema/resolvers/registration.js
```diff
@@ -35,27 +31,39 @@ export default {
...
+      delete args.email
+      args = encryptPassword(args)
+
+      const { driver } = context
+      const session = driver.session()
+      const writeTxResultPromise = session.writeTransaction(async transaction => {
+        const createUserTransactionResponse = await transaction.run(
+          `
+            MATCH(email:EmailAddress {nonce: $nonce, email: $email})
+            WHERE NOT (email)-[:BELONGS_TO]->()
```
mattwr18:
```suggestion
            WHERE NOT (email)-[:BELONGS_TO]->(:User)
```

backend/src/schema/resolvers/users.spec.js
```diff
@@ -542,6 +542,12 @@ describe('DeleteUser', () => {
               mutate({ mutation: deleteUserMutation, variables }),
             ).resolves.toMatchObject(expectedResponse)
           })
+
+          it('deletes user avatar and post hero images', async () => {
+            await expect(neode.all('Image')).resolves.toHaveLength(22)
+            await mutate({ mutation: deleteUserMutation, variables })
+            await expect(neode.all('Image')).resolves.toHaveLength(20)
+          })
```
mattwr18: where are these magic numbers coming from?

webapp/components/Badges.spec.js
```diff
@@ -2,20 +2,33 @@ import { shallowMount } from '@vue/test-utils'
...
+    const Wrapper = () => {
+      return shallowMount(Badges, { propsData })
+    }
 
-  it('has class "hc-badges"', () => {
-    expect(wrapper.contains('.hc-badges')).toBe(true)
-  })
+    it('renders', () => {
+      expect(Wrapper().is('div')).toBe(true)
+    })
```
mattwr18: unnecessary test case... if it doesn't render all other tests would fail

backend/src/schema/resolvers/images/images.js
```diff
@@ -0,0 +1,120 @@
...
+    `,
+    { resource },
+  )
+  const [existingImage] = txResult.records.map(record => record.get('image'))
+  const { upload } = imageInput
+  if (!(existingImage || upload)) throw new UserInputError('Cannot find image for given resource')
+  if (existingImage && upload) deleteImageFile(existingImage, deleteCallback)
+  const url = await uploadImageFile(upload, uploadCallback)
+  const { alt, sensitive, aspectRatio } = imageInput
+  const image = { alt, sensitive, aspectRatio, url }
```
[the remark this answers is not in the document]
roschaefer: `imageInput` contains the `upload`. You don't want that to be part of `image`.

backend/src/schema/resolvers/images/images.js
```diff
@@ -0,0 +1,120 @@
...
+  txResult = await transaction.run(
+    `
+    MATCH (resource {id: $resource.id})
+    MERGE (resource)-[:${relationshipType}]->(image:Image)
+    SET image += $image
+    SET image.createdAt = toString(datetime())
```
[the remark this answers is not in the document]
roschaefer: Yes, earlier I would assume that you wouldn't update an image but delete/create. But you can update meta data and keep the image file. So yes, good point.

backend/src/jwt/decode.spec.js
roschaefer: I'm going to remove. I thought it could be a hint that `email` is not included in the token.

backend/src/schema/resolvers/images/images.js
```diff
@@ -0,0 +1,120 @@
...
+  return uploadCallback({
+    createReadStream,
+    destination: `/uploads/${uniqueFilename}`,
+    mimetype,
+  })
+}
+
+const sanitizeRelationshipType = relationshipType => {
+  // Cypher query language does not allow to parameterize relationship types
+  // See: https://github.com/neo4j/neo4j/issues/340
```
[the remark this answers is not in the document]
roschaefer: that's `MATCH` not `MERGE`

backend/src/schema/resolvers/images/images.spec.js
```diff
@@ -0,0 +1,347 @@
...
+            })
+            const txResult = await transaction.run('RETURN "Hello" as result')
+            const [result] = txResult.records.map(record => record.get('result'))
+            return result
+          })
+        } finally {
+          session.close()
+        }
+        await expect(neode.all('Image')).resolves.toHaveLength(0)
+        await expect(someString).toEqual('Hello')
```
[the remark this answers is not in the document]
roschaefer: Kind of. I would argue that it is the happy path for the actual test case where I introduce a bug which should roll back the entire transaction.

backend/src/schema/resolvers/users.spec.js
roschaefer: Well, I don't know myself, I just ran the test and checked how many images are set up in the `beforeEach`s prior to the test case :point_up: then I wanted it to remove more than 1 image.

backend/src/schema/resolvers/images/images.js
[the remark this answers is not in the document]
mattwr18: thanks for the explanation... I had forgotten about that

backend/src/schema/resolvers/images/images.js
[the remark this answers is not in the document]
mattwr18: yep, I remember we discussed this

backend/src/schema/resolvers/users.spec.js
mattwr18: humm... don't really like that

PR #3051, opened 2020-02-12
refactor(ci): Move translations to scripts/ folder

roschaefer:
And refactor the script a little

## 🍰 Pullrequest
@Mogge here are some improvements. The scripts is not added to our `.travis.yml` and should fail the build.

### Todo
- [ ] Run `./scripts/sort-translations.sh --fix`


Mogge:
Wouldn't it be nice if the script was living somewhere under the webapp folder? And then invoke it by something like `yarn run locales --fix`, to have a nice shorthand?

PR #3056, opened 2020-02-12
test: Write cypress tests for ImageUploader

### LINKED ISSUE #2256: 🐛 [Bug] Uploading a SVG file turns it into pixelated JPG
## :bug: Bugreport

When I write a post with an SVG file like this:

svg

What I see when uploading the image is:

Screenshot - 2019-11-16T144831 171

1. Visit https://develop.human-connection.org/post/88d15ff6-f9bf-4047-8e00-a46aa70c6900/upload-a-svg-1

Alternatively:
1. Login at https://develop.human-connection.org/
2. Create a post https://develop.human-connection.org/post/create
3. Upload a SVG file for the teaser image
5. See the result

### Expected behavior

The SVG file should be uploaded as is. We should support SVG as post image.

### Additional context

Here is the inspiration for the bug report: https://human-connection.social/post/cef51859-2253-4f04-976c-40a229903ca1/novyj-logotip#commentId-29798cf9-c375-4dac-90bb-ca5d83e91f51

### LINKED ISSUE #3054: 🔧 [Tests] Add e2e tests for post image uploads
## :zap: Testing ticket

Add cypress tests to test the functionality of uploading images to posts
### Motive

We should be testing this to ensure the functionality doesn't get broken unintentionally, and also to have executable documentation.

mattwr18:
- have I missed any important features? too many tests?

### Issues

- fixes #3054 
- fixes #2256 

### Todo

- [X] None


backend/src/schema/resolvers/fileUpload/index.js
```diff
@@ -1,24 +1,27 @@
...
-export default async function fileUpload(params, { file, url }, uploadCallback = storeUpload) {
+export default async function fileUpload(params, { file, url }, uploadCallback = localFileUpload) {
   const upload = params[file]
   if (upload) {
     const { createReadStream, filename } = await upload
-    const { name } = path.parse(filename)
-    const fileLocation = `/uploads/${Date.now()}-${slug(name)}`
-    await uploadCallback({ createReadStream, fileLocation })
+    const { name, ext } = path.parse(filename)
+    const uniqueFilename = `/uploads/${uuid()}-${slug(name)}${ext}`
```
roschaefer: @mattwr18 would comment "code creep" here.

mattwr18: yeah, I thought of that, but the challenge I was facing is that we can't cypress test if we call `canvas.toBlob` or at least I couldn't find a workaround. Until I remembered that we wanted to stop compressing/converting `png/svg`.
This is a great example of an instance where if we had used `TDD`, we would have implemented the feature in a way that dealt with bugs we now have

cypress/integration/common/post.js
```diff
@@ -83,3 +86,71 @@ And("the post with title {string} has a ribbon for pinned posts", (title) => {
...
+    .get('.ds-flex-item > .ds-form-item .ds-select ')
+    .click()
+    .get('.ds-select-option')
+    .eq(languages.findIndex(l => l.code === 'en'))
+    .click()
+})
+
+Then("the post was saved successfully with the {string} teaser image", condition => {
+  cy.get(".ds-card-content > .ds-heading")
+    .should("contain", condition === 'new' ? 'new post' : 'to be updated')
```
roschaefer: You could implement this with less line of code, e.g. using a mapping.

cypress/integration/common/post.js
```diff
@@ -83,3 +86,71 @@ And("the post with title {string} has a ribbon for pinned posts", (title) => {
...
+    .should("contain", 'new post content')
+    .get('.post-page')
+    .should('exist')
+    .get('.post-page img.ds-card-image')
+    .should('not.exist')
+})
+
+Then('I should be able to remove it', () => {
+  cy.get('.crop-cancel')
+    .click()
```
roschaefer: Already clicking on it? You don't want to check the effect?

webapp/components/TeaserImage/TeaserImage.vue
```diff
@@ -113,29 +114,41 @@ export default {
     },
     cropImage() {
       this.showCropper = false
+      if (this.file.type === 'image/png' || this.file.type === 'image/svg+xml') {
```
roschaefer: Hm, what about other file extensions? Couldn't you check for `jpg` only here?

mattwr18: I thought about that... I can change it

cypress/integration/common/post.js
mattwr18: what do you mean? like check that there is a `thumbnail-preview`?

cypress/integration/common/post.js
mattwr18: I'm not sure what you mean...

mattwr18: I think I did what you suggested

backend/src/db/factories.js
```diff
@@ -27,7 +27,7 @@ export const cleanDatabase = async (options = {}) => {
 Factory.define('category')
   .attr('id', uuid)
   .attr('icon', 'globe')
-  .attr('name', 'global-peace-nonviolence')
+  .attr('name', 'Global Peace & Nonviolence')
```
roschaefer: This `name` gets translated, @mattwr18

mattwr18: no, the name does not, it is the `slug` that gets translated, and because it is calling this twice, there is no key in `en.json` so it just appears as `category.name.null`. This is a separate issue that was introduced by the `rosie` refactor, but I realized it doesn't matter, cause I don't care what category the user creates the post with, so I just select the first button

PR #3057, opened 2020-02-12
fix: [WIP] Remove Hover Menu from User Teaser

### LINKED ISSUE #3004: 🚀 [Feature] Remove Hover Menu from User Teaser
## :rocket: Feature
The `hover menu` should be completely removed from the `user teaser` component.

### User Problem
At the moment, the `user teaser` has two effects and a third one in planning:
- it opens a small popup on hover
- it links to the `user profile` on click
- planned: it toggles the `user id` on button click

Three different functionalities on one element is no good practice, it clutters the page and is hard to use. While the other two functionalities cannot (sensibly) be moved somewhere else, removing the hover menu only requires one further click to achieve the same goal (e.g. see follower numbers or follow the person).

### Implementation
Remove the `hover menu` from the `user teaser` component – make sure to also delete related CSS. :wink:

### Validation
We will wait for user feedback and see if many people complain about the missing menu. If so, it is easily reintroduced. If not, it probably wasn't used much in the first place.

abdellani:
### Issues
- Fixes #3004 

### Todo

- [X] None

### DIFF
--- a/webapp/components/UserTeaser/UserTeaser.vue
+++ b/webapp/components/UserTeaser/UserTeaser.vue
@@ -3,29 +3,22 @@
     <user-avatar v-if="showAvatar" size="small" />
     <span class="info anonymous">{{ $t('profile.userAnonym') }}</span>
   </div>
-  <div v-else
-      :class="[{ 'disabled-content': user.disabled }]"
-    placement="top-start"
-
-  >
-      <nuxt-link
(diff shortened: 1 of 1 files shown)


mattwr18:
closed in favor of https://github.com/Human-Connection/Human-Connection/pull/3093
thanks @abdellani

PR #3063, opened 2020-02-13
feature: Delete_user_as_admin_through_API_only

### LINKED ISSUE #3038: 🚀 [Feature] Delete user as admin through API only
## :rocket: Feature
This is a required step in order to finish #17. 

### Problem
So, let's say for safety reasons we might not implement #17 at all - but we allow admins to delete users. Admins can use external UIs for a graphQL endpoint e.g. [Postman](https://www.postman.com/).

### Implementation
Basically you only need to change [this line](https://github.com/Human-Connection/Human-Connection/blob/master/backend/src/middleware/permissionsMiddleware.js#L136) into something like this:

```js
      DeleteUser: or(isDeletingOwnAccount, isAdmin),
```
And then provide sufficient backend testing for this critical security feature.

### Design & Layout
None.

ogerly:
## 🍰 Pullrequest
Delete_user_as_admin_through_API_only

Human Connection API - HC.postman_collection.json
https://github.com/ogerly/HC_postman_collection

### Issues
- fixes #3038

### DIFF
--- a/backend/src/schema/resolvers/users.spec.js
+++ b/backend/src/schema/resolvers/users.spec.js
@@ -244,7 +244,7 @@ describe('UpdateUser', () => {
   })
 })
 
-describe('DeleteUser as another user', () => {
+describe('DeleteUser', () => {
   const deleteUserMutation = gql`
     mutation($id: ID!, $resource: [Deletable]) {
       DeleteUser(id: $id, resource: $resource) {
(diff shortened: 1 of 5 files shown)


backend/src/schema/resolvers/users.spec.js
```diff
@@ -568,3 +568,330 @@ describe('DeleteUser', () => {
...
+  })
+
+  describe('unauthenticated', () => {
+    it('throws authorization error', async () => {
+      const { errors } = await mutate({ mutation: deleteUserMutation, variables })
+      expect(errors[0]).toHaveProperty('message', 'Not Authorised!')
+    })
+  })
+
+  describe('authenticated as Admin', () => {
```
mattwr18: can you add a test, for authenticated as `moderator` and `user` and make sure it throws an `Authorization` error?

backend/src/schema/resolvers/users.spec.js
```diff
@@ -568,3 +568,330 @@ describe('DeleteUser', () => {
...
+      about: 'along with my about',
+      id: 'u343',
+    })
+    await Factory.build(
+      'user',
+      {
+        id: 'not-my-account',
+      },
+      {
+        email: 'friends-account@example.org',
```
mattwr18:
```suggestion
        email: 'harassing-user@example.org',
```

maybe? not a big deal, but why is the `admin` deleting a `friends` account?

backend/src/schema/resolvers/users.spec.js
```diff
@@ -568,3 +568,330 @@ describe('DeleteUser', () => {
...
+        {
+          email: 'admin@example.org',
+          password: '1234',
+        },
+      )
+      authenticatedUser = await admin.toJson()
+    })
+ 
+
+    describe('attempting to delete a foreign account by an Admin', () => {
```
mattwr18:
```suggestion
    describe('deleting a user account', () => {
```
this would then read like `authenticated as Admin deleting a user account`

backend/src/schema/resolvers/users.spec.js
```diff
@@ -568,3 +568,330 @@ describe('DeleteUser', () => {
...
+              id: 'c156',
+              content: "A comment by someone else on user u343's post",
+            },
+            {
+              postId: 'p139',
+            },
+          )
+        })
+
+        it("deletes my account, but doesn't delete posts or comments by default", async () => {
```
mattwr18:
```suggestion
        it("deletes user account, but doesn't delete posts or comments by default", async () => {
```

backend/src/schema/resolvers/users.spec.js
```diff
@@ -568,3 +568,330 @@ describe('DeleteUser', () => {
...
+              },
+            },
+            errors: undefined,
+          }
+          await expect(mutate({ mutation: deleteUserMutation, variables })).resolves.toMatchObject(
+            expectedResponse,
+          )
+        })
+
+        describe('deletion of all post requested', () => {
```
mattwr18:
```suggestion
        describe('deletion of all posts requested', () => {
```

backend/src/schema/resolvers/users.spec.js
```diff
@@ -568,3 +568,330 @@ describe('DeleteUser', () => {
...
+          beforeEach(() => {
+            variables = { ...variables, resource: ['Post'] }
+          })
+
+          describe("marks user's posts as deleted", () => {
+            it('posts on request', async () => {
```
mattwr18:
```suggestion
            it('on request', async () => {
```

backend/src/schema/resolvers/users.spec.js
```diff
@@ -568,3 +568,330 @@ describe('DeleteUser', () => {
...
+              },
+              errors: undefined,
+            }
+            await expect(
+              mutate({ mutation: deleteUserMutation, variables }),
+            ).resolves.toMatchObject(expectedResponse)
+          })
+        })
+
+        describe('deletion of all post and comments requested', () => {
```
mattwr18:
```suggestion
        describe('deletion of all posts and comments requested', () => {
```

backend/src/schema/resolvers/users.spec.js
```diff
@@ -568,3 +568,330 @@ describe('DeleteUser', () => {
...
+            ).resolves.toMatchObject(expectedResponse)
+          })
+        })
+      })
+
+      describe('connected `EmailAddress` nodes', () => {
+        it('will be removed completely', async () => {
+          await expect(neode.all('EmailAddress')).resolves.toHaveLength(3)
+          await mutate({ mutation: deleteUserMutation, variables })
+          await expect(neode.all('EmailAddress')).resolves.toHaveLength(2)
```
mattwr18: without looking into this, it's not clear to me why there is `3` to start and then `2`... One is the user to be deleted, one is the admin, who is the third user?

backend/src/schema/resolvers/users.spec.js
mattwr18: or are there already tests for this :point_up_2: ??

ogerly: ok

ogerly:
@mattwr18  I have adjusted everything, but I have no real idea if this is really right with the tests? theoretically you can pull everything together. but i still lack the experience :)

backend/src/schema/resolvers/users.spec.js
```diff
@@ -244,7 +244,435 @@ describe('UpdateUser', () => {
...
+            await expect(
+              mutate({ mutation: deleteUserMutation, variables }),
+            ).resolves.toMatchObject(expectedResponse)
+          })
+        })
+      })
+
+      describe('connected `EmailAddress` nodes', () => {
+        it('will be removed completely', async () => {
+          await expect(neode.all('EmailAddress')).resolves.toHaveLength(2)
```
roschaefer: @ogerly similar to this test case it would be great if you could check with `neode` or a `neo4j` driver connection *directly in the database* if the data was really wiped out.

1. You take the `id` of the response or the `variables`
2. You run a database statement to check that the data is there
3. You run your delete mutation
4. You run the same database statement to check that the data *is not there* anymore

roschaefer: And I'm talking about the scope `deletion of all posts and comments requested` for example. :point_up:

PR #3083, opened 2020-02-17
fix(webapp): Remove websockets to deploy security hot fix

### LINKED ISSUE #3082: 💥 [DevOps] Remove wsEndpoint to unblock deployment
## :fire: DevOps ticket

At the moment, we have an issue with deploying websockets due to ability to configure websockets_uri for different production environments. See https://github.com/Human-Connection/Human-Connection/issues/3064 and https://github.com/nuxt/nuxt.js/issues/5100

This is blocking our deployment of other features/bug fixes, but most importantly a security related bug fix. If we remove the `wsEndpoint`, it will not try to connect to websockets and we can work on a solution.
### Motive

see :point_up_2: 
### Related issues

see :point_up_2: 
### Implementation

comment out the line in `apollo-config.js`
### Validation

when it's deployed and does not try to connect to websockets, it should be considered successful.

mattwr18:
## 🍰 Pullrequest

Did not 
### Issues

- fixes #3082


mattwr18:
no, we don't
do you think that is acceptable for a week, or so?

mattwr18:
it might not be difficult actually @Tirokk

roschaefer:
@mattwr18 yes, shall we merge?

mattwr18:
It's failing the build... I added back the poll notifications and it was throwing a `Subscription.notificationAdded returned undefined` error

mattwr18:
Closing since @roschaefer fixed our next build/runtime env variables!!!

PR #3075, opened 2020-02-14
feat: 🍰 Expose sensitive report type to moderators only

Tirokk:
…iddleware

## 🍰 Pullrequest

Set sensible report properties to moderators only in `permissionsMiddleware`.

### Issues

- relates #3074

### Todo

- [X] set GQL permission
- [x] write tests
- [x] refactor now failing tests
- [ ] database manipulation to set `Report.disable` to the final decision, last review rules

Issue #3074, opened 2020-02-14
🔧 [Refactor] Don’t Expose Some Properties Of The Report Node In GraphQL


Tirokk:
## :zap: Refactor ticket

In some cases there is the danger that informations of the `Report` node getting exposed to the frontend. Especially the `filed` property.

### Motive

Security.

### Related issues

None.

### Implementation

- Set permissions in `permissionMiddleware`.
- Write tests for it.

### Todo

- [X] set GQL permission (done by #3075 )
- [x] write tests (done by #3075 )
- [x] refactor now failing tests (done by #3075 )
- [ ] database manipulation to set `Report.disable` to the final decision, last review rules

mattwr18:
closing this as we are no longer exposing `Report` properties to unauthorized users. 
The ToDo to write a database migration has been extracted into its own issue. See https://github.com/Human-Connection/Human-Connection/issues/3412

PR #3084, opened 2020-02-17
fix(backend): 🍰 Fix unintentional exposure of reports

Tirokk:
## 🍰 Pullrequest

Hot fix: Set sensible report properties to moderators only in permissionsMiddleware.

### Issues

- relates #3074
- Original PR #3075

### Todo

None

### DIFF
--- a/backend/src/schema/resolvers/reports.js
+++ b/backend/src/schema/resolvers/reports.js
@@ -33,8 +33,7 @@ export default {
       })
       try {
         const [reportedResource] = await reportWriteTxResultPromise
-        if (!reportedResource) return null
-        return reportedResource
+        return reportedResource || null
       } finally {
         session.close()
       }
--- a/backend/src/schema/resolvers/reports.js
+++ b/backend/src/schema/resolvers/reports.js
@@ -1,15 +1,5 @@
 import log from './helpers/databaseLogger'
 
-const transformReturnType = record => {
-  return {
-    ...record.get('report').properties,
-    resource: {
-      __typename: record.get('type'),
-      ...record.get('resource').properties,
-    },
-  }
-}
-
 export default {
   Mutation: {
     fileReport: async (_parent, params, context, _resolveInfo) => {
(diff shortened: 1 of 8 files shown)


mattwr18:
hey @Tirokk ... I tried to add back the tests, but I also refactored to return the `resource` instead of needlessly exposing the `Report`, without just cause.
I hope I am not stepping on your toes. I know that you said you had another `PR` that you are working on, or one in mind... we have been dealing with `build` issues since last `Friday`, at least, so we couldn't deploy any fix, until we get it figured out. 
I guess we can talk about this approach and see if people like it, or have other ideas and go from there. 
Again, sorry if I haven't understood what you meant, maybe you could let me know if there is any way I can improve this. I have reread your comments in `Discord` and I am still lost, to be honest.

mattwr18:
please let's discuss this in the daily tomorrow?

backend/src/schema/resolvers/reports.js
```diff
@@ -38,12 +29,12 @@ export default {
...
-        return reportTransactionResponse.records.map(transformReturnType)
+        return reportTransactionResponse.records.map(record => record.get('returnedResource'))
       })
       try {
-        const [createdRelationshipWithNestedAttributes] = await reportWriteTxResultPromise
-        if (!createdRelationshipWithNestedAttributes) return null
-        return createdRelationshipWithNestedAttributes
+        const [reportedResource] = await reportWriteTxResultPromise
+        if (!reportedResource) return null
+        return reportedResource
```
roschaefer:
```suggestion
        return reportedResource || null
```
If you want to convert `undefined` to `null`.

mattwr18: that makes sense!

Tirokk: Yes, I like this better! 😍 @roschaefer

Tirokk:
Will be solved as a whole in #3075

Tirokk:
Will be solved as a whole in #3075

PR #3090, opened 2020-02-17
fix(jwt): Whitelist encoded JWT attributes

### LINKED ISSUE #2202: 🐛 [Bug] Remove potentially sensitive information from JWT encoded token
## :bug: Bugreport
While I'm writing [the slides for my lecture](https://docs.google.com/presentation/d/1UiBKO0v0dUlCIH1bb6O6UbbH30IURjqi0D30OBieS5I/edit?usp=sharing) I see that JWT documentation warns about putting sensitive data into the token. We put the `email` of our user there. We never use the `email` when we decode the token. Let's get rid of it.

1. Get a Json Web Token: http://localhost:4000/
```graphql
mutation {
  login(email: "user@example.org", password: "1234")
}
```
2.  Copy+paste the token here https://jwt.io/
3. See the email in the base64 decoded payload

### Expected behavior
E-Mail shouldn't be there.

### Additional context
Here is where we decode the token: https://github.com/Human-Connection/Human-Connection/blob/master/backend/src/jwt/decode.js

Here is the JWT documentation: https://jwt.io/introduction/

> You also should not store sensitive session data in browser storage due to lack of security.

roschaefer:
This will prevent unintentional encoding of users email addresses in the
JWT.

@steffi201028 this might be interesting for you as well.

### Issues

- fix #2202 

### Todo

- [X] None

### DIFF
--- a/backend/src/jwt/encode.spec.js
+++ b/backend/src/jwt/encode.spec.js
@@ -1,0 +1,62 @@
+import encode from './encode'
+import jwt from 'jsonwebtoken'
+import CONFIG from './../config'
+
+describe('encode', () => {
+  let payload
+  beforeEach(() => {
+    payload = {
+      name: 'Some body',
+      slug: 'some-body',
+      id: 'some-id',
+    }
+  })
+
+  it('encodes a valided JWT bearer token', () => {
+    const token = encode(payload)
+    expect(token.split('.')).toHaveLength(3)
+    const decoded = jwt.verify(token, CONFIG.JWT_SECRET)
+    expect(decoded).toEqual({
+      name: 'Some body',
+      slug: 'some-body',
+      id: 'some-id',
+      sub: 'some-id',
+      aud: expect.any(String),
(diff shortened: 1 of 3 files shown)


roschaefer:
I don't know but maybe that might also fix our stalling builds. After all, we `encode` the entire user object in our cypress tests @mattwr18

backend/src/jwt/encode.spec.js
```diff
@@ -0,0 +1,62 @@
...
+  let payload
+  beforeEach(() => {
+    payload = {
+      name: 'Some body',
+      slug: 'some-body',
+      id: 'some-id',
+    }
+  })
+
+  it('encodes a valided JWT bearer token', () => {
```
mattwr18:
```suggestion
  it('encodes a valid JWT bearer token', () => {
```
normally, I would suggest this change, but with the build server so difficult to get passing, I think we can make this change some other time

PR #3094, opened 2020-02-18
refactor(factories): Avoid unique constraint error

roschaefer:
## 🍰 Pullrequest
If @mattwr18's assumption is right that our build server is hanging
because we run into unique constraint errors then this should always
create a unique category slug for any created category along the way.

### Issues
Intermittently failing tests.

### DIFF
--- a/backend/src/db/factories.js
+++ b/backend/src/db/factories.js
@@ -27,7 +27,9 @@ export const cleanDatabase = async (options = {}) => {
 Factory.define('category')
   .attr('id', uuid)
   .attr('icon', 'globe')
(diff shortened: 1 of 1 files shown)


mattwr18:
@roschaefer we don't have a unique constraint on `name` for category... it is required, but there is no unique constraint. There is however, on `slug`... which is needed to display the buttons in `CategoriesSelect` ... it looks to me like you are conflating the 2... since the name is `Global Peace & Nonviolence`, the slug is `global-peace-nonviolence` ...

roschaefer:
@mattwr18 is totally right, closing

PR #3110, opened 2020-02-19
chore(cypress): Remove debug statements

roschaefer:
## 🍰 Pullrequest
If that's the reason for our hanging cypress tests I kill myself (metaphorically).
### Issues

- None

### Todo

- [X] None


mattwr18:
I was able to replicate the error locally, when it stopped at the `cy.debug()`

PR #3112, opened 2020-02-19
feat(activitypub) [WIP]: First steps for activity-delete

### LINKED ISSUE #3008: 🔧 [Refactor] ActivityPub - Delete Feature
## :zap: Refactor ticket
Look at this [cucumber feature](https://github.com/Human-Connection/Human-Connection/blob/master/backend/test/features/activity-delete.feature). Apparently you can delete posts and comments ("objects") from a federated server. Probably, the signature of the actor on the 3rd party server must be verified before you can do that.

You could contribute the implementation for it and:
* Add a better feature description
* If the signature must be verified, explain this in the cucumber scenario
* Reduce the verbosity of the `Background` of the cucumber test

### Motive
Become activity pub compliant.

### Implementation
Next to our GraphQL endpoint, which should remain the primary API of the backend, we need a Rest-API as activity pub is built around REST APIs.

### Additional context
Maybe this user story will do for a valid cucumber feature?

```gherkin
Feature: Delete resource
  As a user on a 3rd party server
  I would like to delete a comment or post and this deletion should be propagated to all other servers
  Because there should be a "right of forget"
```

mattwr18:
### Issues

- fixes #3008 

### Todo

- [X] None

### DIFF
--- a/backend/src/activitypub/routes/inbox.js
+++ b/backend/src/activitypub/routes/inbox.js
@@ -1,54 +1,30 @@
 import express from 'express'
-import { activityPub } from '../ActivityPub'
+import cors from 'cors'
 
 const debug = require('debug')('ea:inbox')
 
-const router = express.Router()
-
-// Shared Inbox endpoint (federated Server)
-// For now its only able to handle Note Activities!!
-router.post('/', async function(req, res, next) {
+export async function handler(req, res) {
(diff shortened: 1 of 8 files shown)


mattwr18:
we'll pick this up after we have the ability to verify third-party servers and create resources

PR #3115, opened 2020-02-20
fix: Favor Cypress.Promise over async/await in e2e tests

mattwr18:


package.json
```diff
@@ -53,5 +55,6 @@
   },
   "resolutions": {
     "set-value": "^2.0.1"
-  }
+  },
+  "dependencies": {}
```
roschaefer:
```suggestion
```

PR #3121, opened 2020-02-20
chore(cypress): Favor firefox in cypress

mattwr18:
### Issues

- None

### Todo

- [X] None


package.json
```diff
@@ -18,8 +18,8 @@
     "cypress:backend": "cd backend && yarn run dev",
     "cypress:webapp": "cd webapp && yarn run dev",
     "cypress:setup": "run-p cypress:backend cypress:webapp",
-    "cypress:run": "cross-env cypress run",
-    "cypress:open": "cross-env cypress open",
+    "cypress:run": "cross-env cypress run --browser firefox",
+    "cypress:open": "cross-env cypress open --browser firefox",
```
roschaefer: That has the drawback that we cannot record our cypress tests anymore :cry: can we close?

cypress/integration/common/report.js
```diff
@@ -169,7 +169,7 @@ Then('each list item links to the post page', () => {
 Then('I can visit the post page', () => {
   cy.contains(annoyingUserWhoMutedModeratorTitle).click()
   cy.location('pathname').should('contain', '/post')
-    .get('h3').should('contain', annoyingUserWhoMutedModeratorTitle)
+    .get('.ds-card-content .hyphenate-text').should('contain', annoyingUserWhoMutedModeratorTitle)
```
roschaefer: Code creep?

package.json
mattwr18: No, actually it doesn't, if you look at the build, or click on cypress, you can see that all the videos were recorded

cypress/integration/common/report.js
mattwr18: Could be, yes.
It failed locally, but that was in headless. Also, I was reading cypress docs and they suggest to never use tags as selectors.
We could open another PR to clean up our cypress code though

package.json
mattwr18: I wouldn't close, is there any justifiable reason to use a different browser than what we recommend our users to use for e2e tests?

cypress/integration/common/report.js
```diff
@@ -169,7 +169,7 @@ Then('each list item links to the post page', () => {
 Then('I can visit the post page', () => {
   cy.contains(annoyingUserWhoMutedModeratorTitle).click()
   cy.location('pathname').should('contain', '/post')
-    .get('title').should('contain', annoyingUserWhoMutedModeratorTitle)
+    .get('.base-card .title').should('contain', annoyingUserWhoMutedModeratorTitle)
```
mattwr18: not related to this `PR` directly, but not even sure how this was passing!? the `title` doesn't have a `.` indicating it's a class :woman_shrugging:

cypress.json
```diff
@@ -1,5 +1,8 @@
 {
   "projectId": "qa7fe2",
   "ignoreTestFiles": "*.js",
-  "baseUrl": "http://localhost:3000"
+  "baseUrl": "http://localhost:3000",
+  "env": {
+    "RETRIES": 2
+  }
```
mattwr18: Also, not related to `firefox` as a test runner, but unless I'm missing something, does anyone think `retries` are actually occurring? hopefully, we can see some evidence that this works

PR #3130, opened 2020-02-21
[WIP] Start to implement verification for activitypub

### LINKED ISSUE #3129: 🔧 [Refactor] ActivityPub - Create Feature
## :zap: Refactor ticket
Look at this [cucumber feature](https://github.com/Human-Connection/Human-Connection/blob/master/backend/test/features/object-article.feature). You can create posts and comments ("objects") from a federated server. 
**NOTE**The signature of the actor on the 3rd party server must be verified for any `WRITE` actions.
See [verify](https://github.com/Human-Connection/Human-Connection/blob/master/backend/src/activitypub/routes/verify.js)

You could contribute the implementation for it and:
* Add a better feature description
* If the signature must be verified, explain this in the cucumber scenario

### Motive
Become activity pub compliant.

### Related issues

this is part of this `EPIC` https://github.com/Human-Connection/Human-Connection/issues/3009
### Implementation

Next to our GraphQL endpoint, which should remain the primary API of the backend, we might need a Rest-API as activity pub is built around REST APIs. 

### Additional context
Maybe this user story will do for a valid cucumber feature?

```gherkin
Feature: Delete resource
  As a user on a 3rd party server
  I would like to create a post or comment and this creation should be propagated to all other servers
  Because creating content in this way saves time and energy
```

mattwr18:
- so many questions!

### Issues

- fixes #3129 

### Todo

- [X] None


mattwr18:
still haven't been able to verify... I am understanding more I guess... 
@Mogge and @roschaefer ... I think that the `axios` request is actually to the url where the request comes in, and responds with the users `publicKey` and `publicKeyPem` ... then we can verify that the `post` request signature has not been altered.

what I am using for references, 
https://blog.joinmastodon.org/2018/06/how-to-implement-a-basic-activitypub-server/ https://blog.joinmastodon.org/2018/07/how-to-make-friends-and-verify-requests/

if this is the case, we should mock this `axios.get`, I guess.

I had a look at moxios, but it's been a while since I've used, I got to mock the entire post request, but that is not what we want.

I also found this https://gitlab.com/evanp/activitypub-mock/, which we could explore :woman_shrugging:

PR #3139, opened 2020-02-22
[WIP] feat: comment  toToast Messages  is displayed NOT scrolled anymore.

### LINKED ISSUE #3000: 💬 [Question] which possibilities would you suggest to make scrolling unnecessary?
## :speech_balloon: Question
if a user gets a lot of comments for a post then the following problem occurs. 
if you want to answer several users then it scrolls to the editor for each name. 
which possibilities would you suggest to make scrolling unnecessary? 
i see scrolling as an optical signal. we would have to find another signal which tells the user that the username is written in the answer field. 
maybe you should check if the username you want to include is the last one in the editor. 
or each reply button can only be triggered once as an event.

ogerly:
## 🍰 Pullrequest
when responding to a comment of a user a toast is displayed NOT scrolled anymore.

ezgif com-video-to-gif

### Issues
- fixes #3000

### DIFF
--- a/webapp/components/CommentCard/CommentCard.vue
+++ b/webapp/components/CommentCard/CommentCard.vue
@@ -45,7 +45,6 @@
       class="reply-button"
       circle
       size="small"
-      v-scroll-to="'.editor'"
       @click="reply"
     />
   </base-card>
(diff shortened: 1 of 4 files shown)


webapp/components/CommentCard/CommentCard.vue
```diff
@@ -149,8 +148,9 @@ export default {
       return `#${this.anchor}` === anchor
     },
     reply() {
-      const message = { slug: this.comment.author.slug, id: this.comment.author.id }
+      const message = { slug: this.comment.author.slug + ' ', id: this.comment.author.id }
       this.$emit('reply', message)
+      this.$toast.success(this.$t(`post.comment.addReply`))
```
roschaefer: I really dislike this solution. You get a success message even before a comment has been saved.

ogerly:
this will not be implemented now

Issue #2198, opened 2019-11-11
🚀 [Feature] Privacy options for posts


sushidave:
## :rocket: Feature

As a user I would like to restrict the visibility of my posts.

1. Private: Post is visible to the editing user only, can be used as draft.
2. Not listed: Post is visible to anyone, does not appear on the post feed appear and in search results. Can be used for private communication with other users and as a workaround for private messaging.

### User Problem

- Users cannot save posts as draft.
- Users cannot privately communicate with other users.

### Implementation

1. Add a privacy selection field to the post editing page.
2. Exclude such posts from the post feed and from search query.
3. Restrict access to private posts for other users.

### Design & Layout

![post-privacy-option](https://user-images.githubusercontent.com/17728384/68622479-84731800-04d2-11ea-8080-ff34a5190069.png)

DennisHack:
Super!

Tirokk:
I would put the draft as well in the `Visibility` menu, or?
@alina-beck

sushidave:
Let's have a look at how WordPress does it:

They offer a draft status and other visibility options, too. In the UI the two functions are seperated for a good reason: _Draft_ and _public_ are statuses of publication while the other ones restrict visibility. 

In the UI, WordPress shows the publication options in a fixed bar on top of the post and the visibility options in a scrollable sidebar:
![wp-ui-layout](https://user-images.githubusercontent.com/17728384/75980768-6b5c0000-5ee3-11ea-8576-9d5690f5c74e.png)

**1) Publication options:**

There are sets of publication options (both offering a preview function, too):

a) For new posts or saved drafts:

![wp-publication-options-publish](https://user-images.githubusercontent.com/17728384/75979490-e7a11400-5ee0-11ea-9975-43344b6b3e86.png)

By clicking the Publish... button a visibility selection menu appears:
![wp-publish-menu](https://user-images.githubusercontent.com/17728384/75979938-d573a580-5ee1-11ea-84ea-4876047ee280.png)


b) For published posts:

![wp-publication-options-update](https://user-images.githubusercontent.com/17728384/75979668-48305100-5ee1-11ea-8d64-1ff1ebf2d5f5.png)
 
**2) Visibility options:**

Apart from the top bar, visibility options can also be found in the post settings of the sidebar:
![wp-visibility-options-publish](https://user-images.githubusercontent.com/17728384/75980255-6054a000-5ee2-11ea-93e3-e427f06d8037.png)

[...]

PR #3169, opened 2020-02-26
feat: post visibility can be set to "public" or "registered users"

roschaefer:
### Issues

- relates #2198

### Todo

- [X] None

### DIFF
--- a/webapp/components/ContributionForm/ContributionForm.vue
+++ b/webapp/components/ContributionForm/ContributionForm.vue
@@ -70,6 +70,17 @@
         <ds-chip v-if="errors && errors.language" size="base" color="danger">
           <base-icon name="warning" />
         </ds-chip>
+        <ds-select
+          model="visibility"
+          icon="eye"
+          class="select-field"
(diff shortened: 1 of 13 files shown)


webapp/components/ContributionForm/ContributionForm.vue
```diff
@@ -70,6 +70,17 @@
...
+          model="visibility"
+          icon="eye"
+          class="select-field"
+          :options="visibilityOptions"
+          :placeholder="$t('contribution.visibilitySelectText')"
+          :label="$t('contribution.visibilitySelectLabel')"
+        />
+        <ds-chip v-if="errors && errors.visibility" size="base" color="danger">
+          <base-icon name="warning" />
+        </ds-chip>
```
mattwr18: aren't we trying to avoid adding new syleguide components?

mattwr18:
we still want this at some point, but we decided that now is not the right time to finish off the other step of this, which is to make the content available without registering/signing in. closing for now

PR #3199, opened 2020-03-03
feat: Specs For Searches

### LINKED ISSUE #2798: 🐛 [Bug] Searches with a hyphen return no results

### LINKED ISSUE #2799: 🐛 [Bug] Search input with a space ignores minimal character length

Mogge:
### 🍰 Pullrequest

This. Adding specs for searches in Backand and improve the search queries

 - fixes #2798 
 - fixes  #2799

### DIFF
--- a/backend/src/schema/resolvers/searches.js
+++ b/backend/src/schema/resolvers/searches.js
@@ -1,22 +1,18 @@
 import log from './helpers/databaseLogger'
 
+// see http://lucene.apache.org/core/8_3_1/queryparser/org/apache/lucene/queryparser/classic/package-summary.html#package.description
+
 export default {
   Query: {
     findResources: async (_parent, args, context, _resolveInfo) => {
       const { query, limit } = args
       const { id: thisUserId } = context.user
-      // see http://lucene.apache.org/core/8_3_1/queryparser/org/apache/lucene/queryparser/classic/package-summary.html#package.description
-      const myQuery = query
-        .replace(/\s+/g, ' ')
-        .replace(/[[@#:*~\\$|^\]?/"'(){}+?!,.-;]/g, '')
-        .split(' ')
-        .map(s => (s.toLowerCase().match(/^(not|and|or)$/) ? '"' + s + '"' : s + '*'))
-        .join(' ')
+
       const postCypher = `
       CALL db.index.fulltext.queryNodes('post_fulltext_search', $query)
       YIELD node as resource, score
       MATCH (resource)<-[:WROTE]-(author:User)
-      WHERE score >= 0.5
+      WHERE score >= 0.2
       AND NOT (
         author.deleted = true OR author.disabled = true
         OR resource.deleted = true OR resource.disabled = true
@@ -53,7 +49,7 @@ export default {
           thisUserId,
         })
         const userTransactionResponse = transaction.run(userCypher, {
-          query: myQuery,
+          query: createUserQuery(query),
           limit,
           thisUserId,
         })
@@ -64,6 +60,8 @@ export default {
         const [postResults, userResults] = await searchResultPromise
         log(postResults)
         log(userResults)
+        // console.log(postResults.summary.query.parameters)
+        // console.log(userResults)
         return [...postResults.records, ...userResults.records].map(r => r.get('resource'))
       } finally {
         session.close()
@@ -72,14 +70,30 @@ export default {
   },
 }
 
+function createUserQuery(str) {
+  // match the whole text
+  const normalizedString = normalizeWhitespace(str)
+  const escapedString = escapeSpecialCharacters(normalizedString)
+  const result = normalizedString.includes(' ') ? quoteString(escapedString) : escapedString
+  // console.log('"' +  + '"')
+  return result
+}
+
 function createPostQuery(str) {
   // match the whole text
-  // console.log('"' + escapeSpecialCharacters(normalizeWhitespace(str)) + '"')
-  return '"' + escapeSpecialCharacters(normalizeWhitespace(str)) + '"'
+  const normalizedString = normalizeWhitespace(str)
+  const escapedString = escapeSpecialCharacters(normalizedString)
+  const result = normalizedString.includes(' ') ? quoteString(escapedString) : escapedString
+  // console.log('"' +  + '"')
+  return result
 }
 
 function normalizeWhitespace(str) {
   return str.replace(/\s+/g, ' ')
+}
+
+function quoteString(str) {
+  return '"' + str + '"'
 }
 
 function escapeSpecialCharacters(str) {
--- a/backend/src/schema/resolvers/searches.js
+++ b/backend/src/schema/resolvers/searches.js
@@ -42,7 +42,7 @@ export default {
       LIMIT $limit
       `
       const myQuery = queryString(query)
-	
+
       const session = context.driver.session()
       const searchResultPromise = session.readTransaction(async transaction => {
         const postTransactionResponse = transaction.run(postCypher, {
--- a/backend/src/schema/resolvers/searches.js
+++ b/backend/src/schema/resolvers/searches.js
@@ -1,5 +1,5 @@
 import log from './helpers/databaseLogger'
-import queryString from './searches/queryString'
+import { queryString } from './searches/queryString'
 
 // see http://lucene.apache.org/core/8_3_1/queryparser/org/apache/lucene/queryparser/classic/package-summary.html#package.description
 
--- a/backend/src/schema/resolvers/searches.js
+++ b/backend/src/schema/resolvers/searches.js
@@ -48,7 +48,7 @@ export default {
       const session = context.driver.session()
       const searchResultPromise = session.readTransaction(async transaction => {
         const postTransactionResponse = transaction.run(postCypher, {
-          query: myQuery,
+          query: createPostQuery(query),
           limit,
           thisUserId,
         })
@@ -70,4 +70,18 @@ export default {
       }
     },
   },
+}
+
+function createPostQuery(str) {
+  // match the whole text
+  // console.log('"' + escapeSpecialCharacters(normalizeWhitespace(str)) + '"')
+  return '"' + escapeSpecialCharacters(normalizeWhitespace(str)) + '"'
+}
+
+function normalizeWhitespace(str) {
+  return str.replace(/\s+/g, ' ')
+}
+
+function escapeSpecialCharacters(str) {
+  return str.replace(/(["[\]&|\\{}+!()^~*?:/-])/g, '\\$1')
 }
--- a/backend/src/schema/resolvers/searches.js
+++ b/backend/src/schema/resolvers/searches.js
@@ -95,9 +95,8 @@ const createPostQuery = str => {
   }
   // start globbing ...
   escapedString.split(' ').forEach(s => {
-    if (!s.match(/^(AND|OR|NOT)$/i)) {
+    if (s.length > 3) {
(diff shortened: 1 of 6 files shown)


mattwr18:
this was in the list of `PR`s ready for review, but then there is a `WIP` and a comment that the tests should be continued to be refactored, so should we give a `WIP` PR review, or hold off?

backend/src/schema/resolvers/searches.spec.js
```diff
@@ -0,0 +1,452 @@
...
+        slug
+        name
+      }
+    }
+  }
+`
+let user
+
+describe('resolvers', () => {
+  describe('searches', () => {
```
mattwr18: why are there so many nested describe blocks? typically, I would see a nested describe block if they set up for those set of tests differs in some way from the others. it seems to me in this file that many of them are for the description only, is that right?

backend/src/schema/resolvers/searches.spec.js
```diff
@@ -0,0 +1,452 @@
...
+          slug: 'john-doe',
+        })
+        authenticatedUser = await user.toJson()
+      })
+
+      const factoryOptions = {
+        authorId: 'a-user',
+      }
+
+      describe('query contains first name of user', () => {
```
mattwr18: here for example...

backend/src/schema/resolvers/searches.spec.js
```diff
@@ -0,0 +1,452 @@
...
+
+        describe('query consists of words not present in the corpus', () => {
+          it('returns empty search results', async () => {
+            await expect(
+              query({ query: searchQuery, variables: { query: 'Unfug' } }),
+            ).resolves.toMatchObject({ data: { findResources: [] } })
+          })
+        })
+
+        describe('testing different post content', () => {
```
mattwr18: this is a good use of a describe block, since the set up is different. Although, it's not immediately clear to me why these posts could be created in a up the stream describe block and then tested in `it` blocks further down...
maybe you wanted to have the `build` closer to the test to make it easier to see the specifics of each post?

backend/src/schema/resolvers/searches.spec.js
```diff
@@ -0,0 +1,452 @@
...
+Ihm ist, als ob es tausend Stäbe gäbe
+und hinter tausend Stäben keine Welt.`,
+                },
+                factoryOptions,
+              ),
+            ])
+          })
+
+          describe('a post which content contains the title of the first post', () => {
+            describe('query contains the title of the first post', () => {
```
mattwr18: same as :point_up_2:

backend/src/schema/resolvers/searches.spec.js
```diff
@@ -0,0 +1,452 @@
...
+                        id: 'g-post',
+                        title: 'Zusammengesetzte Wörter',
+                        content: `Ein Bindestrich kann zwischen zwei Substantiven auch dann gesetzt werden, wenn drei gleichlautende Buchstaben aufeinandertreffen. Das ist etwa bei einem „Teeei“ der Fall, das so korrekt geschrieben ist. Möglich ist hier auch die Schreibweise mit Bindestrich: Tee-Ei.`,
+                      },
+                    ],
+                  },
+                })
+              })
+            })
+          })
```
mattwr18: I'm confused by these two test cases... is there some docs you can link to that explain why this was an issue in the past?
I went to the link you provided https://lucene.apache.org/core/8_3_1/queryparser/org/apache/lucene/queryparser/classic/package-summary.html#Escaping_Special_Characters , but I'm not sure where the German quotation marks fit into the equation.

Also looking at the tests, it's not clear. The first test searches for `tee-ei`, which the content does contain, so I would expect it to appear in the list, but maybe it wasn't because of the German quotations, somehow?

backend/src/schema/resolvers/searches.spec.js
```diff
@@ -0,0 +1,452 @@
...
+
+          describe('a post that contains a simple mathematical exprssion and linebreaks', () => {
```
mattwr18:
```suggestion
          describe('a post that contains a simple mathematical expression and line breaks', () => {
```

backend/src/schema/resolvers/searches.spec.js
```diff
@@ -0,0 +1,452 @@
...
+            describe('query a part of the mathematical expression', () => {
```
mattwr18: this is cool, and seeing the docs, I can see how this is possible. I wonder if this is an actual use case, but I guess it doesn't hurt to have a test case for it?

backend/src/schema/resolvers/searches.spec.js
```diff
@@ -0,0 +1,452 @@
...
+3. binomische Formel: (a + b)(a - b) = a² - b²`,
+                      },
+                    ],
+                  },
+                })
+              })
+            })
+          })
+
+          describe('a post that contains a poem', () => {
```
mattwr18: I don't understand where the `poem` part makes a search different :thinking:

backend/src/schema/resolvers/searches/queryString.js
```diff
@@ -0,0 +1,53 @@
...
+  return `
+${matchWholeText(escapedString)}
+${matchEachWordExactly(escapedString)}
+${matchSomeWordsExactly(escapedString)}
+${matchBeginningOfWords(escapedString)}
+`
+}
+
+const matchWholeText = (str, boost = 8) => {
+  return `"${str}"^${boost}`
```
mattwr18: thank you for giving the docs so that your reviewers can understand how the boost works!
I didn't really understand how you came to the defaults. Maybe you could explain?

backend/src/schema/resolvers/searches/queryString.spec.js
```diff
@@ -0,0 +1,43 @@
...
+  describe('special characters', () => {
+    it('does escaping correctly', () => {
+      expect(escapeSpecialCharacters('+ - && || ! ( ) { } [ ] ^ " ~ * ? : \\ / ')).toEqual(
+        '\\+ \\- \\&\\& \\|\\| \\! \\( \\) \\{ \\} \\[ \\] \\^ \\" \\~ \\* \\? \\: \\\\ \\/ ',
+      )
+    })
+  })
+
+  describe('whitespace', () => {
+    it('is normalized correctly', () => {
```
mattwr18:
```suggestion
    it('normalizes correctly', () => {
```

backend/src/schema/resolvers/searches/queryString.spec.js
```diff
@@ -0,0 +1,43 @@
...
+        '("a" AND "couple" AND "of" AND "words")^4',
+      )
+    })
+  })
+
+  describe('match at least one word exactly', () => {
+    it('boosts score by factor 2', () => {
+      expect(queryString('a couple of words')).toContain('"a"^2 "couple"^2 "of"^2 "words"^2')
+    })
+  })
```
mattwr18: I guess these tests give a hint of how the boosting works... I would infer that 8 is the maximum and you can evenly distribute that amongst words?

backend/src/schema/resolvers/searches.spec.js
```diff
@@ -0,0 +1,452 @@
...
+      describe('adding one post', () => {
+        beforeAll(async () => {
+          await Factory.build(
+            'post',
+            {
+              id: 'a-post',
+              title: 'Beitrag',
+              content: 'Ein erster Beitrag',
+            },
+            factoryOptions,
```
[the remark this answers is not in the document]
Mogge: I tried this first, and it did not work.

backend/src/schema/resolvers/searches.js
```diff
@@ -64,6 +61,8 @@ export default {
         const [postResults, userResults] = await searchResultPromise
         log(postResults)
         log(userResults)
+        // console.log(postResults.summary.query.parameters)
+        // console.log(userResults)
```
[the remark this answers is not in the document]
Mogge: Thanks. This is much better

backend/src/schema/resolvers/searches/queryString.js
```diff
@@ -0,0 +1,61 @@
...
+const matchEachWordExactly = (str, boost = 4) => {
+  if (str.includes(' ')) {
+    const tmp = str
+      .split(' ')
+      .map((s, i) => (i === 0 ? `"${s}"` : `AND "${s}"`))
+      .join(' ')
+    return `(${tmp})^${boost}`
+  } else {
+    return ''
+  }
```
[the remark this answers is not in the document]
Mogge: Thanks. This is much better

backend/src/schema/resolvers/searches.spec.js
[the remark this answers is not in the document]
Mogge: `{ authorId: user.id, },`

backend/src/schema/resolvers/searches.spec.js
Mogge: Yes. This is the way @roschaefer wanted it. At least this is what I understood that he wanted

backend/src/schema/resolvers/searches.spec.js
Mogge: I need async calls to the factory. describe cannot be used async. This would leed to a beforeAll call before each new test and therefore a even deeper nesting of describe blocks. So I decided to write all the data to the DB firsrt, and then explain each test case later.

backend/src/schema/resolvers/searches.spec.js
Mogge: The quotation marks are a testcase for unicode chars. I will put that in the description.

backend/src/schema/resolvers/searches.spec.js
Mogge: It is just for the beauty of the poem. Be happy, that I did not include the second and third verse.

backend/src/schema/resolvers/searches/queryString.js
Mogge: I thought: 1, 2, 4, 8
Next would be 16
Matching the whole query exactly -> 8
Matching all words exactly -> 4
Matching at least one word exactly -> 2
Match the beginning of some words -> 1

backend/src/schema/resolvers/searches/queryString.spec.js
Mogge: As far as I know, there is no maximum.

Mogge:
Thank you for the review, @mattwr18

mattwr18:
Thanks @Mogge !!!

PR #3207, opened 2020-03-04
perf(neo4j): Improve currentUser read performance

### LINKED ISSUE #3208: 🐛 [Bug] 504 timeout on login
## :bug: Bugreport
We had a user who cannot log in to the network since 2 weeks. I debugged the problem with her and apparently, the graphql request to login succeeds but the subsequent request to `currentUser` leads to a 504 gateway timeout. I could reproduce the problem with her JWT and Postman. I could *not* reproduce the problem with the same request via Postman but with *my* JWT. So I assume it has something to do with the implementation of `currentUser` and the problem only occurs on her account.

1. Ask Geli for her JWT
2. Open Postman and send `{ currentUser { name } }` to https://human-connection.social/api
3. Wait
4. See 504 timeout

### Expected behavior
`currentUser` should resolve.

### Version & Environment
We tested both Chrome and Firefox.

roschaefer:
## 🍰 Pullrequest
This is a
* small refactoring, we want to keep all graphql in the frontend under `graphql`
* performance improvement, because we don't query for expensive user counts and avoid N+1 queries
* hopefully a fix for our user who receives 504 timeout for `currentUser` query.

### Issues
- fixes #3208 

### Todo

- [X] None

### DIFF
--- a/backend/src/schema/types/type/User.gql
+++ b/backend/src/schema/types/type/User.gql
@@ -27,7 +27,7 @@ type User {
   id: ID!
   actorId: String
   name: String
-  email: String! @cypher(statement: "MATCH (this)-[: PRIMARY_EMAIL]->(e: EmailAddress) RETURN e.email")
+  email: String! @cypher(statement: "MATCH (this)-[:PRIMARY_EMAIL]->(e:EmailAddress) RETURN e.email")
   slug: String!
   avatar: String
   coverImg: String
@@ -38,7 +38,7 @@ type User {
   invitedBy: User @relation(name: "INVITED", direction: "IN")
   invited: [User] @relation(name: "INVITED", direction: "OUT")
 
-  location: Location @cypher(statement: "MATCH (this)-[: IS_IN]->(l: Location) RETURN l")
+  location: Location @cypher(statement: "MATCH (this)-[:IS_IN]->(l:Location) RETURN l")
   locationName: String
   about: String
   socialMedia: [SocialMedia]! @relation(name: "OWNED_BY", direction: "IN")
@@ -53,13 +53,13 @@ type User {
   showShoutsPublicly: Boolean
   locale: String
   friends: [User]! @relation(name: "FRIENDS", direction: "BOTH")
-  friendsCount: Int! @cypher(statement: "MATCH (this)<-[: FRIENDS]->(r: User) RETURN COUNT(DISTINCT r)")
+  friendsCount: Int! @cypher(statement: "MATCH (this)<-[:FRIENDS]->(r:User) RETURN COUNT(DISTINCT r)")
 
   following: [User]! @relation(name: "FOLLOWS", direction: "OUT")
-  followingCount: Int! @cypher(statement: "MATCH (this)-[: FOLLOWS]->(r: User) RETURN COUNT(DISTINCT r)")
+  followingCount: Int! @cypher(statement: "MATCH (this)-[:FOLLOWS]->(r:User) RETURN COUNT(DISTINCT r)")
 
   followedBy: [User]! @relation(name: "FOLLOWS", direction: "IN")
-  followedByCount: Int! @cypher(statement: "MATCH (this)<-[: FOLLOWS]-(r: User) RETURN COUNT(DISTINCT r)")
+  followedByCount: Int! @cypher(statement: "MATCH (this)<-[:FOLLOWS]-(r:User) RETURN COUNT(DISTINCT r)")
 
   # Is the currently logged in user following that user?
   followedByCurrentUser: Boolean! @cypher(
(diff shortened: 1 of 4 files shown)


backend/src/schema/types/type/User.gql
```diff
@@ -97,14 +97,14 @@ type User {
...
-  MATCH (this)-[: WROTE]->(r: Post)
+  MATCH (this)-[:WROTE]->(r:Post)
   WHERE NOT r.deleted = true AND NOT r.disabled = true
   RETURN COUNT(r)
   """
   )
 
   comments: [Comment]! @relation(name: "WROTE", direction: "OUT")
-  commentedCount: Int! @cypher(statement: "MATCH (this)-[: WROTE]->(: Comment)-[: COMMENTS]->(p: Post) WHERE NOT p.deleted = true AND NOT p.disabled = true RETURN COUNT(DISTINCT(p))")
+  commentedCount: Int! @cypher(statement: "MATCH (this)-[:WROTE]->(:Comment)-[:COMMENTS]->(p:Post) WHERE NOT p.deleted = true AND NOT p.disabled = true RETURN COUNT(DISTINCT(p))")
```
roschaefer: Does anybody know why we're not checking `WHERE NOT comment.deleted = true AND NOT comment.disabled = true` here?

backend/src/schema/types/type/User.gql
```diff
@@ -97,14 +97,14 @@ type User {
...
 
   shouted: [Post]! @relation(name: "SHOUTED", direction: "OUT")
   shoutedCount: Int! @cypher(statement: "MATCH (this)-[: SHOUTED]->(r: Post) WHERE NOT r.deleted = true AND NOT r.disabled = true RETURN COUNT(DISTINCT r)")
```
Mogge: Eliminate WS here too?

webapp/graphql/User.js
```diff
@@ -258,3 +258,24 @@ export const checkSlugAvailableQuery = gql`
...
+  ${userFragment}
+  query {
+    currentUser {
+      ...user
+      email
+      role
+      about
+      locationName
+      locale
+      allowEmbedIframes
```
Mogge: I miss  `contributionsCount` and `commentedCount`

backend/src/schema/types/type/User.gql
roschaefer: what is WS?

webapp/graphql/User.js
roschaefer: That's on purpose. Fetching the two will cause unnecessary N+1 queries to the database. It seems that we don't use the counts in the frontend.

backend/src/schema/types/type/User.gql
Mogge: I understand that we count the posts commented by the user and not the comments themselves. So I think the cypher statement would be more complex. We should look for all the comments by the user that are nether disabled nor deleted and then count the posts commented by these comments.

backend/src/schema/types/type/User.gql
Mogge: whitespace

Mogge: `: Shouted` instead of `:Shouted`

webapp/graphql/User.js
Mogge: And the profile page?

Mogge: see webapp/pages/profile/_id/_slug.vue line 179 and 191
`<hc-count-to slot="count" :end-val="user.contributionsCount" />`
`<hc-count-to slot="count" :end-val="user.commentedCount" />`

backend/src/schema/types/type/User.gql
Mogge: Does this work:
```cypher
MATCH (this)-[:WROTE]->(c:Comment)-[:COMMENTS]->(p:Post)
WHERE NOT p.deleted = true AND NOT p.disabled = true
WHERE NOT c.deleted = true AND NOT c.disabled = true
RETURN COUNT(DISTINCT(p))
```

webapp/graphql/User.js
roschaefer: The `user` is not coming from `store/auth` but from a separate graphql query: https://github.com/Human-Connection/Human-Connection/blob/master/webapp/pages/profile/_id/_slug.vue#L507

So that's no problem. The `currentUser` is what you see when you go to your avatar menu and update e.g. your social media.

Mogge: OK. Got it. In the profile we call the UserQuery where the counts are included. So you are right. It is not needed here.

backend/src/schema/types/type/User.gql
roschaefer: Exactly that would work. If you comment a post and disable the comment, that would first increase and decrease the `commentedCount`. Whereas without the extra condition, the `commentedCount` would stay the same.

Now that we're talking about it, I remember that we distinguish between `commentedCount` and `commentsCount`. One is the number of posts commented, the other is the number of comments written.

The `DISTINCT` in the query above will make sure that `commentedCount` does not increase if you comment the same post twice.

roschaefer: I'm going to leave this as it is though.

PR #3240, opened 2020-03-09
feat: [WIP] 🍰 Send Feedback To Reporting User (second approach)

### LINKED ISSUE #1709: 🚀 [Feature] [Moderation] Send Email to reporting Users
## :rocket: Feature

An Email should be send to an reportings user's email account,

1. when he had submitted a report,
2. when a decision is met by the moderator

### Design & Layout

- For 1. the Email should include the category of violation and the reported free from text
- For 2. the Email should include just the decision: Content was **disabled** or **not disabled** and some friendly words, but no backgrund to the decision, so that no discussion comes up.

### Additional context

Depends on the moderator beeing able to protocol his decisions.

Tirokk:
## 🍰 Pullrequest

Notify the reporting user.

### Issues

- fixes #1709

### Todo

- [x] write test for `Notification.vue` in `Notification.spec.js`
- [x] implement new notifications on `Notifications Page`
- [x] write new story notification for the filed report
- [ ] show `createdAt` and `updatedAt` on the notification?
- [x] write database manipulation script
  - [ ] rewrite the manipulation script to the new database migrations


Tirokk:
There are still open questions to me, see open checkboxes in the description: ☝🏼

- Shall the notification contain information about `updatedAt`?
  At the moment it only shows `createdAt`, but they are ordered by `updatedAt_desc`. This could seem a little bit odd for the user?
- @roschaefer suggested once that we may set a property on reported resources to check if it is already reported by the current user.
  This property could be used for disabling the `Report` menu item on the resource.
  Should we do this?
  I would say yes, but we should create a separate issue.
- My Cypher query solution for notifications in the resolver `notifications` in file `notifications.js` may not be performant enough.
  We have to test it on staging before we deploy I guess.
  If we would have Neo4j 4.x there may be a better solution with `UNION` that can be then limited and paginated.

Tirokk:
Need reviews …

PR #3251, opened 2020-03-10
feat: more prominent output of ./scripts/translations/sort.sh

rbeer:
## 🍰 Pullrequest
The output of ./scripts/translations/sort.sh tends to get overlooked in the TravisCI logs.
This change makes the "x.json is not sorted by keys" prominent red and adds a hint to the sort.sh --fix feature.

Tests obviously fail now to show the error output. I will rebase out the 'temp:' commits, if approved.

Before:
3251-travis-log-before

After:
3251-travis-log-after

Messages in bold, --fix hint in blue - which I like better as the hint is info, not a warning and it really makes it stand out:
3251-travis-log-bold_n_blue

### DIFF
--- a/scripts/translations/sort.sh
+++ b/scripts/translations/sort.sh
@@ -3,6 +3,12 @@
 ROOT_DIR=$(dirname "$0")/../..
 tmp=$(mktemp)
 exit_code=0
+errors=0
+
+TEXT_RED="\e[31m"
+TEXT_BLUE="\e[34m"
+TEXT_RESET="\e[0m"
+TEXT_BOLD="\e[1m"
 
 for locale_file in $ROOT_DIR/webapp/locales/*.json
 do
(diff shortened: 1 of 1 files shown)


Mogge:
The other problem that could happen is, that the keys in the German locale file defer from the keys in the English locale file. This is the first test done by travis. After that, it checks that the keys are sorted alphabetically in all locale files.
Running in the first issue must be fixed manually. The latter can be done by this script.
I would suggest to run `yarn locales --fix` in the `webapp` folder, which does exactly the same as the travis test, as a hint to the user how to fix the problem.

Mogge:
It looks like `enhance` is not welcome by the semantic test. Change it to `feature`

rbeer:
> I would suggest to run `yarn locales --fix` in the `webapp` folder, which does exactly the same as the travis test, as a hint to the user how to fix the problem.

The thing is that option flags are simply appended to npm-scripts. So if then someone comes along and, let's say, reverts the order to `"locales": "../scripts/translations/sort.sh && ../scripts/translations/missing-keys.sh"`, or adds something after `sort.sh` the whole thing falls flat, because `--fix` it doesn't reach `sort.sh` anymore.

webapp/locales/en.json
```diff
@@ -10,9 +10,9 @@
   },
   "admin": {
     "categories": {
-      "categoryName": "Name",
       "name": "Categories",
-      "postCount": "Posts"
+      "postCount": "Posts",
+      "categoryName": "Name"
```
roschaefer: Is that to let Travis CI fail?

roschaefer:
@rbeer can you restart the build and merge?

webapp/locales/en.json
rbeer: Yes. I'll rebase-force the branch and then the tests should pass.

PR #3257, opened 2020-03-11
refactor: FilterMenu as a first step to remove ds-flex

### LINKED ISSUE #3096: 🎨 [Styleguide Migration] Flex
## :wrench: Styleguide Migration: Flex
Remove `ds-flex` and `ds-flex-item` from the project and set `flexbox` properties directly on the nested elements.

### Additional context
part of the styleguide migration EPIC: #2074

alina-beck:
## 🍰 Pullrequest
Migrates the `ds-flex` and `ds-flex-item` component from the styleguide by refactoring them out of the project and using `CSS flexbox` instead.

### Issues
-fixes #3096 

### ToDo
- [ ] add story (and test) for `LabeledButton` component


webapp/components/_new/generic/LabeledButton/LabeledButton.story.js
```diff
@@ -0,0 +1,35 @@
...
+    }),
+    template: `
+      <labeled-button
+        icon="check"
+        :filled="filled"
+        label="All"
+        @click="filled = !filled"
+      />
+    `,
+  }))
```
alina-beck: @mattwr18 cool that you added the story! 💚 

I think, since it's interactive, we don't really need the `filled` story and could instead maybe give the button in the `default` story the label `Toggle me` to make it clear that the button is reacting to input? What do you think?

mattwr18: Great idea!

mattwr18:
> I'm happy with both the design and the code now. Loved working with you on this PR, @mattwr18! heart_eyes

yeah, it was really cool!! :grin:
let's pair more!!

webapp/components/FilterMenu/CategoriesFilter.spec.js
```diff
@@ -0,0 +1,80 @@
...
+  describe('mount', () => {
+    it('starts with all categories button active', () => {
+      const allCategoriesButton = wrapper.find('.categories-filter .sidebar .base-button')
+      expect(allCategoriesButton.attributes().class).toContain('--filled')
+    })
+
+    it('sets category button attribute `filled` when corresponding category is filtered', async () => {
+      getters['posts/filteredCategoryIds'] = jest.fn(() => ['cat9'])
+      wrapper = await Wrapper()
+      democracyAndPoliticsButton = wrapper.findAll('.categories-filter .item .base-button').at(2)
```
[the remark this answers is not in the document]
mattwr18: hey, thanks for the feedback! I agree with this, and have requested this change on other `PR`s as well, but I spent at least an hour, if not longer actually trying to get this to work, and it doesn't in this case because the button does not have a text, it has a separate `<label>`, and I couldn't workout how to filter the button based on it's sibling element.

What we could do is add a `data-test="categories-button-${category.slug}` and target the button that way. we remove `data-test` in production, so I think this would be better, what do you think?

Issue #2117, opened 2019-11-04
🚀 [Feature] Refactor deprecated slot syntax


mattwr18:
## :zap: Refactor

Following https://vuejs.org/v2/guide/components-slots.html#Named-Slots

We are using deprecated syntax in various places in our code base

In short, `<template slot="index" slot-scope="scope">` can be refactored to `<template #index="scope">`

Motivation for this change, and more examples can be found here https://github.com/vuejs/rfcs/blob/master/active-rfcs/0001-new-slot-syntax.md

mattwr18:
Let' make this into a bounty issue @Human-Connection/hc-dev-team ??

alina-beck:
Hah, yes! I just had the same idea and started adding an issue for it... Here are some more hints:

Summary of the [official Vue docs](https://vuejs.org/v2/guide/components-slots.html) and [our own guidelines](https://github.com/Human-Connection/Human-Connection/blob/bbd678caa6215865df78aff90de4a518d5d39136/webapp/vue.md#we-use-shorthands):
- define default slots as `<slot />` and named slots as `<slot name='slotName' />`
- use the shorthand `#default` or `#slotName` instead of `slot='slotName'`
- make sure the `#someSlot` attribute is only set on `<template>` elements

rbeer:
done in #3258

PR #3258, opened 2020-03-11
refactor: deprecated slot syntax [2117]

rbeer:
## 🍰 Pullrequest

> Following https://vuejs.org/v2/guide/components-slots.htmlNamed-Slots
We are using deprecated syntax in various places in our code base
In short, <template slot="index" slot-scope="scope"\> can be refactored to <template #index="scope"\>
Motivation for this change, and more examples can be found here https://github.com/vuejs/rfcs/blob/master/active-rfcs/0001-new-slot-syntax.md

### Issues

- addresses #2117 

### Todo
1. [x] use shorthand for default slots
2. [x] ~Some test failures are seemingly caused by <ds-*> not handling the new syntax well. If so, and since they are supposed to be refactored out anyway, this might be a blocker.~
Turns out, components are not the cause. Just didn't understand the new syntax, fully. ([see comment](https://github.com/Human-Connection/Human-Connection/pull/3258#issuecomment-598454126))

### DIFF
--- a/webapp/components/CommentForm/CommentForm.vue
+++ b/webapp/components/CommentForm/CommentForm.vue
@@ -1,6 +1,6 @@
 <template>
   <ds-form v-model="form" @submit="handleSubmit" class="comment-form">
-    <template slot-scope="{ errors }">
+    <template #default="{ errors }">
       <base-card>
         <hc-editor ref="editor" :users="users" :value="form.content" @input="updateEditorContent" />
         <div class="buttons">
--- a/webapp/components/CommentForm/CommentForm.vue
+++ b/webapp/components/CommentForm/CommentForm.vue
@@ -1,6 +1,6 @@
 <template>
   <ds-form v-model="form" @submit="handleSubmit" class="comment-form">
-    <template v-slot="{ errors }">
+    <template slot-scope="{ errors }">
       <base-card>
         <hc-editor ref="editor" :users="users" :value="form.content" @input="updateEditorContent" />
         <div class="buttons">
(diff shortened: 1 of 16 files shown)


rbeer:
Looks like the new Syntax is truly breaking.

tl;dr/ The new syntax requires everything to be wrapped in <template>s, when using named scopes.

From the Vue.js docs, [Compilation Scope](https://vuejs.org/v2/guide/components-slots.html#Compilation-Scope):
>As a rule, remember that:
>Everything in the parent template is compiled in parent scope; everything in the child template is compiled in the child scope.

Further down, in the section [Named Scope](https://vuejs.org/v2/guide/components-slots.html#Named-Slots) it moreover says:
>Note that v-slot can only be added to a <template\> (with one exception), unlike the deprecated slot attribute.

So I was actually wrong, @mattwr18, when I said that `v-slot` wouldn't work only on _native_ tags and usage on components being fine.

Chaning
```js
<ds-menu :routes="routes">
    <ds-menu-item
      #menuitem="item"
      :route="item.route"
      :parents="item.parents"
      @click.stop.prevent="openItem(item.route, toggleMenu)"
    >
    <base-icon :name="item.route.icon" />
      {{ item.route.label }}
    </ds-menu-item>
</ds-menu>
```
to
```js
<ds-menu :routes="routes">
  <template #menuitem="item">
    <ds-menu-item
      :route="item.route"
      :parents="item.parents"
      @click.stop.prevent="openItem(item.route, toggleMenu)"
    >
    <base-icon :name="item.route.icon" />
      {{ item.route.label }}
    </ds-menu-item>
  </template>
</ds-menu>
```

makes the tests for []() pass; explaining why `item` is undefined, thus throwing `TypeError: Cannot read property 'route' of undefined`.

The question remains whether this is a good change, looking at the fact that all these <ds-*> components are about to be refactored anyways.

rbeer:
thihi :relaxed: :tada:

PR #3262, opened 2020-03-12
feat(backend): upload original image files on S3 object storage

### LINKED ISSUE #3250: 🚀 [Feature] Upload image files to S3 compatible object storage
## :rocket: Feature
We would like to store image file on Digital Spaces or any other S3 compatible object storage.

### User Problem
Currently we save images in a kubernetes volume. This creates issues:
* We cannot scale the backend on Digital Ocean as they don't support multi-read-write access
* It's more expensive than a simple object storage
* It creates unnecessary load on our backend 

### Implementation
We need to create a migration which bulk uploads the image files which can be referenced by URLs saved in our database (user avatars, post hero images). The image URLs in the database need to be updated accordingly.

### Validation
A request for an image file should not hit the kubernetes cluster anymore.

### Additional context
https://www.digitalocean.com/products/spaces/

roschaefer:
### Issues

- close #3250 

### Todo

- [X] None


mattwr18:
thanks @Mogge ... I'm testing out the check to see if the file exists... I'll ping you, if/when I get it working

mattwr18:
Leaving some things that I've discovered while looking into this here.

1) there is a method `headObject` that can be called to attempt to grab the `HEAD` of an object in a bucket, if it's doesn't exist it should throw an `error.code === 'Not Found'`, then we know to upload an image. Testing this out yesterday, it didn't throw this error when there was now image in our `image-upload` bucket in `DO Spaces`... :woman_shrugging: 

2) It would be nice to use `sync` to do this. There exists `s3_sync` for `ansible` that does this, which we could, in theory, use to upload a local copy of the `public/uploads` directory to `DO Spaces`, and maybe write out a `csv` with the `original filename/spacesUrl` and run a migration just to update the `image.url`s.

3) there exists a `sync` command for the aws-cli, which also handles this.

4) there does not seem to be the equivalent for `aws-sdk-js`. See https://github.com/aws/aws-sdk-js/issues/1159

backend/src/db/migrations/20200312140328-bulk_upload_to_s3.js
```diff
@@ -32,39 +32,38 @@ export async function up(next) {
...
             const fileLocation = path.join(__dirname, `../../../public/${pathname}`)
             const s3Location = `original${pathname}`
-            const mimeType = mime.lookup(fileLocation)
+            if (existsSync(fileLocation)) {
+              const mimeType = mime.lookup(fileLocation)
+              const params = {
+                Bucket,
+                Key: s3Location,
+                ACL: 'public-read',
+                ContentType: mimeType || 'image/jpeg',
```
roschaefer: `mimeType` or whatever :laughing:

backend/src/db/migrations/20200312140328-bulk_upload_to_s3.js
```diff
@@ -32,39 +32,38 @@ export async function up(next) {
...
+    let urls = records.map((r) => r.get('url'))
+    urls = urls.filter((url) => url.startsWith('/uploads'))
     const locations = await Promise.all(
       urls
-        .map(url => {
+        .map((url) => {
           return async () => {
             const { pathname } = new URL(url, 'http://example.org')
             const fileLocation = path.join(__dirname, `../../../public/${pathname}`)
             const s3Location = `original${pathname}`
```
roschaefer: I see that the images now live on our community staging. Do we want to have them in a folder `uploads/...`? I would say, we should omit that folder name, no?

roschaefer: Or, maybe not. We can clearly distinguish the images that have been uploaded with this migration. I case we have to do some disaster recovery, this could come in handy.

backend/src/db/migrations/20200312140328-bulk_upload_to_s3.js
```diff
@@ -32,39 +32,38 @@ export async function up(next) {
...
-            const mimeType = mime.lookup(fileLocation)
+            if (existsSync(fileLocation)) {
```
roschaefer: Can we maybe do sth. in `else`? E.g. delete the `url` from the Image node? Shall we do that in another migration?

backend/src/db/migrations/20200312140328-bulk_upload_to_s3.js
mattwr18: yeah, `ContentType` is not required, so maybe it doesn't make sense to have it at all?

backend/src/db/migrations/20200312140328-bulk_upload_to_s3.js
mattwr18: the reason it lives there is because it's part of the `Image.url`
yeah, agree with the second part of your message, let's KISS

backend/src/db/migrations/20200312140328-bulk_upload_to_s3.js
mattwr18: I thought about that, like we shouldn't have any `Image` nodes that point to files that don't exist in the `fileSystem`... like maybe we could have done that in the migration that created the `Image` nodes, but since that has already been deployed, we could iterate through all `Image`, check if they exist  in the file system and if not delete them and their relationship to `User/Post`, oder?

PR #3271, opened 2020-03-14
[WIP] #2144 add search results page

### LINKED ISSUE #2144: 🚀 [Feature] Add Search Results Page
## :rocket: Feature
Users should see search results on a separate search results page.

### User Problem
At the moment users only get very limited results in a search results dropdown. Viewing the results on a separate page would give a better overview and allow for a larger number of results.

### Implementation
- There should be a separate route for search results.
- All search results for each data type (`posts`, `users`, `hashtags`) should be fetched when the user confirms their input by hitting `return` or clicking the `🔍` button.
- The search term should be displayed on top of the page (we need a design for that).
- The posts on the page should look similar to the ones on the index page.
- We need a design + component for `users`.
- We need a design + component for `hashtags` (they could maybe be grouped on one card?).
- Users should be able to clear the search (and get back to the index page?), e.g. by clicking a `x` button next to the search term.

### Design & Layout
Missing! :(

### Validation
- users visiting the search page

### Additional context
This is **Step 4** in our search implementation – see #???.
**Step 1** is adding users to the list of search results: #1463.
**Step 2** is limiting the number of search results to 15: #2142.
**Step 3** is adding hashtags to the search results: #2140.

ogerly:
## 🍰 Pullrequest
at the event "onEnter" a new page is called and the search results are retrieved. 

### Issues
- fixes #2144 
 
### Todo
[X]  create new page
[X]  link with onEnter Event
[X]  search query on new page
[X]  customize specs in SearchableInput.spec.js
[X]  display of the data on search-results.vue
[ ]  style of the data on search-results.vue
[ ]  create test page search-results..spec.js
[ ]  write test page on search-results..spec.js

last Stand:

ezgif com-video-to-gif (5)


ogerly:
ezgif com-video-to-gif (8)

ogerly:
ezgif com-optimize

mattwr18:
don't know who will work on this next, but in case I don't get a chance to talk with them, I made some pretty big changes in the course of paginating the search results page... I tried to explain the rationale behind the changes in the commit message. Open for discussion, if there's a better idea how to do it.
Also, there maybe is a way to clean up the "duplicate" methods/data attributes. Didn't really have much time after I got it working to check it out.
I also added some `dummy seed data` that allows us to search for `Jenny` and get many results back for `posts` and `users`...

mattwr18:
Also, I spent a good amount of time on `close search results dropdown onEnter`... and it seems like it would be straightforward, but it's not really. Or I'm missing something. The issue is that the `open` is handled in two ways, by the styleguide, where `ds-select` comes from - either by `v-click-outside` or by changing `isOpen`, but we can't emit an event to change this :angry: 
I would love to migrate it @alina-beck , but I guess it's not so straightforward.

alina-beck:
Great that you got the pagination working @mattwr18! :tada:

I also had a quick look at the `close onEnter` issue and came to the same conclusion: it's not at all straightforward. I was hoping I had missed something but if you couldn't find a solution then I guess there is none, with the current styleguide-input-thingy... So, let's leave it out of this `PR` and open a new issue to fix this (after migrating the `input` component from the styleguide)?

ogerly:
ezgif com-video-to-gif

backend/src/schema/resolvers/searches.js
```diff
@@ -3,90 +3,200 @@ import { queryString } from './searches/queryString'
...
+const postWhereClause = `WHERE score >= 0.0
+  AND NOT (
+    author.deleted = true OR author.disabled = true
+    OR resource.deleted = true OR resource.disabled = true
+    OR (:User {id: $userId})-[:MUTED]->(author)
+  )`
+
+const searchPostsSetup = {
+  fulltextIndex: 'post_fulltext_search',
+  match: 'MATCH (resource)<-[:WROTE]-(author:User)',
```
Tirokk: I guess the type is already defined by `fulltextIndex: 'post_fulltext_search' `.

I would prefer to have the type on the `resource ` for clearness and better readability:
`match: 'MATCH (resource:Post)<-[:WROTE]-(author:User)‘,`

backend/src/schema/resolvers/searches.js
```diff
@@ -3,90 +3,200 @@ import { queryString } from './searches/queryString'
...
+    author: properties(author),
+    commentsCount: toString(size(comments)),
+    shoutedCount: toString(size(shouter))
+  }`,
+  limit: 'LIMIT $limit',
+}
+
+const searchUsersSetup = {
+  fulltextIndex: 'user_fulltext_search',
+  match: 'MATCH (resource)',
```
Tirokk: I guess the type is already defined by `fulltextIndex: 'user_fulltext_search' `.

Same here:
`match: 'MATCH (resource:User)',`

backend/src/schema/resolvers/searches.js
```diff
@@ -3,90 +3,200 @@ import { queryString } from './searches/queryString'
...
+  whereClause: simpleWhereClause,
+  withClause: '',
+  returnClause: 'resource {.*, __typename: labels(resource)[0]}',
+  limit: 'LIMIT $limit',
+}
+
+const searchHashtagsSetup = {
+  fulltextIndex: 'tag_fulltext_search',
+  match: 'MATCH (resource)',
```
Tirokk: I guess the type is already defined by `fulltextIndex: 'user_fulltext_search' `.

Same here:
`match: 'MATCH (resource:Tag)',`

backend/src/db/seed.js
```diff
@@ -928,16 +928,25 @@ const languages = ['de', 'en', 'es', 'fr', 'it', 'pt', 'pl']
...
     ])
 
-    const additionalUsers = await Promise.all(
-      [...Array(30).keys()].map(() => Factory.build('user')),
+    await Promise.all([...Array(30).keys()].map(() => Factory.build('user')))
+    await Promise.all(
+      [...Array(30).keys()].map((index) => Factory.build('user', { name: `Jenny${index}` })),
     )
     await Promise.all(
-      additionalUsers.map(async (user) => {
```
Mogge, line 935: Is there any reason to eliminate the relations?

backend/src/schema/resolvers/searches.spec.js
```diff
@@ -24,12 +24,12 @@ beforeAll(async () => {
 })
 
 afterAll(async () => {
-  await cleanDatabase()
+  //  await cleanDatabase()
```
Mogge: Comment must be removed.

webapp/components/_new/features/SearchResults/SearchResults.story.js
```diff
@@ -0,0 +1,148 @@
...
+    createdAt: faker.date.past(),
+    updatedAt: faker.date.recent(),
+    deleted: false,
+    disabled: false,
+    typename: 'User',
+    ...fields,
+  }
+}
+
+const posts = [
```
Mogge, line 40: Use the helpers in `~/storybook/helpers.js`?
`fakeUser(n)`
`fakePost(n)`

webapp/components/_new/generic/PaginationButtons/PaginationButtons.vue
```diff
@@ -1,32 +1,56 @@
 <template>
   <div class="pagination-buttons">
     <base-button
+      v-if="hasPreviousResult"
       @click="$emit('back')"
-      :disabled="!hasPrevious"
+      :disabled="!hasPreviousResult"
```
Mogge: As we use `v-if` disabled is not necessary

webapp/components/_new/generic/PaginationButtons/PaginationButtons.vue
```diff
@@ -1,32 +1,56 @@
...
+    <span class="pagination-pageCount">
+      {{ $t('search.page') }} {{ activePage + 1 }} /
+      {{ Math.floor((activeResourceCount - 1) / pageSize) + 1 }}
+    </span>
+
     <base-button
+      v-if="hasMoreResults"
       @click="$emit('next')"
-      :disabled="!hasNext"
+      :disabled="!hasMoreResults"
```
Mogge: see above

webapp/components/generic/SearchableInput/SearchableInput.spec.js
```diff
@@ -117,5 +110,15 @@ describe('SearchableInput.vue', () => {
         expect(mocks.$router.push).toHaveBeenCalledWith('?hashtag=Hashtag')
       })
     })
+
+    it('searches for the term when enter is pressed', async () => {
```
Mogge: Please update the text. It is the old one and very misleading

webapp/graphql/Search.js
```diff
@@ -1,15 +1,17 @@
 import gql from 'graphql-tag'
-import { userFragment, postFragment } from './Fragments'
+import { userFragment, postFragment, tagsCategoriesAndPinnedFragment } from './Fragments'
```
Mogge, line 2: Why do we need the tagsCategoriesAndPinnedFragment here?
It leeds the cypress tests to fail and I cannot see any reason why we would need this information for the search results.

backend/src/db/seed.js
Tirokk: You talking about the `following` relation @Mogge ?
But here nothing has changed. Or I’m mistaken?

Tirokk: Okay it seems like this is an old comment and the relations were removed in an earlier commit GitHub blame told me.

backend/src/db/factories.js
```diff
@@ -64,7 +64,7 @@ Factory.define('basicUser')
     password: '1234',
     role: 'user',
     about: faker.lorem.paragraph,
-    termsAndConditionsAgreedVersion: '0.0.1',
+    termsAndConditionsAgreedVersion: '0.0.4',
```
Tirokk: I really understand why you are doing this …
Avoiding the nerv breaking agreeing … 

In this case we developer do kind of a live testing all the time: When the database is fresh we have to agree to the terms and conditions.
But as far as I can see we have no frontend test for this behaviour.

Because this very important behaviour should be tested I can only agree to this change if we have written a frontend or e2e (Cypress) test.

On the other hand, what I like about that the developer has always to agree to the terms after seeding the database:
We always are aware of that the database is fresh when we have to agree! 🍀

backend/src/schema/resolvers/searches.js
```diff
@@ -3,90 +3,200 @@ import { queryString } from './searches/queryString'
...
+          limit: firstHashtags,
+        }),
       }
     },
+    searchResults: async (_parent, args, context, _resolveInfo) => {
+      const { query, limit } = args
+      const { id: userId } = context.user
+
+      const searchType = query.replace(/^([!@#]?).*$/, '$1')
+      const searchString = query.replace(/^([!@#])/, '')
```
Tirokk, line 181: I would really like to have it commented what both regex' do in your opinion. 🤗
For me regex' are not self explaining …

ogerly:
nice :)

backend/src/schema/resolvers/searches.js
```diff
@@ -3,90 +3,200 @@ import { queryString } from './searches/queryString'
...
+      if (searchType === '')
+        return [
+          ...(await getSearchResults(context, searchPostsSetup, params)),
+          ...(await getSearchResults(context, searchUsersSetup, params)),
+          ...(await getSearchResults(context, searchHashtagsSetup, params)),
+        ]
+
+      params.limit = 15
+      const type = multiSearchMap.find((obj) => obj.symbol === searchType)
+      return getSearchResults(context, type.setup, params)
```
Tirokk, line 199: This way only the first found search class character of `['!', '@', '#']` in the search string rules the result.
The others play no rule!

For a discussion.
May be changed in an additional PR …

This I would do different, but the hierarchy is okay.
I would suggest:

- `'!’` is included:
  - Search the text after the `'!’` (before you hit the next search class character) in the title and content of the post.
  - If there is `'@‘`or `'#’` as well, then only show posts which have this author and this hashtag as well.
  - Handle author and hashtag tabs like `'!’` would be not included, see the following points …
- `'!’` is not(!) included:
  - `’@'` is included
    - Post tab shows only posts by this author and may have the hashtag as well, if present.
  - `’@'` is not(!) included, but `'#’` is:
    - Post tab shows only posts with hashtag.
    - Authors tab shows authors used this hashtag.

Additional:

- Have the search class characters as separators: Search under their label all the text after them before the next appears.

backend/src/schema/resolvers/searches.spec.js
```diff
@@ -416,6 +431,128 @@ und hinter tausend Stäben keine Welt.`,
         })
       })
 
+      describe('adding a user and a hashtag with a name that is content of a post', () => {
```
Tirokk, line 434: Hey, thanks for all the new tests here! 😍

backend/src/schema/resolvers/searches.spec.js
```diff
@@ -477,6 +614,30 @@ und hinter tausend Stäben keine Welt.`,
           })
         })
       })
+
+      describe('searchPostQuery', () => {
```
Tirokk, line 618: Thanks again for the new test! 😍

cypress/integration/common/search.js
```diff
@@ -85,6 +85,26 @@ Then(
   }
 );
 
+Then("I should see the search results page", () => {
```
Tirokk, line 88: Nice to have e2e tests! 😍

backend/src/db/factories.js
Tirokk: See my comment above ☝🏼

Text above:

I really understand why you are doing this …
Avoiding the nerv breaking agreeing …

In this case we developer do kind of a live testing all the time: When the database is fresh we have to agree to the terms and conditions.
But as far as I can see we have no frontend test for this behaviour.

Because this very important behaviour should be tested I can only agree to this change if we have written a frontend or e2e (Cypress) test.

On the other hand, what I like about that the developer has always to agree to the terms after seeding the database:
We always are aware of that the database is fresh when we have to agree! 🍀

backend/src/schema/resolvers/searches.js
Tirokk: See my comment above ☝🏼

backend/src/schema/resolvers/searches.js
Tirokk: See my comment above ☝🏼

Should we make an issue out of my suggestions there?
Or you think it’s not important? @Mogge

webapp/components/_new/features/SearchResults/SearchResults.spec.js
```diff
@@ -0,0 +1,219 @@
...
+  const Wrapper = () => {
+    const store = new Vuex.Store({
+      getters,
+    })
+    return mount(SearchResults, { mocks, localVue, propsData, store })
+  }
+
+  beforeEach(() => {
+    mocks = {
+      $t: jest.fn(),
```
Tirokk, line 26:
```suggestion
      $t: jest.fn((a) => a),
```

I suggest this one as a standard for testing, because then the translations are not empty and can be identified by their identifiers as the content.

webapp/components/_new/features/SearchResults/SearchResults.story.js
Tirokk: See @Mogge comment above ☝🏼

This seems to be still valid. Or?

@mogge said:

Use the helpers in `~/storybook/helpers.js`?
`fakeUser(n)`
`fakePost(n)`

ogerly:
FiN

backend/src/schema/resolvers/searches.js
ogerly: /
^([!@#]?).*$
/
gm
^ asserts position at start of a line
1st Capturing Group ([!@#]?)
Match a single character present in the list below [!@#]?
? Quantifier — Matches between zero and one times, as many times as possible, giving back as needed (greedy)
!@# matches a single character in the list !@# (case sensitive)
.* matches any character (except for line terminators)
* Quantifier — Matches between zero and unlimited times, as many times as possible, giving back as needed (greedy)
$ asserts position at the end of a line
Global pattern flags
g modifier: global. All matches (don't return after first match)
m modifier: multi line. Causes ^ and $ to match the begin/end of each line (not only begin/end of string)

backend/src/schema/resolvers/searches.js
ogerly: @Mogge. i make a new issues for @Tirokk  suggestion?

webapp/components/_new/features/SearchResults/SearchResults.story.js
```diff
@@ -0,0 +1,148 @@
+import { storiesOf } from '@storybook/vue'
```
Tirokk, line 1: Unfortunately `yarn storybook` is not working at my machine at the moment.

What’s about you?

Tirokk:
There're two errors in the in the backend tests @Mogge .

webapp/components/_new/features/SearchResults/SearchResults.vue
```diff
@@ -0,0 +1,328 @@
...
+        <pagination-buttons
+          v-if="activeResourceCount > pageSize"
+          :hasMoreResults="hasMoreResults"
+          :hasPreviousResult="hasPreviousResult"
+          :activePage="activePage"
+          :activeResourceCount="activeResourceCount"
+          @back="previousResults"
+          @next="nextResults"
+          :key="'Top'"
+          :pageSize="pageSize"
```
Tirokk: We had the format rule to put the events at last in the properties list?! 😉

webapp/components/_new/features/SearchResults/SearchResults.vue
```diff
@@ -0,0 +1,328 @@
...
+        </masonry-grid>
+
+        <ul v-if="activeTab === 'User'" class="user-list">
+          <li v-for="resource in activeResources" :key="resource.key" class="item">
+            <base-card :wideContent="true">
+              <user-teaser :user="resource" />
+            </base-card>
+          </li>
+        </ul>
+        <ul v-if="activeTab === 'Hashtag'" class="hashtag-list">
```
Tirokk: Empty line between this like above? ☝🏼

webapp/components/_new/features/SearchResults/SearchResults.vue
```diff
@@ -0,0 +1,328 @@
+<template>
+  <div id="search-results" class="search-results">
+    <div>
+      <ds-section>
+        <ds-text class="total-search-results">
+          <strong>{{ searchCount }}</strong>
+          {{ $t('search.results', {}, searchCount) }}
+        </ds-text>
```
Tirokk: Put the search string here a well? 🤔

I had situations where the the search string in the search field was not the same as for the results found. Happened as I went back `←` an forth `→` in the browser.

The text above the tab `X results found` is a bit technical in design as well with the frame around it. Is this even needed if we have the counts of the separate found types in the tabs?
But may be for the search string like I stated above.

webapp/components/_new/features/SearchResults/SearchResults.vue
```diff
@@ -0,0 +1,328 @@
...
+        <pagination-buttons
+          v-if="activeResourceCount > pageSize"
+          :hasMoreResults="hasMoreResults"
+          :hasPreviousResult="hasPreviousResult"
+          :activePage="activePage"
+          :activeResourceCount="activeResourceCount"
+          @back="previousResults"
+          @next="nextResults"
+          :key="'Bottom'"
+          :pageSize="pageSize"
```
Tirokk: Same same ☝🏼

webapp/components/_new/generic/TabNavigation/TabNavigation.vue
```diff
@@ -0,0 +1,65 @@
+<template>
```
Tirokk, line 1: Cool, that you implement a tab component! 😍

This component looks a bit technical. 😉

We use a kind of similar thing in the user profile to switch between `posts`, `commented`, and `shouted`. That would be a nice design. Or? @Mogge

Tirokk:
I fixed the backend tests @mogge , but the webapp test for the `PaginationButtons.vue` is still failing.

webapp/components/generic/SearchableInput/SearchableInput.vue
```diff
@@ -107,15 +107,12 @@ export default {
...
-      clearTimeout(this.searchProcess)
-      if (!this.loading) {
-        this.previousSearchTerm = this.unprocessedSearchInput
-        this.$emit('query', this.unprocessedSearchInput)
-      }
+      this.$router.push({
+        path: '/search/search-results',
+        query: { search: this.unprocessedSearchInput },
+      })
+      this.$emit('clearSearch')
```
Tirokk, line 115: If you hit enter, it clears the menu under the search field. But the field is not empty. The menu says then `Nothing found`. That’s a bit misleading, because the results are just underneath on the search page.

webapp/locales/de.json
```diff
@@ -590,12 +590,15 @@
...
+      "Post": "Beitrag ::: Beiträge",
+      "Tag": "Hashtag ::: Hashtags",
+      "User": "Benutzer ::: Benutzer"
     },
     "hint": "Wonach suchst Du?",
-    "placeholder": "Suchen"
+    "no-results": "Keine Ergebnisse für \"{search}\" gefunden. Versuch' es mit einem anderen Begriff!",
+    "page": "Seite",
+    "placeholder": "Suchen",
+    "results": "Ergebniss gefunden ::: Ergebnisse gefunden"
```
Tirokk: `Ergebnis` would be correct. 😝

webapp/pages/notifications/index.spec.js
```diff
@@ -90,7 +90,7 @@ describe('PostIndex', () => {
         expect(wrapper.vm.selected).toEqual(propsData.filterOptions[1].label)
       })
 
-      it('refreshes the notificaitons', () => {
+      it('refreshes the notifications', () => {
```
Tirokk, line 93: 😍

webapp/pages/search/search-results.vue
```diff
@@ -0,0 +1,24 @@
...
+
+<script>
+import SearchResults from '~/components/_new/features/SearchResults/SearchResults'
+
+export default {
+  layout: 'default',
+  watchQuery: ['search'],
+  head() {
+    return {
+      title: 'SearchResults',
```
Tirokk: `SearchResults` without a space and not translated … 😉

webapp/pages/search/search-results.vue
```diff
@@ -0,0 +1,24 @@
+<template>
+  <search-results :search="search" />
+</template>
+
+<script>
+import SearchResults from '~/components/_new/features/SearchResults/SearchResults'
+
+export default {
+  layout: 'default',
+  watchQuery: ['search'],
```
Tirokk, line 10: Cool! 😍

Tirokk (changes requested):
Hey @oger and @Mogge ,

really cool feature!!! 😍🚀💫

I have some suggestions for changes and some questions.
But you are close!!! 👍🏼

The webapp tests are failing with in `webapp/components/_new/generic/PaginationButtons/PaginationButtons.spec.js`.

ogerly:
project closed

PR #3286, opened 2020-03-17
feat(WIP): Filtering Searches

Mogge:
## 🍰 Pullrequest
The filters set by the users, e.g. category or language, should be applied to the search results


Mogge:
Feature is not needed

PR #3293, opened 2020-03-18
feat: FollowerList [1741]

### LINKED ISSUE #1741: 🐛 [Bug] Follower/Following Lists are not expandable
## :bug: Bugreport
If somebody has a lot of followers or is following more people, just a few of them is shown in the profile. There is no way to expand the list with a click or to open a new window where all followers are shown.
contacts-do-not-expand

1. Go into your profile
2. See the followers on the left
3. have no chance to see all followers

### Expected behavior
I would espect to click below the list and have the list expanded in place. Another solution could be to let it contain every follower from scratch, but the first approach may be better.

### Version & Environment
 Type: [Desktop]      
 - OS: [Linux|Windows]      
 - Browser: [Firefox] 
 - Version [69]  
 - Device: [Notebook]  

### Additional context
There should be a general conecpt about all expanding or not expanding lists like this.

Because of #2479 no lists should be displayed to the public at all. Though, the single user should see (only) whome he is following.

### See additional informations in the dialog below

rbeer:
## 🍰 Pullrequest
This
- extracts follower/-edBy lists from _id/_slug page into own feature component FollowerList
- turns info text about number of more connections into button
- adds fetching of all remaining connections
- adds filtering of the list when remaining connections have been fetched

### Issues
- fixes #1741

### Todo

- [x] Turn list of <user-teaser\>s into overflow container
- [x] Add filter <input\> and logic
- [ ] integration tests


webapp/components/features/FollowList.spec.js
```diff
@@ -0,0 +1,228 @@
...
+              localVue,
+            })
+          })
+
+          it(`shows the users ${type}`, () => {
+            expect(wrapper.findAll('.user-teaser').length).toEqual(user[type].length)
+          })
+          it(`has a button to load all remaining users ${type}`, async () => {
+            wrapper.find('button').trigger('click')
+            //await wrapper.vm.$nextTick()
```
mattwr18:
```suggestion
            await Vue.nextTick()
```
you must `import Vue from 'vue'` then run this :point_up_2:

rbeer: It's available both ways.
It's literally just switching from node <12 to node >=12 that makes the difference. I can reproduce that locally.

mattwr18: ok, I haven't seen it used the way you added it, we have it `34` different places as `Vue.nextTick()` and as you noted, we use node version `lts`, which is `12`, so it doesn't make sense that it would work in all the other places and not for you.

Unfortunately, I am dealing with issues locally, but I will try to fix them and pull it down locally.

rbeer: Same with `Vue.nextTick`. What really bugs me is that I couldn't even reproduce that with a clean Vue project.

passing-at-node-11
failing-at-node-12

rbeer: fixed in 0b91601

webapp/components/features/FollowList/FollowList.vue
```diff
@@ -0,0 +1,157 @@
+<template>
+  <base-card class="hc-connections">
```
mattwr18:
```suggestion
  <base-card class="follow-list">
```
we give the top level `class` in our components the same name as the component.
we should probably document this, but we can do that in another `PR`

webapp/components/features/FollowList/FollowList.vue
```diff
@@ -0,0 +1,157 @@
...
+    <ds-space v-if="this.connections && this.connections.length" margin="x-small">
```
mattwr18:
```suggestion
    <ds-space v-if="connections && connections.length" margin="x-small">
```
don't know how this isn't throwing an error, but you don't have access to this in the `html`

webapp/components/features/FollowList/FollowList.vue
```diff
@@ -0,0 +1,157 @@
...
+      <ds-text tag="h5" color="soft">
+        {{ userName | truncate(15) }} {{ $t(`profile.network.${type}`) }}
+      </ds-text>
+    </ds-space>
+    <template v-else>
+      <p class="no-connections">{{ userName }} {{ $t(`profile.network.${type}Nobody`) }}</p>
+    </template>
+    <template v-if="this.connections && this.connections.length <= 7">
```
mattwr18: same as :point_up_2:

webapp/components/features/FollowList/FollowList.vue
```diff
@@ -0,0 +1,157 @@
...
+      <ds-space v-for="follow in uniq(this.connections)" :key="follow.id" margin="x-small">
```
mattwr18: same as :point_up_2:

webapp/components/features/FollowList/FollowList.vue
```diff
@@ -0,0 +1,157 @@
...
+        <!-- TODO: find better solution for rendering errors -->
+        <client-only>
+          <user-teaser :user="follow" />
+        </client-only>
+      </ds-space>
+      <ds-space v-if="this.allConnectionsCount - this.connections.length" margin="small">
```
mattwr18: same as :point_up_2: 
instead of writing this for every case, it's true for every `this` in the `html`

webapp/components/features/FollowList/FollowList.vue
```diff
@@ -0,0 +1,157 @@
...
+          v-for="follow in uniq(this.filteredConnections)"
+          :key="follow.id"
+          margin="x-small"
+        >
+          <client-only>
+            <user-teaser :user="follow" />
+          </client-only>
+        </ds-space>
+      </div>
+      <ds-space margin="x-small">
```
mattwr18: we are trying to refactor out all uses of `ds-space`, would you mind helping us with that?

webapp/components/features/FollowList/FollowList.vue
```diff
@@ -0,0 +1,157 @@
...
+
+          return {
+            user,
+            score: match ? match[0].length * (match.index + 1) : -1,
+          }
+        })
+        .filter((score) => score.score !== -1)
+        .sort((a, b) => a.score - b.score)
+
+      return fuzzyScores.map((score) => score.user)
```
mattwr18: this looks complex, I'll have to sit down and think about what this does, and if there's not a way to simplify it

webapp/components/features/FollowList/FollowList.vue
```diff
@@ -0,0 +1,157 @@
...
+      <ds-space v-for="follow in uniq(this.connections)" :key="follow.id" margin="x-small">
+        <!-- TODO: find better solution for rendering errors -->
+        <client-only>
+          <user-teaser :user="follow" />
+        </client-only>
+      </ds-space>
+      <ds-space v-if="this.allConnectionsCount - this.connections.length" margin="small">
+        <base-button
+          @click="fetchConnections"
+          :loading="this.isLoading"
```
mattwr18:
```suggestion
          :loading="$apollo.loading"
```

webapp/components/features/FollowList/FollowList.vue
```diff
@@ -0,0 +1,157 @@
...
+    allConnectionsCount() {
+      return this.user[`${this.type}Count`]
+    },
+  },
+  methods: {
+    uniq(items, field = 'id') {
+      return uniqBy(items, field)
+    },
+    async fetchConnections() {
+      this.$set(this, 'isLoading', true)
```
mattwr18: nice, I guess... haven't seen this before... can't we just use `$apollo.loading`? see :point_up_2:

webapp/components/features/FollowList/FollowList.vue
```diff
@@ -0,0 +1,157 @@
...
+      const { data } = await this.$apollo.query({
+        query: this.queries[this.type],
+        variables: { id: this.user.id },
+        // neither result nor update are being called when defined here (?)
+      })
+      this.additionalConnections = data.User[0][this.type]
+      this.$set(this, 'isLoading', false)
+    },
+    setFilter(evt) {
+      this.$set(this, 'filter', evt.target.value)
```
mattwr18: is there a benefit to use `this.$set` over just `this.filter = evt.target.value`?

webapp/components/features/FollowList/FollowList.vue
```diff
@@ -0,0 +1,157 @@
...
+    },
+  },
+}
+</script>
+
+<style lang="scss">
+.hc-connections {
+  position: relative;
+  max-height: 424px;
```
mattwr18: can we somehow use one of our tokens from `tokens.scss`
we favor token values over magic numbers. See https://github.com/Human-Connection/Human-Connection/blob/master/webapp/scss.md#we-use-design-tokens-instead-of-magic-numbers

webapp/components/features/FollowList/FollowList.vue
```diff
@@ -0,0 +1,157 @@
...
+      return uniqBy(items, field)
+    },
+    async fetchConnections() {
+      this.$set(this, 'isLoading', true)
+      const { data } = await this.$apollo.query({
+        query: this.queries[this.type],
+        variables: { id: this.user.id },
+        // neither result nor update are being called when defined here (?)
+      })
+      this.additionalConnections = data.User[0][this.type]
```
mattwr18: I think we could use the same query for `User` and update the apollo cache by using `updateQuery`
there are some examples of this, that have some differences, but you could check them out, or we could look together. For example, `/pages/index.vue`
Also, see https://vue-apollo.netlify.com/api/smart-query.html#fetchmore
It would be nice to take advantage of some of the power of `smart` queries, reactive variables, and apollo caching to avoid adding extra queries.

webapp/plugins/vue-directives.js
```diff
@@ -7,7 +7,14 @@ export default ({ app }) => {
...
         if (binding.value !== false) {
-          el.focus()
+          let target = el
+          const isInput = target instanceof HTMLInputElement
+          if (!isInput) {
+            target = el.querySelector('input')
+          }
+          if (target && typeof target.focus === 'function') {
+            target.focus()
+          }
```
mattwr18: is the fix to the issue you were facing in the tests? do you have a link to where you found this fix?
Also, would need to have a closer look at this.

webapp/components/features/FollowList/FollowList.vue
rbeer: resolved in f69e7e4

webapp/components/features/FollowList/FollowList.vue
rbeer: resolved in f69e7e4

webapp/components/features/FollowList/FollowList.vue
rbeer: good idea :smile: 
"resolved in f69e7e4"

webapp/components/features/FollowList/FollowList.spec.js
```diff
@@ -0,0 +1,129 @@
...
+
+  describe('mount', () => {
+    beforeAll(() => {
+      store = new Vuex.Store({
+        getters,
+      })
+    })
+
+    describe('given a user with connections', () => {
+      ;['following', 'followedBy'].forEach((type) =>
```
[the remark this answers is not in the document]
rbeer: changed in [e6361c9](https://github.com/Human-Connection/Human-Connection/pull/3293/commits/e6361c9b740f104c0a0d86b463be835f10311f3d#diff-970486bf70217f927b18aadea0c8b1b4)

webapp/components/features/FollowList/FollowList.spec.js
```diff
@@ -0,0 +1,129 @@
...
+          })
+
+          it(`has a button to load all remaining users ${type}`, async () => {
+            jest.useFakeTimers()
+
+            wrapper.find('button').trigger('click')
+            await jest.runAllTicks()
+            await wrapper.vm.$nextTick()
+
+            expect(wrapper.vm.connections.length).toBe(propsData.user[`${type}Count`])
```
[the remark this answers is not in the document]
rbeer: changed in [e6361c9](https://github.com/Human-Connection/Human-Connection/pull/3293/commits/e6361c9b740f104c0a0d86b463be835f10311f3d#diff-970486bf70217f927b18aadea0c8b1b4L80)

webapp/components/features/FollowList/FollowList.spec.js
```diff
@@ -0,0 +1,129 @@
...
+            expect(queryMock).toHaveBeenCalledWith({
+              query: wrapper.vm.queries[type],
+              variables: { id: propsData.user.id },
```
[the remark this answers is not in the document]
rbeer: removed, because not needed anymore in [e6361c9](https://github.com/Human-Connection/Human-Connection/pull/3293/commits/e6361c9b740f104c0a0d86b463be835f10311f3d#diff-970486bf70217f927b18aadea0c8b1b4)

In this case would have been necessary, since the user is generated by `faker`, i.e. the id not known before the test runs.

webapp/components/features/FollowList/FollowList.vue
rbeer: You will see that I have refactored FollowList to not issue any queries and get the `:loading` state from its parent. `$apollo.loading` doesn't seem to have a global effect.

webapp/components/features/FollowList/FollowList.vue
rbeer: The only crucial difference I've read about is that `this.$set` re-renders the component when you change a nested value of a reactive variable.

```js
...
  data() {
    return {
      foo: true
    }
  }
...
  methods() {
    assignFoo() {
      // re-renders as expected from a reactive variable
      this.foo = false
    }
  }
```

```js
...
  data() {
    return {
      foo: { bar: true }
    }
  }
...
  methods() {
    assignBar() {
      // does _not_ re-render
      this.foo.bar = false 
    }
    setBar() {
      // re-renders
      this.$set(this.foo, 'bar', false)
    }
  }
```

webapp/components/features/FollowList/FollowList.vue
rbeer: changed in d62be2b

webapp/plugins/vue-directives.js
rbeer: I don't think so.
`v-focus` just existed; it was not used by any component, so I took the liberty to make it target <input\> elements.

webapp/components/features/FollowList/FollowList.vue
rbeer: changed in a6484e3

webapp/components/features/FollowList/FollowList.vue
```diff
@@ -0,0 +1,133 @@
+<template>
+  <base-card class="follow-list">
+    <template v-if="connections && connections.length">
+      <ds-text tag="h5" color="soft" class="spacer-x-small">
+        {{ userName | truncate(15) }} {{ $t(`profile.network.${type}`) }}
+      </ds-text>
```
mattwr18: let's refactor out all styleguide components? you can use a `h5` tag here and apply `CSS` to make it soft, etc...
and if you give this a class of `title`, for example, you can target it with `CSS` like: 
```
.follow-list {
  > .title {
   ....
  }
}
```
which says only target the class title that is a direct child of `follow-list`

webapp/components/features/FollowList/FollowList.vue
```diff
@@ -0,0 +1,133 @@
...
+        <user-teaser
+          v-for="connection in filteredConnections"
+          class="spacer-x-small"
+          :user="connection"
+          :key="connection.id"
+        />
+      </div>
+      <base-button
+        v-if="allConnectionsCount - connections.length"
+        @click="$emit('fetchAllConnections', type)"
```
mattwr18: we have been in the habit of putting the `events` at the bottom of the list.

webapp/pages/profile/_id/_slug.vue
```diff
@@ -89,65 +89,19 @@
...
-          <template v-else>
-            <p style="text-align: center; opacity: 0.5;">
-              {{ userName }} {{ $t('profile.network.followingNobody') }}
-            </p>
-          </template>
-        </base-card>
+        <follow-list
+          :user="user"
+          type="followedBy"
+          @fetchAllConnections="fetchAllConnections"
```
mattwr18: same as :point_up_2:

webapp/pages/profile/_id/_slug.vue
```diff
@@ -507,7 +459,11 @@ export default {
         return UserQuery(this.$i18n)
       },
       variables() {
-        return { id: this.$route.params.id }
+        return {
+          id: this.$route.params.id,
+          followedByCount: this.followedByCount || 7,
+          followingCount: this.followingCount || 7,
```
mattwr18: do we need to have both a default in `data` and a fallback here?

webapp/plugins/vue-directives.js
mattwr18: I meant the issue with `Vue.nextTick()` not doing what we expected in the test you wrote.

rbeer: No, this vue-directives `v-focus` change is unrelated to that.

Like you suggested, I had a look at other use cases of Vue.nextTick() in the tests and noticed that none of those other instances waits for the actual result of a function. It's always just used like
```js
wrapper.find('button').emit('click');
await Vue.nextTick();
expect(getFooAsync).toHaveBeenCalled()
```
i.e. not testing with what the Promise to resolved, merely whether its producing function had been called.

So I went to the [jest documentation](https://jestjs.io/docs/en/timer-mocks) and found that jest manages its own micro-queue for timers.

There is a package called flush-promises which I think is mentioned in the vue docs, but I didn't want to install another package. Turns out, that it is [already in use](https://github.com/Human-Connection/Human-Connection/blob/f13eb29ad342392edee61b82c02a126a5fdbae3b/webapp/pages/settings/my-social-media.spec.js#L64).

webapp/components/features/FollowList/FollowList.vue
rbeer: Was just hesitating to do that. So basically, whenever I come across a ds-* tag, refactor it out?
The question is then, what design-tokens do I use? Just any that fit and if nothing applies, create one? We don't want to end up with 20 definitions of a card-heading is what I'm concerned about. @alina-beck 

Anyway, there is a ds-input later on, I'll start with by turning it into a base component.

webapp/components/features/FollowList/FollowList.vue
rbeer: I bumped into that question, :smile: 
Just apply [Vue's styleguide regarding attribute order](https://vuejs.org/v2/style-guide/#Element-attribute-order-recommended)?

webapp/plugins/vue-directives.js
mattwr18: yes, that's great! nice catch!

webapp/components/features/FollowList/FollowList.vue
mattwr18: yes, please. There should be design tokens for all the sizes that are used in `ds-text`... otherwise maybe see if one of them looks good, ie not too big or too small.
As far as the `ds-input`, I would probably leave that because it would actually involve migrating the `ds-input` component and refactoring, ie separate ticket(code creep).
My larger point was that if it's possible to accomplish the same with `html` tags and `CSS`, let's do that, so we don't make it harder on ourselves to refactor later.

webapp/components/features/FollowList/FollowList.vue
mattwr18: yes, please!

webapp/components/features/FollowList/FollowList.spec.js
```diff
@@ -0,0 +1,133 @@
...
+        it(`renders the connections`, () => {
+          expect(followedByWrapper.findAll('.user-teaser').length).toEqual(user.followedBy.length)
+          expect(followingWrapper.findAll('.user-teaser').length).toEqual(user.following.length)
+        })
+
+        it(`has a button to load all remaining connections`, async () => {
+          followingWrapper.find('.base-button').trigger('click')
+          followedByWrapper.find('.base-button').trigger('click')
+          expect(followingWrapper.emitted('fetchAllConnections')).toBeTruthy()
+          expect(followedByWrapper.emitted('fetchAllConnections')).toBeTruthy()
```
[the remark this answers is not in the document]
rbeer: Then let's test the type explicitly: daa8816

webapp/components/features/FollowList/FollowList.vue
rbeer: done in 8e98570

webapp/pages/profile/_id/_slug.vue
rbeer: done in 8e98570

webapp/pages/profile/_id/_slug.vue
rbeer: no, we actually don't
removed in 9dd65aa

webapp/components/features/FollowList/FollowList.vue
rbeer: done in 72dfeb3

webapp/components/features/FollowList/FollowList.vue
rbeer: done in e6361c9

webapp/components/features/FollowList/FollowList.vue
```diff
@@ -0,0 +1,150 @@
...
+      <div :class="connectionsClass">
+        <user-teaser
+          v-for="connection in filteredConnections"
+          :key="connection.id"
+          :user="connection"
+          class="spacer-x-small"
+        />
+      </div>
+      <base-button
+        v-if="allConnectionsCount - connections.length"
```
alina-beck, line 16: So `0` is `falsy`? Is this maybe more readable:
```suggestion
        v-if="allConnectionsCount > connections.length"
```

webapp/components/features/FollowList/FollowList.vue
```diff
@@ -0,0 +1,150 @@
...
+    },
+    allConnectionsCount() {
+      return this.user[`${this.type}Count`]
+    },
+    connections() {
+      return this.user[this.type]
+    },
+    connectionsClass() {
+      const overflow = this.connections.length > 7 ? ' --overflow' : ''
+      return `connections${overflow}`
```
alina-beck, line 75: Since it's just one property you could maybe just
```suggestion
      return `connections${this.connections.length > 7 ? ' --overflow' : ''}`
```

webapp/components/features/FollowList/FollowList.vue
```diff
@@ -0,0 +1,150 @@
...
+        @click="$emit('fetchAllConnections', type)"
+      >
+        {{
+          $t('profile.network.andMore', {
+            number: allConnectionsCount - connections.length,
+          })
+        }}
+      </base-button>
+      <ds-input
+        v-if="connections.length > 7"
```
alina-beck, line 29: does it make sense to set the `7` as constant somewhere?

webapp/components/features/FollowList/FollowList.vue
alina-beck: Second that. I'll skip this bit for now – if you want to have a look together, let me know @mattwr18 😅

webapp/components/features/FollowList/FollowList.vue
```diff
@@ -0,0 +1,150 @@
...
+  },
+}
+</script>
+
+<style lang="scss">
+.follow-list {
+  display: flex;
+  flex-direction: column;
+  position: relative;
+  //max-height: ($size-avatar-small + $space-x-small * 2) * 8;
```
alina-beck, line 127: since this is commented out it's probably not needed?

webapp/components/features/FollowList/FollowList.vue
```diff
@@ -0,0 +1,150 @@
...
+  width: auto;
+
+  > .title {
+    color: $text-color-soft;
+    font-size: $font-size-small + 0.03rem;
+  }
+
+  .connections.--overflow {
+    height: ($size-avatar-base + $space-x-small * 2) * 5;
```
alina-beck, line 136: I see what you're doing here, but why not set a fixed number instead? Imho this is overly complicated and since it still includes a "magic number" it's not extremely helpful to do this calculation. If you agree, please define the height as a variable in `tokens` and set it on `.connections` directly! :grin:

webapp/components/features/FollowList/FollowList.vue
```diff
@@ -0,0 +1,150 @@
...
+.follow-list {
+  display: flex;
+  flex-direction: column;
+  position: relative;
+  //max-height: ($size-avatar-small + $space-x-small * 2) * 8;
+  width: auto;
+
+  > .title {
+    color: $text-color-soft;
+    font-size: $font-size-small + 0.03rem;
```
alina-beck, line 132: Why the `+ 0.03rem`? Can't you pick a font size that fits? That's sort of the whole point of using variables. 😅 I don't mind if it's a bit bigger or smaller than it was before – if you use an existing token it's definitely an improvement!

webapp/components/features/FollowList/FollowList.story.js
```diff
@@ -0,0 +1,123 @@
...
+      components: { FollowList },
+      store: helpers.store,
+      data() {
+        return { user: fuzzyFilterUser }
+      },
+      template: wrapTemplates([
+        '<follow-list :user="user" />',
+        '<follow-list :user="user" type="followedBy">',
+      ]),
+    }
```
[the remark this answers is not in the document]
rbeer: Helps understanding the fuzzy filter logic; that's why a fixed set of followers (their names).
Try for example 'ger' as filter in http://localhost:3002/?path=/story/followlist--fuzzy-filter to see the full-match preference, i.e. full matches rank higher, even when they are in the back of the name.

webapp/graphql/User.js
```diff
@@ -14,7 +14,7 @@ export default (i18n) => {
     ${userCountsFragment}
     ${locationAndBadgesFragment(lang)}
 
-    query User($id: ID!) {
+    query User($id: ID!, $followedByCount: Int, $followingCount: Int) {
```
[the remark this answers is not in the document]
rbeer: Backend returns all entities, when `last` parameter is not set or `null`.

https://github.com/Human-Connection/Human-Connection/pull/3293/files/72dfeb3983689737afbdafeda1ec1a5862da3754#diff-399a892519b3cfbb5e4649fc25b0be0cR28

webapp/components/features/FollowList/FollowList.story.js
```diff
@@ -0,0 +1,123 @@
...
+        },
+      },
+      template: wrapTemplates([
+        '<follow-list :user="user" @fetchAllConnections="fetchAllConnections"/>',
+        '<follow-list :user="user" type="followedBy" @fetchAllConnections="fetchAllConnections"/>',
+      ]),
+    }
+  })
+
+  .add('with all connections', () => {
```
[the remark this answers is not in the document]
rbeer: The profile page had the count of displayed connections hardcoded to 7; I just took that over, when I started.
It makes way more sense now to introduce a `hasMore` computed value of `user.followingCount > user.following.length`

webapp/components/features/FollowList/FollowList.vue
rbeer: > The profile page had the count of displayed connections hardcoded to 7; I just took that over, when I started.
It makes way more sense now to introduce a `hasMore` computed value of `user.followingCount > user.following.length`

webapp/components/features/FollowList/FollowList.vue
rbeer: Not really, no. Next ones are either too small or too big.

webapp/components/features/FollowList/FollowList.vue
rbeer: >The profile page had the count of displayed connections hardcoded to 7; I just took that over, when I started.
It makes way more sense now to introduce a `hasMore` computed value of `user.followingCount > user.following.length`

webapp/components/features/FollowList/FollowList.vue
rbeer: >The profile page had the count of displayed connections hardcoded to 7; I just took that over, when I started.
It makes way more sense now to introduce a `hasMore` computed value of `user.followingCount > user.following.length`

webapp/pages/profile/_id/_slug.vue
```diff
@@ -310,12 +263,14 @@ export default {
...
     MasonryGrid,
     MasonryGridItem,
+    FollowList,
   },
   transition: {
     name: 'slide-up',
     mode: 'out-in',
   },
   data() {
+    console.log('component data()')
```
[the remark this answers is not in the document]
rbeer: 2c7ca87ac

webapp/pages/profile/_id/_slug.vue
```diff
@@ -483,6 +433,9 @@ export default {
       this.user.followedByCurrentUser = followedByCurrentUser
       this.user.followedBy = followedBy
     },
+    fetchAllConnections(type) {
+      this[`${type}Count`] = Infinity
```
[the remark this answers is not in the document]
rbeer: 2c7ca87ac

webapp/pages/profile/_id/_slug.vue
```diff
@@ -89,65 +89,19 @@
...
-            <p style="text-align: center; opacity: 0.5;">
-              {{ userName }} {{ $t('profile.network.followedByNobody') }}
-            </p>
-          </template>
-        </base-card>
+        <follow-list
+          :user="user"
+          type="following"
+          @fetchAllConnections="fetchAllConnections"
+          :loading="$apollo.loading"
```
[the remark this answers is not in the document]
rbeer: 2c7ca87ac

webapp/components/features/FollowList/FollowList.story.js
```diff
@@ -0,0 +1,123 @@
...
+  followedBy: [],
+  followedByCount: 0,
+  following: [],
+  followingCount: 0,
+}
+
+const wrapTemplates = (templates) => `
+<div style="display: flex; flex-wrap: wrap;">
+${templates.map((template) => `<div style="margin: 8px;">${template}</div>`).join('')}
+</div>
```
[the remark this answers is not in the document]
rbeer: 7ff9d617d

webapp/components/features/FollowList/FollowList.vue
rbeer: b3f113700

webapp/components/features/FollowList/FollowList.vue
rbeer: b3f113700

webapp/components/features/FollowList/FollowList.vue
rbeer: b3f113700

webapp/components/features/FollowList/FollowList.vue
rbeer: e018e2ead

webapp/components/features/FollowList/FollowList.vue
rbeer: e018e2ead

webapp/components/features/FollowList/FollowList.vue
rbeer: e018e2ead

webapp/components/features/FollowList/FollowList.story.js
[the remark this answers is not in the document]
rbeer: e018e2ead

webapp/components/features/FollowList/FollowList.spec.js
```diff
@@ -0,0 +1,150 @@
...
+          expect(wrapper.emitted('fetchAllConnections')).toEqual([['following']])
+        })
+      })
+
+      describe('of `followedBy`', () => {
+        it('uses the `followedBy` data on :user', () => {
+          const wrapper = Wrapper({ type: 'followedBy' })
+          wrapper.find('.base-button').trigger('click')
+
+          expect(wrapper.vm.connections).toEqual(user.followedBy)
```
Mogge, line 83: Same here

webapp/components/features/FollowList/FollowList.spec.js
```diff
@@ -0,0 +1,150 @@
...
+          expect(wrapper.vm.allConnectionsCount).toBe(user.followedByCount)
+          expect(wrapper.emitted('fetchAllConnections')).toEqual([['followedBy']])
+        })
+      })
+    })
+
+    describe('given no type', () => {
+      it('defaults type to `following`', () => {
+        const wrapper = Wrapper()
+        expect(wrapper.vm.type).toBe('following')
```
Mogge, line 93: May be
```javascript
expect(wrapper.find('h5').text()).toContain('profile.network.following')
```

webapp/components/features/FollowList/FollowList.spec.js
```diff
@@ -0,0 +1,150 @@
...
+      })
+    })
+
+    describe('given a user', () => {
+      describe('without connections', () => {
+        it('displays the followingNobody message', () => {
+          const wrapper = Wrapper({ user: noConnectionsUser })
+          expect(wrapper.find('.nobody-message').text()).toBe(
+            `${noConnectionsUser.name} ${wrapper.vm.$t(`profile.network.followingNobody`)}`,
+          )
```
Mogge, line 103: `toContain('profile.network.followingNobody')` should be enough

webapp/components/features/FollowList/FollowList.spec.js
```diff
@@ -0,0 +1,150 @@
...
+        })
+
+        it('displays the followedByNobody message', () => {
+          const wrapper = Wrapper({ user: noConnectionsUser, type: 'followedBy' })
+          expect(wrapper.find('.nobody-message').text()).toBe(
+            `${noConnectionsUser.name} ${wrapper.vm.$t(`profile.network.followedByNobody`)}`,
```
Mogge, line 109: same here

webapp/components/features/FollowList/FollowList.spec.js
```diff
@@ -0,0 +1,150 @@
...
+      describe('with more than 7 loaded connections', () => {
+        let wrapper
+        beforeAll(() => {
+          wrapper = Wrapper({ user: allConnectionsUser })
+        })
+
+        it('renders the connections', () => {
+          expect(wrapper.findAll('.user-teaser')).toHaveLength(allConnectionsUser.followingCount)
+        })
+
```
Mogge, line 139: Here is what I expected before. Looks good. But how does the `.--overflow` appear without triggering the click?

webapp/components/features/FollowList/FollowList.vue
Mogge: I like it and copied it to my knowledge base :)

webapp/components/features/FollowList/FollowList.spec.js
rbeer: The `connectionsClass` string is [computed](https://github.com/Human-Connection/Human-Connection/pull/3293/files#diff-68c74572b7e55ffbcef954e9c4d9d487R77) and then used in [line 7](https://github.com/Human-Connection/Human-Connection/pull/3293/files#diff-68c74572b7e55ffbcef954e9c4d9d487R7)

webapp/components/features/FollowList/FollowList.vue
rbeer: Thanks, @Mogge 
It fails to prioritize full matches when the strings are longer (>15). In some cases, the first match is long enough to overtake a 1-2 separation match further to the start.
```js
// works as snippet in browser dev-tools
(() => {
  const fuzzyFilter = (str, filter) => {
    const fuzzyExpression = new RegExp(
      `${filter.split('').reduce((expr, c) => `${expr}([^${escape(c)}]*${escape(c)})`, '')}`,
      'i',
    )
    const match = str.match(fuzzyExpression)
    if (!match) {
      return false
    }
    let score = 1
    for (let i = 1; i <= filter.length; i++) {
      score *= match[i].length
    }
    return {
      str,
      filter,
      fuzzyExpression,
      match,
      score,
    }
  }
  // .score === 13
  console.log(fuzzyFilter('thank you, very much', 'ery'))
  // .score === 12 (i.e. higher in the list)
  console.log(fuzzyFilter('Pecurly', 'ery'))
})();

File: webapp/components/features/FollowList/FollowList.spec.js [hunk 41]
```diff
@@ -0,0 +1,150 @@
...
+      })
+    })
+
+    describe('given a type', () => {
+      describe('of `following`', () => {
+        it('uses the `following` data on :user', () => {
+          const wrapper = Wrapper({ user, type: 'following' })
+          wrapper.find('.base-button').trigger('click')
+
+          expect(wrapper.vm.connections).toEqual(user.following)
```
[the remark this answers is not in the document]
rbeer: ee3373117

webapp/components/features/FollowList/FollowList.spec.js
rbeer: ee3373117

webapp/components/features/FollowList/FollowList.spec.js
rbeer: Just did the whole wrapper's text.
ee3373117

webapp/components/features/FollowList/FollowList.spec.js
rbeer: ee3373117

webapp/components/features/FollowList/FollowList.spec.js
rbeer: ee3373117

webapp/components/_new/generic/UserAvatar/UserAvatar.vue
```diff
@@ -3,9 +3,11 @@
     <span class="initials">{{ userInitials }}</span>
     <base-icon v-if="isAnonymous" name="eye-slash" />
     <img
-      v-else
+      v-if="user && user.avatar"
       :src="user.avatar | proxyApiUrl"
       class="image"
+      :alt="user.name"
+      :title="user.name"
```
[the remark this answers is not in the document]
rbeer: Really deserves its own refactoring session/PR.
I got this from the storybook/ally-plugin hints tab. There are a lot more, I eventually not pursued, 'cause that would have involved touching/refactoring 2-3 other components.

webapp/storybook/helpers.js
```diff
@@ -57,6 +60,16 @@ const helpers = {
       </layout>`,
     }
   },
+  fakeUser(n) {
```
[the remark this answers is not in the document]
rbeer: Thanks :)
The only thing I don't like is that I'm using it again, in the tests.
https://github.com/Human-Connection/Human-Connection/pull/3293/files/1ac34a8fbadccacbdf92b074eff1caa02d59afeb#diff-970486bf70217f927b18aadea0c8b1b4R4

Tirokk:
If you would just add the half line of an additional follower appearing after the click of the button that the user can see there is something to scroll I would call it perfect! 🤗 @rbeer

webapp/storybook/layout.vue
```diff
@@ -1,7 +1,7 @@
 <template>
-  <ds-container class="container">
+  <div class="container">
     <slot />
-  </ds-container>
```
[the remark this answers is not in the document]
rbeer: I'll make it a habit :)

webapp/components/features/FollowList/FollowList.vue
```diff
@@ -0,0 +1,150 @@
+<template>
+  <base-card class="follow-list">
+    <template v-if="connections && connections.length">
+      <h5 class="title spacer-x-small">
+        {{ userName | truncate(15) }} {{ $t(`profile.network.${type}`) }}
+      </h5>
+      <div :class="connectionsClass">
```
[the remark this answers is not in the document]
rbeer: 93b4e6381

Tirokk:
Nice adjustments.
They work for me and looking really good to me! @rbeer 💫🚀

I would very much like to merge this feature @alina-beck @mattwr18 @Mogge @roschaefer @ogerly !!! 😍

ogerly:
ezgif com-video-to-gif (1)

PR #3297, opened 2020-03-19
feat: Search for Hashtags

### LINKED ISSUE #2143: 🚀 [Feature] Search by hashtags
## :rocket: Feature
Users should be able to search by hashtags by entering a keyword in the search field.

### User Problem
At the moment users can only find posts by hashtag by navigating to a post and clicking the hashtag in its content. This takes a long time and can be very frustrating, especially since they first need to _find_ a post that contains the hashtag they are interested in.

### Implementation
The list of search results should include a section with `hashtags` where the relevant hashtags are listed. Selecting a hashtag initiates the hashtag search the way it is working now (i.e. displaying the hashtag on top of a page with posts containing that hashtag).

### Design & Layout
The hashtag results should look similar to the post and user results in this example:

### Additional context
This is **Step 3** in our search implementation – see #2146.
**Step 1** is adding users to the list of search results: #1463.
**Step 2** is limiting the number of search results to 15: #2142.

related to #1068

Mogge:
## 🍰 Pullrequest
Adding the search results for hashtags to the drop down menu

- fixes #2143

### Todo

- [X] None


webapp/pages/index.vue
```diff
@@ -130,6 +130,11 @@ export default {
       return this.$apollo.loading || (this.posts && this.posts.length > 0)
     },
   },
+  watch: {
+    '$route.query.hashtag'() {
+      this.hashtag = this.$route.query.hashtag ? this.$route.query.hashtag : null
+    },
+  },
```
mattwr18: I think there is a better way to do this... we just used `watchQuery` in the other `PR` 
have a look @Mogge https://github.com/Human-Connection/Human-Connection/pull/3271/files#diff-a15746a1a8163b3facc5cf6d0458d9e5R10

Mogge: Thanks for the tip. Much better to use watchQuery

backend/src/db/migrations/20200207080200-fulltext_index_for_tags.js
```diff
@@ -0,0 +1,51 @@
...
+export async function down(next) {
+  const driver = getDriver()
+  const session = driver.session()
+  const transaction = session.beginTransaction()
+
+  try {
+    // Implement your migration here.
+    await transaction.run(`
+      CALL db.index.fulltext.drop("tag_fulltext_search")
+    `)
```
mattwr18: I really like that you are adding the `down` function as well. I think I haven't added it on every migration, and some are easier to add then others, but just thought I'd point out that this is a great habit!

PR #3323, opened 2020-03-23
fix(migration): return null for Image.url

### LINKED ISSUE #3322: 🐛 [Bug] Image.url returns undefined and crashes site
## :bug: Bugreport

Image.url should never return undefined

1. dump production data locally
2. visit site
3. site crashes on index
4. 
5. Profit

mattwr18:
### Issues

- fixes #3322 

should we add an id to `Image`, which is non-nullable?
can we somehow not return `deleted` users in our following/followedBy lists?


backend/src/middleware/softDelete/softDeleteMiddleware.js
```diff
@@ -17,10 +17,10 @@ const obfuscate = async (resolve, root, args, context, info) => {
     root.contentExcerpt = 'UNAVAILABLE'
     root.title = 'UNAVAILABLE'
     root.slug = 'UNAVAILABLE'
-    root.avatar = 'UNAVAILABLE'
+    root.avatar = null
```
roschaefer: Good catch :+1:

backend/src/schema/resolvers/users/location.spec.js
```diff
@@ -19,48 +19,48 @@ const updateUserMutation = gql`
 const newlyCreatedNodesWithLocales = [
   {
     city: {
-      lng: -74.5763,
-      lat: 41.1534,
+      id: expect.stringContaining('place'),
```
roschaefer: Why do these ids change? Do we make API calls during testing?

mattwr18:
I asked the question about an id that is `non-nullable` at the point directly after I had removed the `non-nullable` from the `url`, but I've since added it back because we can use the `softDeleteMiddleware` to set these values to null, if a user is deleted.

I like your idea to remove the deleted user from the following list, but I couldn't work out how to do that yesterday. I guess it could be a function that adds a filter like `filterForMutedUsers`, but `filterForDeletedUsers`?

PR #3324, opened 2020-03-23
fix(build): make locations spec more resilient

mattwr18:
## 🍰 Pullrequest

Fixes build... do not hard code the id's as they are subject to change from `mapbox`'s side.
it is enough that the ids contain either `place`, `region`, `country`
### Issues

- None

### Todo

- [X] None


mattwr18:
> This surely does the trick. I am still asking myself how we suddenly receive different ids from mapbox.

yeah, I think that's strange, I wonder if this means that we are going to be creating multiple identical locations because the id's have changed :woman_shrugging:

PR #3343, opened 2020-03-26
Remove dangling image urls

roschaefer:
## 🍰 Pullrequest
As discussed in https://github.com/Human-Connection/Human-Connection/pull/3262#discussion_r398634249 we would like to remove dangling urls.

### Issues
- None

### Todo
- [X] None

### DIFF
--- a/backend/src/db/migrations/20200326160326-remove_dangling_image_urls.js
+++ b/backend/src/db/migrations/20200326160326-remove_dangling_image_urls.js
@@ -1,0 +1,61 @@
+import { getDriver } from '../../db/neo4j'
+import { existsSync } from 'fs'
+
+export const description = `
+  In this review:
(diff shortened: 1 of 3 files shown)


backend/src/db/migrations/20200326160326-remove_dangling_image_urls.js
```diff
@@ -0,0 +1,61 @@
...
+    const urls = records.map((record) => record.get('url'))
+    const danglingUrls = urls.filter((url) => {
+      const fileLocation = `public${url}`
+      return !existsSync(fileLocation)
+    })
+    await transaction.run(
+      `
+      MATCH(image:Image)
+      WHERE image.url IN $danglingUrls
+      REMOVE image.url
```
mattwr18: I don't think it's enough to remove image.url
I believe we need to detach delete image, oder

PR #3344, opened 2020-03-26
[WIP] Image resize service

roschaefer:
## 🍰 Pullrequest
We don't have to change the backend. [imgproxy](https://github.com/imgproxy/imgproxy) defines a URL scheme which the frontend can build on its own.

### Issues

- None

### Todo

- [X] None


backend/.env.template
```diff
@@ -23,3 +23,5 @@ AWS_SECRET_ACCESS_KEY=
 AWS_ENDPOINT=
 AWS_REGION=
 AWS_BUCKET=
+IMGPROXY_KEY=21b2b7b270d5fd48f02491e520fc89b8f31bab66ef2a00b3f254ce3f02c6fae6407345f3b394580ffa751973f624ce0740566ff39960f75da6ec9f45dfc7b9c2
+IMGPROXY_SALT=8144a980d252b21aa5ea8c82379f3b7948dfb6373f60a2f633de40d64be8a326de46490aef2f4c2ed4278b76cf1309fd004338026a7737c8ca805415c38605a4
```
roschaefer: This secret will be different in production

webapp/components/Badges.vue
```diff
@@ -1,7 +1,7 @@
 <template>
   <div :class="[badges.length === 2 && 'hc-badges-dual']" class="hc-badges">
     <div v-for="badge in badges" :key="badge.id" class="hc-badge-container">
-      <img :title="badge.key" :src="badge.icon | proxyApiUrl" class="hc-badge" />
+      <img :title="badge.key" :src="`/img/badges/${badge.id}.svg`" class="hc-badge" />
```
roschaefer: I decided to move the image files into webapp docker image. Why were they in backend image anyways? :thinking:

webapp/components/_new/generic/UserAvatar/UserAvatar.vue
```diff
@@ -2,18 +2,18 @@
   <div :class="['user-avatar', size && `--${this.size}`]">
     <span class="initials">{{ userInitials }}</span>
     <base-icon v-if="isAnonymous" name="eye-slash" />
-    <img
-      v-else
-      :src="user.avatar | proxyApiUrl"
-      class="image"
-      @error="$event.target.style.display = 'none'"
-    />
+    <responsive-image v-else :image="user.avatar" class="image" sizes="120px" />
```
roschaefer: So this is the only place where I decided to set a `sizes`. It's fixed to one size only. I went for 120px which is a little more than the 114px avatar image on the profile page. The avatar image on the post card is 34px wide on my machine. Good @alina-beck ?

If we want to keep that, maybe we want to query for 120px only in the respective graphql request.

webapp/components/Badges.vue
mattwr18: that's a good question @roschaefer ...
could we just upload these to `Spaces` as well? then others could more easily reuse while customizing the badges, no? Also, it means smaller docker images

webapp/components/_new/generic/UserAvatar/UserAvatar.vue
mattwr18: and giving it `34px, 120px` sizes would do anything because the size it uses is determined by the viewport width, correct? I guess it's not any different on mobile either, as far as I understand how it works.

mattwr18:
yeah, so I didn't have a lot of time to go through this, but left some thoughts, asked for some small changes. I think you mentioned it already, but when I tried to use locally without docker, it crashed.
I'll take a closer look when I get a chance.

webapp/components/Badges.vue
roschaefer: The badges are small .svg files. At the moment we don't have a feature to add more badges. So I decided to put them into the docker image.

webapp/components/_new/generic/UserAvatar/UserAvatar.vue
roschaefer: Yes, you the developer must give the browser a hint how wide an image is going to be for a given viewport.

Tirokk:
Not stale at all …

PR #3413, opened 2020-04-04
fix(database): set Report.disable

### LINKED ISSUE #3412: 💥 [DevOps] Database manipulation to set Report.disable
## :fire: DevOps ticket

Create a database migration to set  `Report.disable` to the final decision, last review rules.

### Motive

Currently, we have about 300 Reports with `disable` as `null`, and it's labeled as non-nullable in `Report.gql`. Also, we might need toggle the state of a resource disable in different `Report` decisions. 
This can make it difficult to find out which `Report` disabled it, under which `rule` . Right now, we are limited to `latestReviewUpdatedAtRules`, but in the future we might move towards more of a community moderated model. 
### Related issues

This was a todo on issue https://github.com/Human-Connection/Human-Connection/issues/3074
but it was kind of forgotten about, and the other parts of that issue have already been merged in a while ago. It made sense to extract this into its own issue. 
### Implementation

Write a database migration.
### Validation

```
MATCH (report:Report) WHERE report.disable IS NULL RETURN count(report);
```
should return `0`

mattwr18:
- we should be able to look at a resource's latest report and tell
  whether the reviews decided it should be disabled

### Issues

- fixes #3412 

### Todo

- [ ] How can we test that this has used the latest review other than our knowledge of cypher queries?

### DIFF
--- a/backend/src/db/migrations/20200404100445-set_reportdisable_to_latest_review_decision.js
+++ b/backend/src/db/migrations/20200404100445-set_reportdisable_to_latest_review_decision.js
@@ -1,0 +1,58 @@
+import { getDriver } from '../../db/neo4j'
+
+export const description = ''
+
+export async function up(next) {
+  const driver = getDriver()
+  const session = driver.session()
+  const transaction = session.beginTransaction()
+
+  try {
+    const { records } = await transaction.run(`
+      MATCH (report:Report)
+      WHERE NOT exists(report.disable)
+      AND report.closed = TRUE
+      RETURN report.id as reportId;
+    `)
+    const reportIds = await records.map((record) => record.get('reportId'))
+    const updatedReports = await Promise.all(
+      reportIds
+        .map((reportId) => {
+          return async () => {
+            const updatedReportsTransactionPromise = await transaction.run(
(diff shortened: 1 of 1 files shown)


mattwr18:
@Tirokk and I discussed whether we want to add the `reivew.closed` in this migration, or not. It was removed with the rational that we could just find the `Report.closed` and then grab it's latest review and that must be the one that closed it. While this is true for now, under the `ReportRule: latestReviewUpdatedAtRules`, this is not necessarily true when/if we change the rule.
Also, I found it a bit more difficult to grab the latest review, then it sounds. In the end, I was able to do 
```
MATCH (report:Report {id: $reportId})<-[reviews:REVIEWED]-(:User)
WITH report, reviews
ORDER BY reviews.updatedAt DESC LIMIT 1
```
but this requires a knowledge of great knowledge of `cypher` and a higher chance that someone will believe they understand how it works, but is wrong, for some reason, than if we we just said give me the `review where review.closed = true`

Thoughts? @Human-Connection/hc-dev-team

Tirokk:
I agree to apply this Cipher to this migration.
Then we have a clear database again and don’t have to think about this in the future. @mattwr18 👍🏼
Thanks for caring! 💫

PR #3414, opened 2020-04-04
feat: [WIP] limit notifications to 25 for menu

mattwr18:
## 🍰 Pullrequest

There are some things to pay attention to with this, like the counter updating with `markAsRead`/`susbcriptions`, but in general I think we want to move away from asking for all of a users notifications every time they log in or refresh the page.
We have the `All notifications` page for this.
### Issues

- None

### Todo

- [X] None

### DIFF
--- a/backend/src/schema/resolvers/notifications.js
+++ b/backend/src/schema/resolvers/notifications.js
@@ -15,10 +15,9 @@ export default {
   },
   Query: {
     notifications: async (_parent, args, context, _resolveInfo) => {
-      const { user: currentUser } = context
-      const session = context.driver.session()
+      const { user: currentUser, driver } = context
+      const session = driver.session()
       let whereClause, orderByClause
-
       switch (args.read) {
         case true:
           whereClause = 'WHERE notification.read = TRUE'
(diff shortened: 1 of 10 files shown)


backend/src/schema/resolvers/notifications.js
```diff
@@ -52,17 +52,19 @@ export default {
           [(resource)-[:COMMENTS]->(post:Post)<-[:WROTE]-(author:User) | post{.*, author: properties(author)} ] AS posts
           WITH resource, user, notification, authors, posts,
           resource {.*, __typename: labels(resource)[0], author: authors[0], post: posts[0]} AS finalResource
-          RETURN notification {.*, from: finalResource, to: properties(user)}
           ${orderByClause}
+          RETURN { notificationsCount: toString(size(collect(notification))), notifications: collect(notification {.*, from: finalResource, to: properties(user)})} as noticationsResult
```
Mogge, line 56: I think this is not working. An object with length 1 is created and returned. So you cannot apply LIMIT or SKIP anymore on the result value

PR #3484, opened 2020-04-15
🍰[WIP][What todo with it?] Administer Digital Ocean Volumes

### LINKED ISSUE #3478: 💥 [DevOps] Remove uploads-claim and scale backend
## :fire: DevOps ticket

We are now using DigitalOcean Spaces for our uploads, we need to clean up by removing the uploads-claim in staging and production.
After this, we can scale the backend.
### Motive

We no longer use these persisent volumes/claims
 ### Related issues

### Validation

when the backend is scaled to 2 replicas and everything running properly

Tirokk:
## 🍰 Pullrequest

Changes Digital Ocean Volume snapshot files.

### Issues

- fixes #3478

### Todo

What we gonna do with the changes @mattwr18 ?
- Merge the changes for the storage amount in the master?
- Discard the changed `deployment/volumes/volume-snapshots/uploads.yaml` file or merge it?

### DIFF
--- a/deployment/volumes/uploads.yaml
+++ b/deployment/volumes/uploads.yaml
@@ -9,4 +9,4 @@
       - ReadWriteOnce
     resources:
       requests:
-        storage: {{ .Values.uploadsStorage }}
+        storage: 25Gi
--- a/deployment/volumes/volume-snapshots/uploads.yaml
+++ b/deployment/volumes/volume-snapshots/uploads.yaml
@@ -2,11 +2,15 @@
   kind: PersistentVolumeClaim
   apiVersion: v1
   metadata:
-    name: neo4j-data-claim
+    name: uploads-claim
     namespace: human-connection
   spec:
+    dataSource:
+      name: 2020-04-06-uploads-snapshot
+      kind: VolumeSnapshot
+      apiGroup: snapshot.storage.k8s.io
     accessModes:
       - ReadWriteOnce
     resources:
       requests:
-        storage: 5Gi
+        storage: 25Gi


mattwr18:
good question @Tirokk 
I think these are really only used for documentation, since we have decided to edit in place to make sure we don't accidentally delete some attributes.
I guess we should make copies(from production) of anything that is not secret(?)

deployment/volumes/neo4j-data.yaml
```diff
@@ -9,4 +9,4 @@
       - ReadWriteOnce
     resources:
       requests:
-        storage: {{ .Values.neo4jStorage }}
+        storage: 5Gi
```
Tirokk, line 12: But why we have changed this here if we don’t merge? @mattwr18 
We haven’t used it for our DevOps at all.

deployment/volumes/uploads.yaml
```diff
@@ -9,4 +9,4 @@
       - ReadWriteOnce
     resources:
       requests:
-        storage: {{ .Values.uploadsStorage }}
+        storage: 25Gi
```
Tirokk, line 12: But why we have changed this here if we don’t merge? @mattwr18

mattwr18: I wasn't suggesting we shouldn't merge... I was suggesting that instead of just merging these changes, we have a quick copy of the actual values in `deployments`, `services`, `pvc`, etc. and make sure there are not other things missing.
Also, we could consider how it's organized. What _we_ actually use, and what is just for others who might want to use local file systems for uploads, for example.

Tirokk:
Still to discuss …

PR #3494, opened 2020-04-21
fixed css so follow button isn't blocked by adblock

nila99:
I change the name of the css classes that held the follow button on a user's profile so that it no longer is blocked by uBlock.

### DIFF
--- a/webapp/components/FollowButton.vue
+++ b/webapp/components/FollowButton.vue
@@ -1,6 +1,6 @@
 <template>
   <base-button
-    class="follow-button"
+    class="track-button"
     :disabled="disabled || !followId"
     :loading="loading"
     :icon="icon"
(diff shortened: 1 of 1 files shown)


Tirokk:
Hi @nila99  ,
like I said already on the issue:
We are in a heavily transition on the project and I was on holidays for two weeks.
Right back now. 🤗
In the weeks before my holidays we had to sort out a lot of things here which bind our energy.

I recognised you pushed on our repository from your fork.
Unfortunately this causes a Travis build error.
Please push your feature branch to the parent repo and create a `PR` using `master` as the base branch.
Therefore you are already volunteer on our team and have the permission todo so.
Hope you are still interested.

I feel sorry that we may have annoyed you and go right on your code and review after you pushed. 💫

PR #3589, opened 2020-05-30
feat: 🍰 Alphabetically sorting tags using compute functions on index and more…

### LINKED ISSUE #2244: 🚀 [Feature] Order tags in tag cloud alphabetically
## :rocket: Feature

Order tags in the tag cloud in posts alphabetically.

### User Problem

The more tags there are in a cloud the harder it gets to find a tag.
hc-tag-cloud-unsorted

### Implementation

Order them alphabetically.

DriesCruyskens:
…-info page. Added test checking tags are present and shown alphabetically.

## 🍰 Pullrequest

Sorting happens in computed functions, seems like the Vue way to do it. It could also be done once in `apollo`'s `update` function. Let me know what you think.

Main post page:
image

More info: 
image

### Issues

- fixes #2244

### Todo

- [X] None

### DIFF
--- a/webapp/pages/post/_id/_slug/index.spec.js
+++ b/webapp/pages/post/_id/_slug/index.spec.js
@@ -3,6 +3,7 @@ import Vuex from 'vuex'
 import Vue from 'vue'
 import PostSlug from './index.vue'
 import CommentList from '~/components/CommentList/CommentList'
+import HcHashtag from '~/components/Hashtag/Hashtag'
 
 config.stubs['client-only'] = '<span><slot /></span>'
 config.stubs['nuxt-link'] = '<span><slot /></span>'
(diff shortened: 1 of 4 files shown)


webapp/pages/post/_id/_slug/index.spec.js
```diff
@@ -143,5 +144,48 @@ describe('PostSlug', () => {
         })
       })
     })
+
+    describe('tags shown in tag cloud', () => {
```
Tirokk, line 148: Nicely done 😍
Thanks for testing! 👍🏼

webapp/pages/post/_id/_slug/index.vue
```diff
@@ -179,6 +179,20 @@ export default {
       if (!author) return false
       return this.$store.getters['auth/user'].id === author.id
     },
+    sortedTags() {
```
Tirokk, line 182: I would love to have this imported as a more general function located in `components/utils/PostHelpers` or even in a more general utility file.
This way it can be reused in `webapp/pages/post/_id/_slug/more-info.vue`. 👇🏼

It can be used to sort the post categories in the future. 🙏🏼
A missing issue as far as I see …

webapp/pages/post/_id/_slug/more-info.vue
```diff
@@ -62,6 +62,20 @@ export default {
     post() {
       return this.Post ? this.Post[0] || {} : {}
     },
+    sortedTags() {
```
Tirokk, line 65: Would be nice to have a shared function … see above ☝🏼

DriesCruyskens:
@Tirokk I moved the sorting function to the `PostHelpers` file as you suggested.

I could indeed make a new file `utils/general.js` and make an `orderAlphabetically` function that accepts the list and a function argument to determine which object property the list should be ordered in (since hashtags and categories have different properties that need to be sorted: `id` and `name` respectively). But at that point I feel like a fully-fletched / battle-tested sort function is more in order. Especially if it's going to be used by different people in different context to prevent bugs. Let me know what you think!

PR #3634, opened 2020-06-13
feat: 🍰 display, select and upload UserProfile header image

### LINKED ISSUE #397: 🚀 [Feature] Display, select and upload UserProfile Header Image
### Select and upload UserProfile Header Image
image

- [ ] Display Header Image
- [ ] Select Dialog
- [ ] Upload and save Image on Server (persist)
- [ ] Crop Image properly

### Design & Layout - Alpha

image

### Design & Layout - Nitro

Not implemented:
image

DriesCruyskens:
## 🍰 Pullrequest

Since this feature is pretty similar to uploading and displaying a user's profile avatar I was able to mostly mirror the already existing implementation. Here you can see the drag and drop in action as well as the standard grey header for when no previous image is available:

dragndrop

Clicking opens up a file browser to select a file to upload just as with a profile avatar.

The issue also mentions cropping the image (to not use unnecessary bandwidth I assume) although I am not sure what to expected behavior is when using the app on mobile. As u can see in the next GIF, the image is mostly visible on mobile screens and even exposes the grey background if the image is not tall enough. Should the container have a max-height depending on the height of the image so no grey background is visible on wide images? Or use a CSS background image with repeat instead of an `` tag?

mobile resize

<details>
<summary>Extra notes</summary>

####  `UserProfileHeader` component
I did not not reuse the `hc-upload` component since that one already has avatar specific code and the styling needs to be different. I instead (sort off) merged the `hc-upload` and `UserAvatar` component into a new `UserProfileHeader` component.

#### back-end
I did not add new tests to `image.spec.js` since already contains test functions for a profile avatar and all functionality is shared. The only thing that had to be done is add the relationship type to the array of relationship types.

I also added a `cleanDatabase()` in `passwordReset.spec.js` and `statistics.spec.js` before all tests because they would fail. Presumably because of artifacts from previous tests?

#### Db seed
Unlike avatars, profile headers are not added to user accounts by default when seeding the database. I lot of tests failed because they do a hard-coded check of the total amount of images in the database. I tried changing the amounts but I can't quite figure out why the amounts are what they are as they seem very arbitrary so I thought it best to leave them as is. This way I can just mock profile headers when needed and not break any previous tests.

#### Locales
Purposefully didn't reuse the `user.avatar.submitted` but made a  new `user.profileHeader.submitted` although the text is the same. There is no text for failed toasts because that just prints the error message.

#### Storybook
I did not add a storybook entry since the component is very basic and does not have different states. Let me know if you still think this is necessary.
</details>

### Issues
- fixes #397

### Todo

- [X] Made header component and added it to the profile page + Graphql functionality.
- [X] Back-end functionality for the header image similar to avatar image.
- [X] Database seeds.
- [X] Unit tests web-app and back-end.
- [X] Integration tests.
- [X] Locales for the successful upload toast.
- [ ] Cropping image?

### DIFF
--- a/cypress/integration/common/profile.js
+++ b/cypress/integration/common/profile.js
@@ -24,13 +24,39 @@ Then("I should be able to change my profile picture", () => {
   );
 });
 
+Then("I should be able to change my profile header picture", () => {
+  const headerUpload = "onourjourney.png";
+
+  cy.fixture(headerUpload, "base64").then(fileContent => {
+    cy.get("#profileHeaderDropzone").upload(
+      { fileContent, fileName: headerUpload, mimeType: "image/png" },
+      { subjectType: "drag-n-drop", force: true }
+    );
+  });
+  cy.get(".profile-header-image")
+    .should("have.attr", "src")
+    .and("contains", "onourjourney");
+  cy.contains(".iziToast-message", "Upload successful").should(
+    "have.length",
+    1
+  );
+});
+
 When("I visit another user's profile page", () => {
   cy.openPage("profile/peter-pan");
 });
 
-Then("I cannot upload a picture", () => {
+Then("I cannot upload a profile picture", () => {
   cy.get(".base-card")
     .children()
     .should("not.have.id", "customdropzone")
     .should("have.class", "user-avatar");
+});
+
+
+Then("I cannot upload a profile header image", () => {
+  cy.get(".base-card")
+    .children()
+    .should("not.have.id", "profileHeaderDropzone")
+    .should("have.class", "profile-header");
 });
--- a/webapp/components/_new/generic/UserProfileHeader/UserProfileHeader.vue
+++ b/webapp/components/_new/generic/UserProfileHeader/UserProfileHeader.vue
@@ -1,0 +1,224 @@
+<template>
+  <div class="profile-header">
+    <vue-dropzone
+      v-if="user && this.editable"
+      id="profileHeaderDropzone"
+      :key="profileHeaderUrl"
+      ref="el"
+      :use-custom-slot="true"
+      :options="dropzoneOptions"
+      @vdropzone-error="verror"
+    >
+      <div class="dz-message" @mouseover="hover = true" @mouseleave="hover = false">
+        <img
+          v-if="user && user.profileHeader"
+          :src="user.profileHeader | proxyApiUrl"
+          @error="$event.target.style.display = 'none'"
+          class="profile-header-image"
+          :alt="imageAlt"
+        />
+        <div class="profileHeader-attachments-upload-area">
+          <div class="profileHeader-drag-marker">
+            <base-icon v-if="hover" name="image" />
+          </div>
+        </div>
+      </div>
+    </vue-dropzone>
+    <div v-else>
+      <img
+        v-if="user && user.profileHeader"
+        :src="user.profileHeader | proxyApiUrl"
+        @error="$event.target.style.display = 'none'"
+        class="profile-header-image"
+        :alt="imageAlt"
+      />
+    </div>
+  </div>
+</template>
+
+<script>
+import vueDropzone from 'nuxt-dropzone'
+import { updateUserMutation } from '~/graphql/User.js'
+
+export default {
+  name: 'UserProfileHeader',
+  components: {
+    vueDropzone,
+  },
+  props: {
+    user: {
+      type: Object,
+      default: null,
+    },
+    editable: {
+      type: Boolean,
+      default: false,
+    },
+  },
+  data() {
+    return {
+      dropzoneOptions: {
+        url: this.vddrop,
+        maxFilesize: 5.0,
+        previewTemplate: this.template(),
+      },
+      error: false,
+      hover: false,
+    }
+  },
+  computed: {
+    imageAlt() {
+      if (!this.user && !this.user.name) return 'Profile header image'
+      return 'Profile header image of ' + this.user.name
+    },
+    profileHeaderUrl() {
+      const { profileHeader } = this.user
+      return profileHeader && profileHeader.url
+    },
+  },
+  watch: {
+    error() {
+      const that = this
+      setTimeout(function () {
+        that.error = false
+      }, 2000)
+    },
+  },
+  methods: {
+    template() {
+      return `<div class="dz-preview dz-file-preview">
+                <div class="dz-image">
+                  <div data-dz-thumbnail-bg></div>
+                </div>
+              </div>
+      `
+    },
+    vddrop(file) {
+      const profileHeaderUpload = file[0]
+      this.$apollo
+        .mutate({
+          mutation: updateUserMutation(),
+          variables: {
+            profileHeader: {
+              upload: profileHeaderUpload,
+            },
+            id: this.user.id,
+          },
+        })
+        .then(() => {
+          this.$toast.success(this.$t('user.profileHeader.submitted'))
+        })
+        .catch((error) => this.$toast.error(error.message))
+    },
+    verror(file, message) {
+      if (file.status === 'error') {
+        this.error = true
+        this.$toast.error(file.status, message)
+      }
+    },
+  },
+}
+</script>
+
+<style lang="scss">
+.profile-header {
+  height: 100%;
+  position: relative;
+  background-color: DarkGrey; /* Fallback color */
+}
+
+.profile-header-image {
+  position: absolute;
+  top: 0;
+  left: 0;
+  right: 0;
+  width: 100%;
+}
+
+#profileHeaderDropzone {
+  height: 100%;
+}
+
+.dz-message {
+  height: 100%;
+}
+
+#profileHeaderDropzone .dz-preview {
+  transition: all 0.2s ease-out;
+  width: 160px;
+  display: flex;
+}
+
+#profileHeaderDropzone .dz-preview .dz-image {
+  width: 100%;
+  height: 100%;
+  object-fit: contain;
+  overflow: hidden;
+}
+
+#profileHeaderDropzone .dz-preview .dz-image > div {
+  width: inherit;
+  height: inherit;
+  border-radius: 50%;
+  background-size: cover;
+}
+
+#profileHeaderDropzone .dz-preview .dz-image > img {
+  width: 100%;
+}
+
+.profileHeader-attachments-upload-area {
+  height: 100%;
+  position: relative;
+  display: flex;
+  align-items: center;
+  justify-content: center;
+  cursor: pointer;
+}
+
+.profileHeader-attachments-upload-button {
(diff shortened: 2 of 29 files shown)


Tirokk:
Hey @DriesCruyskens ,

at my first glance you did again a wonderful job dear !!! 🍀🚀🚀🚀💫⭐️

So far it works really good for me in my browsers. 🙏🏼
Because I have to quit work now I will have a closer look tomorrow I hope.

As far I would say let’s make additional issues out of …

- cropping the header image (example done on the post creation page)
- deleting the header image (dito)
- and for later it would be nice the user gets able to delete the avatar as well (dito)

Otherwise this issue gets to big.

webapp/components/_new/generic/UserProfileHeader/userProfileHeader.spec.js
```diff
@@ -0,0 +1,217 @@
...
+        }
+        wrapper = Wrapper()
+      })
+
+      it('renders an image', () => {
+        expect(wrapper.contains('img')).toBe(true)
+      })
+    })
+
+    describe('with a relative avatar url', () => {
```
Tirokk, line 62: `header` instead of `avatar` …

webapp/components/_new/generic/UserProfileHeader/userProfileHeader.spec.js
```diff
@@ -0,0 +1,217 @@
...
+        }
+        wrapper = Wrapper()
+      })
+
+      it('adds a prefix to load the image from the uploads service', () => {
+        expect(wrapper.find('img').attributes('src')).toBe('/api/profileHeader.jpg')
+      })
+    })
+
+    describe('with an absolute avatar url', () => {
```
Tirokk, line 80: `header` instead of `avatar` …

webapp/components/_new/generic/UserProfileHeader/userProfileHeader.spec.js
```diff
@@ -0,0 +1,217 @@
...
+            name: 'Not Anonymous',
+            profileHeader: {
+              url: 'https://s3.amazonaws.com/uifaces/faces/twitter/sawalazar/128.jpg',
+            },
+          },
+        }
+        wrapper = Wrapper()
+      })
+
+      it('keeps the avatar URL as is', () => {
```
Tirokk, line 93: `header` instead of `avatar` …

webapp/locales/de.json
```diff
@@ -830,6 +830,9 @@
   "user": {
     "avatar": {
       "submitted": "Erfolgreich hochgeladen!"
+    },
+    "profileHeader": {
+      "submitted": "Erfolgreich hochgeladen!"
     }
```
Tirokk, line 836: May we can use one identifier for both cases if we show the same text?
like this:
```json
"image": {
  "submitted": "Erfolgreich hochgeladen!“
}
```

Or we do different texts as example:
- `Avatar-Bild erfolgreich hochgeladen!`
- `Profilbild erfolgreich hochgeladen!`

webapp/locales/en.json
```diff
@@ -830,6 +830,9 @@
   "user": {
     "avatar": {
       "submitted": "Upload successful!"
+    },
+    "profileHeader": {
+      "submitted": "Upload successful!"
     }
```
Tirokk, line 836: Same same ☝🏼

webapp/locales/es.json
```diff
@@ -827,6 +827,9 @@
   "user": {
     "avatar": {
       "submitted": "Carga con éxito"
+    },
+    "profileHeader": {
+      "submitted": "Carga con éxito!"
     }
```
Tirokk, line 833: Same same ☝🏼

webapp/locales/fr.json
```diff
@@ -791,6 +791,9 @@
   "user": {
     "avatar": {
       "submitted": "Téléchargement réussi"
+    },
+    "profileHeader": {
+      "submitted": "Téléchargement réussi"
     }
```
Tirokk, line 797: Same same ☝🏼

webapp/locales/it.json
```diff
@@ -734,6 +734,9 @@
   "user": {
     "avatar": {
       "submitted": ""
+    },
+    "profileHeader": {
+      "submitted": "Caricamento eseguito correttamente!"
     }
```
Tirokk, line 740: Same same ☝🏼

webapp/locales/nl.json
```diff
@@ -172,5 +172,10 @@
     "taxident": "Identificatienummer voor de belasting over de toegevoegde waarde overeenkomstig § 27 a Wet op de belasting over de toegevoegde waarde (Duitsland).",
     "termsAc": "Gebruiksvoorwaarden",
     "tribunal": "registerrechtbank"
+  },
+  "user": {
+    "profileHeader": {
+      "submitted": "Uploaden geslaagd!"
+    }
```
Tirokk, line 179: Same same ☝🏼

But here the avatar is missing …

webapp/locales/pl.json
```diff
@@ -363,6 +363,9 @@
   "user": {
     "avatar": {
       "submitted": "Przesłano pomyślnie"
+    },
+    "profileHeader": {
+      "submitted": "Przesłano pomyślnie"
     }
```
Tirokk, line 369: Same same ☝🏼

webapp/locales/pt.json
```diff
@@ -722,6 +722,9 @@
   "user": {
     "avatar": {
       "submitted": "Carregado com sucesso!"
+    },
+    "profileHeader": {
+      "submitted": "Carregado com sucesso!"
     }
```
Tirokk, line 728: Same same ☝🏼

webapp/locales/ru.json
```diff
@@ -823,6 +823,9 @@
   "user": {
     "avatar": {
       "submitted": "Успешная загрузка!"
+    },
+    "profileHeader": {
+      "submitted": "Успешная загрузка!"
     }
```
Tirokk, line 829: Same same ☝🏼

webapp/components/_new/generic/UserProfileHeader/UserProfileHeader.vue
```diff
@@ -0,0 +1,224 @@
...
+  height: 122px;
+  border-radius: 100%;
+  display: flex;
+  align-items: center;
+  justify-content: center;
+  color: hsl(0, 0%, 25%);
+  transition: all 0.2s ease-out;
+  font-size: 60px;
+
+  background-color: rgba(255, 255, 255, 0.7);
```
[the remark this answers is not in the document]
DriesCruyskens: Yes I changed it because I didn't think it was visible enough depending on the chosen header image.

webapp/locales/de.json
DriesCruyskens: I'll do the different texts one since it is better to letter the user know exactly what has changed to eliminate any confusion.

DriesCruyskens:
@Tirokk I implemented the changes you requested, everything should be OK now. For the locales tho, the English, German, French and Dutch ones are certainly correct but I translated the remaining ones using Google translate so I'm not too sure about them.

webapp/components/_new/generic/UserProfileHeader/UserProfileHeader.vue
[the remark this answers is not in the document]
Tirokk: 👍🏼

webapp/pages/profile/_id/_slug.vue
```diff
@@ -2,7 +2,7 @@
   <div>
     <ds-flex v-if="user" :width="{ base: '100%' }" gutter="base">
       <ds-flex-item width="100%">
-        <base-card class="profile-header">
+        <base-card class="profile-header-card">
```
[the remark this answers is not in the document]
DriesCruyskens: yes off course! my bad

webapp/components/_new/generic/UserProfileHeader/UserProfileHeader.vue
```diff
@@ -169,7 +164,10 @@ export default {
 
 .profileHeader-attachments-upload-area {
   height: 100%;
-  position: relative;
+  position: absolute;
+  top: 0;
+  left: 0;
+  right: 0;
```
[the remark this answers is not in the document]
DriesCruyskens: I yes, I went a little too fast there. Was so focused on getting the profile header right I forgot to check how it looked when no header present.. my bad. Should all be good now (if tests succeed). proof:

fix

DriesCruyskens:
@tirokk I made a couple of other changes to fix build and e2e errors. There are two errors remaining:

**1.** Toast not displaying the correct text when successfully uploading header image.
At first I tought this occurred because of the locale change, but when changing the success string in the e2e test the error remains. When looking at cypress' screenshot you can see the toast shows no text at all:
cypress-screenshot
While it does indeed show when testing locally. Not sure how to approach this problem, has this happened before?

**2.** See comment on code regarding `profile-header-card`.

cypress/integration/common/profile.js
```diff
@@ -28,9 +46,17 @@ When("I visit another user's profile page", () => {
...
     .should("not.have.id", "customdropzone")
     .should("have.class", "user-avatar");
 });
+
+
+Then("I cannot upload a profile header image", () => {
+  cy.get(".base-card")
+    .children()
+    .should("not.have.id", "profileHeaderDropzone")
+    .should("have.class", "profile-header-card");
```
DriesCruyskens, line 61: I suspect the fix you did changing `profile-header` to `profile-header-card` was wrong after all and the errors you thought it caused had in fact different causes. Since this test finds the element with `.base-card` class and then analyzes its children, its easy to see why it can't find `profile-header-card`. The test in its original state was in fact checking the child with the `profile-header` class:

image


tldr; The error regarding this test should be fixed by changing line 61 back to 
```suggestion
    .should("have.class", "profile-header");
```

DriesCruyskens, line 61: I suspect the fix you did changing `profile-header` to `profile-header-card` was wrong after all and the errors you thought it caused had in fact different causes. Since this test finds the element with `.base-card` class and then analyzes its children, its easy to see why it can't find `profile-header-card`. The test in its original state was in fact checking the child with the `profile-header` class:

image


tldr; The error regarding this test should be fixed by changing line 61 back to 
```suggestion
    .should("have.class", "profile-header");
```

Tirokk:
Hey @DriesCruyskens ,

thanks for investigating the error. 🙏🏼
I took spontaneously holidays and am now back to work.
In the next days I have not to much time, but will try to figure out the problem!

Tirokk:
On this cool feature we have an error in the tests.
If you could care for this one when you can’t find any motivation but work on tests @Mogge that would be really nice. 😘

I couldn't find it out so quickly …

PR #3676, opened 2020-06-25
fix: issue #3504 umlaut encoding

### LINKED ISSUE #3504: 🐛 [Bug] Wrong German umlaut handling for usernames at account creation
## :bug: Bugreport

At account creation, usernames are not encoded correctly if the profile name contains one or more German umlauts:
Ä/ä -> a
Ö/ö -> o
Ü/ü -> u

Examples:
- Sandra Märtens -> sandra-martens
- Oliver Stöckel -> oliver-stockel
- Günter Hürst -> gunter-hurst

1. Create a new account using one or more German umlauts in the profile name.
2. Check the username.

### Expected behavior

German umlauts should be encoded like this:
Ä/ä -> ae
Ö/ö -> oe
Ü/ü -> ue

Examples:
- Sandra Märtens -> sandra-maertens
- Oliver Stöckel -> oliver-stoeckel
- Günter Hürst -> guenter-huerst

### Version & Environment
 Type: [any]      
 - OS: [any]      
 - Browser: [any] 
 - Version [any]  
 - Device: [any]

ATOMktn:
Updated slugify middleware to handle umlauts to specification

## 🍰 Pullrequest
Bug caused by unexpected behavior of [Trott/slug](https://github.com/Trott/slug) middleware. Function only encodes to single English characters by default. According to the [documentation](https://github.com/Trott/slug/blob/master/README.md), a character map should be specified for specifically defined cases.

### Issues
- fixes #3504

### DIFF
--- a/backend/src/middleware/slugify/uniqueSlug.js
+++ b/backend/src/middleware/slugify/uniqueSlug.js
@@ -2,6 +2,14 @@ import slugify from 'slug'
 export default async function uniqueSlug(string, isUnique) {
   const slug = slugify(string || 'anonymous', {
     lower: true,
+    charmap: {
+        'Ä' : 'ae',
+        'ä' : 'ae',
+        'Ö' : 'oe',
+        'ö' : 'oe',
+        'Ü' : 'ue',
+        'ü' : 'ue',
+    }
   })
   if (await isUnique(slug)) return slug
 


ATOMktn:
Hi Triokk, thanks for the feedback and the opportunity to contribute. I'd be happy to do those things. 

I was worried the character map was a little clumsy, I will look into a more generalized solution for this.

PR #3695, opened 2020-07-03
Setup Human-Connection to work on codespaces

wenlock:
This sets up the repo so that it starts Human-Connection for dev when we work on codespaces.

### Issues

- None

### Todo

- [X] None


wenlock:
Sorry about that, this is not quiet tested but will open it back up once I have it working for our team.

PR #3769, opened 2020-08-03
fix: 🍰 Comment Counters Are Now Equal

### LINKED ISSUE #2112: 🐛 [Bug] Comment Counters are Different
## :bug: Bugreport
the counter on the card post page counts out deleted comments, the counter on the contribution counter counts out deleted comments with  

FireShot Capture 151 - Spenden- statt werbefinanziert! - Human Connection - human-connection social

FireShot Capture 149 - Human Connection - Human Connection - human-connection social

### LINKED ISSUE #3357: 🐛 [Bug] Comment updates only shown after page reload
## :bug: Bugreport

Sometimes comment updates are not shown instantly but only after a page reload.

1. Update an existing comment.
2. Click on the save button
3. Check the comment. Sometimes the updates are not shown.
4. Reload the page and check the updated comment.

### Expected behavior

Comment updates shown immeditately after clicking on the Save button. 

### Version & Environment
 Type: [Notebook]      
 - OS: [macOS]      
 - Browser: [Firefox] 
 - Version [74.0]  
 - Device: [MacBook]  

### Additional context

Reported by user joeletaxi on 26/27 March 2020

Elweyn:
## [WIP] 🍰 Pullrequest 2112
Changed the filter so that the comments that are deleted are not counted in the comment number anymore

### Issues
- fixes #2112
- fixes #3357

### Todo

- [ ] Test to check if the updated comment get directly updated on the frontend

### DIFF
--- a/webapp/components/CommentList/CommentList.spec.js
+++ b/webapp/components/CommentList/CommentList.spec.js
@@ -1,6 +1,5 @@
 import { config, mount } from '@vue/test-utils'
 import CommentList from './CommentList'
-import CommentCard from '~/components/CommentCard/CommentCard'
 import Vuex from 'vuex'
 import Vue from 'vue'
 
@@ -26,7 +25,10 @@ describe('CommentList.vue', () => {
               id: 'comment134',
               contentExcerpt: 'this is a comment',
               content: 'this is a comment',
-              author: { id: 'some-user' },
+              author: {
+                id: 'some-user',
+                slug: 'some-slug',
+              },
             },
             {
               id: 'comment135',
@@ -49,7 +51,7 @@ describe('CommentList.vue', () => {
         getters: {
           'auth/isModerator': () => false,
           'auth/user': () => {
-            return {}
+            return { id: 'some-user' }
           },
         },
       })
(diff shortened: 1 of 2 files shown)


Tirokk:
Hey @Elweyn ,

can you run the linter on `webapp` and on `backend`.
`yarn lint --fix`
That’s the reason why the Travis build fails.

PS: In case you work on the locale files in the `webapp` please run as well
`yarn locales --fix`
before you push.

Elweyn:
Hey @Tirokk, 
thanks was not sure how the semantic had to be done!

I just made the linter on webapp and backend.

webapp/components/CommentList/CommentList.spec.js
```diff
@@ -28,6 +28,13 @@ describe('CommentList.vue', () => {
               content: 'this is a comment',
               author: { id: 'some-user' },
             },
+            {
+              id: 'comment135',
+              contentExcerpt: 'this is a deleted comment',
+              content: 'this is a deleted comment',
+              deleted: true,
+              author: { id: 'some-user' },
+            },
```
Tirokk, line 37: Because of a different solution I suggest in the next comment👇🏼 this test makes no sense anymore.

Would be cool to replace it with a test in `backend/src/schema/resolvers/posts.spec.js` which checks that only comments not deleted and not disabled are counted by the Cypher queries.

webapp/components/CommentList/CommentList.vue
```diff
@@ -1,7 +1,10 @@
 <template>
   <div id="comments" class="comment-list">
     <h3 class="title">
-      <counter-icon icon="comments" :count="postComments.length" />
+      <counter-icon
+        icon="comments"
+        :count="postComments.filter((comment) => !comment.deleted).length"
```
Tirokk, line 6: I looked at the other place in `webapp/components/PostTeaser/PostTeaser.vue` where the counter is set on a post teaser. This is the part which differs from that here.
Because I know this value of `post.commentsCount` comes direct from the backend and because I have found this doesn’t count `comment.deleted` or `comment.disabled` I will suggest this solution to be taken over here.

```suggestion
        :count="post.commentsCount“
```

A test in `backend/src/schema/resolvers/posts.spec.js` is missing which checks that only comments not deleted and not disabled are counted by the Cypher queries.
If you are motivated would be nice to implement this one. 💫🍀

webapp/components/CommentList/CommentList.spec.js
Elweyn: Thanks for the response. I think it is important to check if their are deleted comments that it does not break the other tests, so we could let this one and add another one in backend/src/schema/resolvers/posts.spec.js to check if the count is setten correctly what do you think?

webapp/components/CommentList/CommentList.vue
Elweyn: Thanks a lot, this one seems a lot better than a filter on the deleted comments!
I will implement this so the frontend code stays clean logic.

Elweyn:
Hi @Tirokk 
after @Mogge and I realised that this class made some trouble with the comment tickets. 
We decided to change it so that we don't work on the computed postComments anymore but on post.comments.

Since the value of post.commentsCount is not updated as long as we don't make a reload of the page, we decided to make a computed property commentsCount that make use the first written filter.
So I added a new comment that was disabled to make sure that the count is stil 1 even when their is a disabled comment and a deleted comment. 

Hope this seems good.
Cheers

webapp/components/CommentList/CommentList.spec.js
```diff
@@ -28,6 +28,20 @@ describe('CommentList.vue', () => {
...
+            {
+              id: 'comment136',
+              contentExcerpt: 'this is a disabled comment',
+              content: 'this is a disabled comment',
+              disabled: true,
+              author: { id: 'some-user' },
+            },
```
Tirokk, line 44: Okay. Cool!
Then this test extension is needed. 😍

webapp/components/CommentList/CommentList.vue
```diff
@@ -39,6 +39,14 @@ export default {
     postComments() {
       return (this.post && this.post.comments) || []
     },
```
Tirokk, line 41: This code is now superfluous, I think.

Mogge:
Testing that a comment is updated after editing would be great.

PR #3867, opened 2020-09-09
[WIP] for archiv status: backend funktion turn of

### LINKED ISSUE #3863: 🚀 Upload  Contributions + [¿Userdata?]
## :rocket: Feature
 Upload of text entries and images for a user. 

### User problem
The user should have the possibility to export his data. 
The user should have the possibility to import exported data again. 

### Data to be exported would be

- written articles 
- User Profile Data (This is a matter of debate! Maybe not absolutely necessary)

best way would be: this data should be importable again.

ogerly:
## 🍰 Pullrequest
the network can be set to a read only mode. all contents can be read. 
all create and update functions are disabled. 
all delete functions remain. 

there is a global variable in webapp and backend 
"activeNetwork" 
which 
"true" = network normal (all functions are available)
or 
"false" = network in archive mode (read only)
can be. 

### Issues
- fixes #3863

### DIFF
--- a/backend/src/middleware/permissionsMiddleware.js
+++ b/backend/src/middleware/permissionsMiddleware.js
@@ -1,8 +1,13 @@
-import { rule, shield, deny, allow, or } from 'graphql-shield'
+import { rule, shield, deny, allow, or, and } from 'graphql-shield'
 import { getNeode } from '../db/neo4j'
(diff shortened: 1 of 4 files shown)


Tirokk:
Not needed anymore …

PR #3899, opened 2020-09-24
feat: Export User Data

Mogge:
## 🍰 Pullrequest
The users can download a Json with their profile data, all the posts they have written, and all the posts they have commented.

### Issues
- #3863


ogerly (approved):

backend/src/schema/resolvers/userData.spec.js
```diff
@@ -0,0 +1,211 @@
...
+        neode,
+        user: authenticatedUser,
+      }
+    },
+  })
+  query = createTestClient(server).query
+})
+
+afterAll(async () => {
+  //  await cleanDatabase()
```
Tirokk, line 38: Should be not commented. 😎

Tirokk (changes requested):

PR #3954, opened 2020-10-16
feat: Export User Data Update

Mogge:
## 🍰 Pullrequest
fetchPolicy added to apollo for userData.
posts that are disbaled or deleted are not available for download.
Clicking on an post title to download image opens new tab for the image.


Tirokk (approved):
Endgame! 💫
Looks good @mogge 👍🏼🚀✨

PR #3971, opened 2020-10-26
feat: Improved Data Export

Mogge:
## 🍰 Pullrequest
spinner added to download button, only the posts written by the user are exported


Tirokk (approved):
Hey @Mogge ,

looks like this makes sense to reduce the data amount load of the query. 👍🏼

PR #4551, opened 2026-05-11
Security Fix: Email enumeration in signup and email-change flows

### LINKED ISSUE #1832: 🐛 [Bug] Unnecessary disclosure if email address exists
## :bug: Bugreport

When you change your email address or when you create a new user account, you get an error message if the email already exists. This is suboptimal as it discloses if an email address exists. 

Screenshot - 2019-10-08T005456 608

1. Try to change your email or alternatively create a new user account
2. Type in an email that exists already
3. Submit
4. See a lovely error message

### Expected behavior
We should just tell the user on the browser sth. like: "Instructions have been sent to the email address" as we do it with password resets already. That way, an attacker cannot tell if the email address was in our database. Only the owner of the email will get a message.

PeterDonate1999:
Closes #1832

## Summary
The signup and email-change flows were leaking whether an email 
already exists in the database, enabling account enumeration attacks.

## Fix
Modified existingEmailAddress.js, registration.js, emails.js, 
and emailMiddleware.js to return uniform responses regardless 
of whether the email exists.

Updated resolver tests to verify new behavior.


PeterDonate1999:
I've submitted a fix for this in #4551. Please review when you get a chance.

